PR #36, opened 2015-02-20
Bug Fixes and Improvements

justinjdickow:
- Fix Empty perform iteration request
  - Fix type of name from string to enum
  - SendLocation implemented on HTML5 HMI
  - Fixed PI response on VR rejection due to high priority.
  - Fix Apps not responsive/not able to start app/apps remain listed on SYNC even after USB disconnect
  - Mobile API change and processing capabilities
  - Change perform interaction request conditions.
  - Fix SDL must always start 3sec timer before resuming the HMILevel of the app
  - Remove redundant StartSavePersistentDataTimer() call.
  - Change wrong predicate name to right.
  - Added stream request handling feature
  - Made streaming timeout in media manager configurable
  - Put navi app in LIMITED in case of phone call
  - Handling of audio state for applications
  - Add stop streaming timeout into ini file
  - Implement HMILevel resumption for job-1
  - Fix result code ABORTED when interrupts it by Voice recognition activation
  - Fix incorrect value parameter unexpectedDisconnect in BCOnAppUnregistered
  - Fix SDL send BC.OnAppUnregistered with "unexpectedDisconnect" set to "true" in case received from HMI OnExitAllApplications {"reason":"MASTER_RESET"}
  - Fix Update ini file for iAP1 support
  - Current working directory added to image path
  - Fix helpers to make it workable with more then 2 parameters
  - DCHECK() for ManageMobileCommand() replaced with log message because the latter returns false in some regular situations (e.g. TOO_MANY_PENDING_REQUESTS, see SDLAQ-CRS-10)
  - Remove connection after closing.


lea255ace:
Requires sqlite > 3.7.11.
Link to build from source for 12.04 here: http://www.linuxfromscratch.org/blfs/view/svn/server/sqlite.html

PR #38, opened 2015-02-23
Implementation of SDL4.0 feature.

AGaliuzov:
The pull request contains changes regarding maintenance of communication by 4th protocol version.

The API changes that are included:

*SystemRequest:* now supporting two additional parameters: `QUERY_APPS` and `LAUNCH_APP`
*UpdateAppList:* now contains the marker whether application is able or not.

All other changes are not related to the public API:

The `ApplicationManager` now keeps two lists of applications - the first one is already registered and the second one is applications which available on mobile devices for registration and activation.
The application from second list will move to the first one during real registration.


lea255ace:
This pull request was against the feature/all_the_features branch, which was deleted when merged into develop. Please re-submit against the develop branch.

PR #47, opened 2015-03-20
Implement ability to send ttsName and vrSynonyms during QUERY_APP

AGaliuzov:
In case QUERY_APP system request executes that received JSON has to have
language entry where descibed proper values for ttsName and vrSynonyms fields.


justinjdickow:
I gave this a 4.1 milestone because ideally we have a server implementation which provides these fields. Right now sdl_server only specifies the JSON but does not implement it.

PR #50, opened 2015-03-23
PoliciesManager: Adding support of RequestType field to the policy table and appropriate logic.

dev-gh:
SystemRequest/Putfile Re-design:
Current issues:
No way to correlate an app's putfile request and a systemRequest request.
No way to regulate systemRequest via policy table.
Current way of identifying current SystemRequest capabilities is not ideal.

Proposal:
The SystemRequest filename parameter can refer to a ‘putfile’ file so that the system knows what to do with the file. For example, how the system identifies an ivsu update vs. traffic information vs. etc.

An additional parameter shall be added to SystemRequest to set a flag to refer to a file already onboard, sent via putfile, instead of the binary data send along with the SystemRequest RPC: putFile=true

example request: SystemRequest(filename=putfile file, RequestType="TRAFFIC", putFile=true)

New field "RequestType" had been added to policy table structure in order to support updated functionality described above. 
Also OnAppPermissionChanged and OnAppRegistered notifications from HMI API had been extended with new "RequestType" parameter support.


PKramchaninov:
The pull request has been merged manually to release/4.0.0

PR #73, opened 2015-04-14
APPLINK-12401

DKlimenko:
SYNC Returns Empty templatesAvailable Array After Opening Non-Media App With Alternate Template. In case templatesAvailable Array is empty use templatesAvailable from UI.GetCapabilities response.


AGaliuzov:
Pull request already in release branch

PR #79, opened 2015-04-16
APPLINK-11722

DKlimenko:
[RTC 566543] Generic Error received for PI with message as 'Wrong Choice id was received from HMI'


AGaliuzov:
merged manually into release/4.0.0 branch

PR #111, opened 2015-05-12
Send hybrid service 0x0F if outgoing message has binary data

### LINKED ISSUE #110: When Binary Data is attached to a message going out from HMI the service type is not set to 0x0F Hybrid Service
1. Need to investigate whether setting this correctly has any implications on current systems
  2. Need to update the application manager (or other?) to set the service type correctly

justinjdickow:
[NOT READY TO MERGE]

This hotfix addresses an issue where a message being sent to mobile is not sent over the hybrid service if it contains binary data.

Fixes #110 

After this change is merged into the release, it should also be fixed in develop.


justinjdickow:
This can be merged when the iOS and Android issues referenced above are verified

PR #118, opened 2015-05-19
APPLINK-13063: During PerformInteraction SDL is not forwarding UI.OnKeyboardInput to non-navigation Apps

DKlimenko:


AGaliuzov:
The pull reuest looks like hotfix. so it has been merged into release/4.0.0 instead

PR #120, opened 2015-05-21
Switch timer to use ms

dev-gh:
Now timer class can use milliseconds instead of seconds.


AGaliuzov:
Please resolve the conflicts before merging

PR #147, opened 2015-06-26
Remove "Image" field from softbuttons with image_type="TEXT"

Anatoliy-Leshin:
Remove "Image" field from softbuttons with image_type="TEXT"


AGaliuzov:
please exclude `genivi` word from branch name

PR #158, opened 2015-07-15
Hotfix/malformed messages counting mechanism fix

Anatoliy-Leshin:
Fix malformed messages counting mechanism
    
We count now only switches from correct message to malformed.
Continous malformed message of any length is one malformed message.


AGaliuzov:
`Not a subject for review` is forbidden comment. Please change it

PR #164, opened 2015-07-17
Add Telnet logging

LuxoftAKutsan:
netcat can't collect logs via SocketHub or socket because of specific
log4cxx binary format of logs

Add to log4cxx.properties logging via telnet to collect
this logs with netcat or telnet in raw text format

Related Issues: APPLINK-14183


AGaliuzov:
Alexandr, please pay attention to the branch naming

Issue #166, opened 2015-07-21
Policy Table Update Process (Ford Vs Genivi Impl)


JackLivio:
The policy table update process outlined in the [Software Detailed Design] (https://app.box.com/s/ohgrvemtx39f8hfea1ab676xxrzvyx1y) on pages 36 and 37, shows the Genivi specific implementation and the Ford specific implementation.

Currently the SDL Core and SDL Server repositories implement the Ford specific version of updating the policy table, and due to limitations of the Web browser, the update cannot be made without a "workaround".

If you look at the diagrams in the Software detailed design document, the Genivi implementation for a policy table update does not require the HMI to parse/decrypt the policy table update data received from SDL Server. The Genivi implemntation instead uses a put file request which then sends the policy table update to the policy handler via ReceiveMessageFromSDK(file_path, content). 

In the diagram, notice that the function ReceiveMessageFromSDK(file_path, content) is used by both Ford's and Genivi's implementation and the rest of process following that function call is the same for both implementations.

The differences between the two implementations are:
- The Genivi specific implementation for receiving a policy table update is initiated by a put file request.
- The Ford specific implementation for receiving a policy table is initiated by a system request.
- Genivi's design does not specifically mention policy table encryption.
- Ford's design requires the HMI to decrypt the policy table and write the decrypted data to a new file(Main problem since the web browser cannot do this!)
- Genivi's design requires less steps and does not require the HMI to do anything (Makes more sense for development/testing).

Question: Why are there two different implementations for the same process?

I would like to open a discussion for choosing one of these processes, or perhaps redefine a new process as the "standard" for updating a policy table.

My suggestion for a possible implementation: 

- Initiate policy table update via system request from server (Same as Ford's design)
- Using #ifdefs within SystemRequest::Run(), if the policy table encryption #ifdef is enabled, continue with Ford's implementation and pass the file path to the hmi to be 'decrypted'. 
- Alternatively, If policy table encryption #ifdef is disabled, then parse the incoming data within SystemRequest::Run() and then call ReceiveMessagaeFromSDK(file_path, content) immediately (similar to Genivi's design except the put file request has been replaced by the system request rpc).

JackLivio:
[...]

PR #175, opened 2015-07-28
Add flags to enable/disable HMI policy table decryption

JackLivio:
This pull request solves the policy table update issue addressed in issue #166 .

Added option for Policy Table to be parsed by SDL Core as an Array object form SDL Server, or the option for the HMI to do the "decryption" if an encrypted string is passed from SDL Server.

The #ifdef USE_HMI_PTU_DECRYPTION is set by the CMAKE option ENABLE_HMI_PTU_DECRYPTION in the top level cmakelists.txt file. When this option is set to "ON", the HMI will have to decode the policy table update (which is not implemented in the current web hmi). When this option is set to "OFF", Core will be expecting an array from sdl server with the policy table update located in the first element of the "data" array.

### DIFF
--- a/src/components/policy/src/policy/src/policy_manager_impl.cc
+++ b/src/components/policy/src/policy/src/policy_manager_impl.cc
@@ -72,12 +72,12 @@ void PolicyManagerImpl::set_listener(PolicyListener* listener) {
 #ifdef USE_HMI_PTU_DECRYPTION
 
 utils::SharedPtr<policy_table::Table> PolicyManagerImpl::Parse(
-  const BinaryMessage& pt_content) {
+    const BinaryMessage& pt_content) {
   std::string json(pt_content.begin(), pt_content.end());
   Json::Value value;
   Json::Reader reader;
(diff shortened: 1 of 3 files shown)


src/components/policy/src/policy/src/policy_manager_impl.cc
```diff
@@ -69,18 +69,44 @@ void PolicyManagerImpl::set_listener(PolicyListener* listener) {
...
+#else
+
+utils::SharedPtr<policy_table::Table> PolicyManagerImpl::ParseArray(
+  const BinaryMessage& pt_content) {
+  std::string json(pt_content.begin(), pt_content.end());
+  Json::Value value;
+  Json::Reader reader;
+  if (reader.parse(json.c_str(), value)) {
+    //For PT Update received from SDL Server.
+    if(value["data"].size()!=0){
```
AGaliuzov: We're using google coding style. Please refer to it in order to fix indentations

PR #189, opened 2015-08-13
Fix inccorrect read of bool values from .ini file

Anatoliy-Leshin:


justinjdickow:
This should be merged into release/4.0.0 since it's a hotfix

PR #192, opened 2015-08-17
App is not unregistered by request while in none hmi level reason

alambin:
App is not unregistered by REQUEST_WHILE_IN_NONE_HMI_LEVEL reason
    
hmi_level member of RequestInfo objects was not used. It caused problem when counting requests with specific hmi_level.


alambin:
App is not unregistered by REQUEST_WHILE_IN_NONE_HMI_LEVEL reason
    
hmi_level member of RequestInfo objects was not used. It caused problem when counting requests with specific hmi_level.

PR #200, opened 2015-08-27
Remove legacy  _EXT and INFO macro

pestOO:
Removed unused _EXT macro definition
Moved logger macro in alphabetical order
Removed legacy _EXT macro usage
Replaced INFO macros to TRACE and debug
Removed symlinks

Implements: APPLINK-15680

### DIFF
--- a/src/appMain/life_cycle.cc
+++ b/src/appMain/life_cycle.cc
@@ -88,7 +88,7 @@ LifeCycle::LifeCycle()
 { }
 
 bool LifeCycle::StartComponents() {
-  LOG4CXX_INFO(logger_, "LifeCycle::StartComponents()");
+  LOG4CXX_AUTO_TRACE(logger_);
   transport_manager_ =
     transport_manager::TransportManagerDefault::instance();
   DCHECK(transport_manager_ != NULL);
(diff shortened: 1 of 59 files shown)


AGaliuzov:
@pestOO , please do rebase the PR since it is not mergable

pestOO:
@AGaliuzov, rebased.

Waht about not hot, but common fixes like that? =)

AGaliuzov:
@pestOO ,
According to the gitflow any bug fix branch should start from `hotfix/` prefix.

In common such naming policy uses for hotfixes into master or release branch.
But in order to be uniform we use such naming policy for develop either

justinjdickow:
@pestOO 
Aleks is correct. If a branch adds a feature it has the `feature/` prefix for the branch name. If a branch fixes a bug is has the `hotfix/` prefix. These are generally committed to release branches (such as release/4.0.0 or master) but may also be applied to pull requests into `develop` when a feature was considered complete but a bug is found after the merge.

justinjdickow:
Basically the only time that a `hotfix/` branch is merged into develop is if it fixes a bug for a feature that exists in develop but doesn't exist in a release or the master branch (so a fix for a new feature).

Issue #199, opened 2015-08-24
Wrong use of 'device_info' for UpdateDeviceList and OnDeviceChosen


JackLivio:
## Issue
When selecting a device from the HMI's "Change Devices" option, I get the error from core "Incorrect parameter from HMI". After doing some investigating I found that OnDeviceChosen is sending back the mac address of the selected device in a string to core. It is being rejected by core because the expected parameter should be an integer. Also this parameter shouldn't be the mac address of the device, but the connection handler id instead.

## Current RPC Behavior
UpdateDeviceList sends an updated list of connected devices to the HMI and stores their "DeviceInfo" for later use. When a device is selected from that device list on the HMI, the notification OnDeviceChosen is sent to core indicating that the user wants to connect SDL to that device. Both of these RPC's are sending the device's mac address as the 'id' parameter.

Core then attempts to find the device and connect to it by using the connection handle id, not the mac address.

```
void OnDeviceChosenNotification::Run() {
  LOG4CXX_AUTO_TRACE(logger_);

  if ((*message_)[strings::msg_params].keyExists(strings::device_info)) {
    ApplicationManagerImpl::instance()->ConnectToDevice(
        (*message_)[strings::msg_params][strings::device_info][strings::id]
            .asInt());
  }
}

```
```
void ApplicationManagerImpl::ConnectToDevice(uint32_t id) {
  // TODO(VS): Call function from ConnectionHandler
  if (!connection_handler_) {
    LOG4CXX_WARN(logger_, "Connection handler is not set.");
    return;
  }

  connection_handler_->ConnectToDevice(id);
}
```
```
void ConnectionHandlerImpl::ConnectToDevice(
    connection_handler::DeviceHandle device_handle) {
  connection_handler::DeviceMap::const_iterator it_in;
  it_in = device_list_.find(device_handle);
  if (device_list_.end() != it_in) {
    LOG4CXX_INFO_EXT(logger_,
                     "Connecting to device with handle " << device_handle);
    if (transport_manager_) {
      if (transport_manager::E_SUCCESS
          != transport_manager_->ConnectDevice(device_handle)) {
        LOG4CXX_WARN(logger_, "Can't connect to device");
      }
    } else {
      LOG4CXX_ERROR(logger_, "Null pointer to TransportManager.");
    }
  } else {
    LOG4CXX_ERROR(
        logger_, "Application Manager wanted to connect to non-existing device");
  }
}

```

### Fix
[...]

PR #205, opened 2015-08-31
Changed device_info id to string and mac_address

JackLivio:
Pull request related to issue #199. Changed device info id field to type String and changed the field to use the mac address across sdl core uniformly, instead of the device handle in some places.

Code was forked from PR #179 and two lines from Pr #150 which were merged into develop. This code was needed as a hotfix into release to allow for OnDeviceChosen to not throw incorrect hmi parameter error, as well as creating uniform use of device_info id throughout core.


justinjdickow:
@AGaliuzov This change was previously made in develop because it also contained a "feature", but this particular logic is needed in the release. Can you have someone review then merge this into release/4.0.0?

PR #210, opened 2015-09-11
Hotfix/incorrect perform interaction response

mrBoome:
Fix incorrect PerformInteraction response

In function "ProcessPerformInteractionResponse" change the location, where SDL search choiceID, when WARNINGS come.


justinjdickow:
@AGaliuzov Should this be merged into release/4.0.0?

AGaliuzov:
@justinjdickow I don't think so. As I do not observe issue on release branch. I think We need to perform additional check for it

PR #212, opened 2015-09-11
Remove parsed JSON and invalid messages from MessageBroker's buffer

asenkk:
With the previous code, messages that cannot be parsed (pure garbage,
invalid JSONs, JSONs that are not objects or does not have jsonrpc
member set correctly) remained in MesageBrocker's TcpServer receive
buffer for the corresponding file descriptor forever.
Because the garbage is in front of the buffer and next messages are
appended at the end, these next messages are not processed, since
parsing stops on first error.
Also if parsed message has different length than the original (this
is possible if there are floating point numbers for example), the
original message is not deleted properly and next message may be
broken.
The fix is to handle invalid messages and remove invalid or parsed
data from the buffer, without removing any potentially valid, but
incomplete data (masked websocket frames or SDL messages).
It is still possible to misinterpret garbage as a beginning of a
incomplete valid frame. In this case we are doing this: when we have
old data in the buffer and a new message arrives, if we can't parse
the buffer WITH the new message - try to parse only the new message
and clear the buffer on success, because if the new message can be
parsed alone, then the old must be garbage.
Fixed problem when deleting receiving buffer of disconnected client.

### DIFF
--- a/src/3rd_party-static/MessageBroker/include/CMessageBroker.hpp
+++ b/src/3rd_party-static/MessageBroker/include/CMessageBroker.hpp
@@ -61,9 +61,10 @@ namespace NsMessageBroker
       * \brief Receive data from TCP server (from client).
       * \param fd FileDescriptor of socket.
       * \param aJSONData JSON string.
+      * \param tryHard give up on first JSON parse error or try to workaround it.
       */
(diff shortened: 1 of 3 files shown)


asenkk:
Wrong branch 'develop'. Should be in 'release/4.0.0'

PR #213, opened 2015-09-11
Fix Logger's TelnetAppender socket reopen on fast restart

asenkk:
When SDL is stopped, TelnetAppender's server socket listening on port
6676 will remain in TIME_WAIT state (if there was a client connected
to it) fr some time, depending on the operating system settings.
With the old code, if we restart SDL soon after stopping it,
TelnetApplender will fail to bind to its port, because 'it is in use'.
The fix is to set SO_REUSEADDR option to server socket when opening
it - this will allow the port to be reused when it is in TIME_WAIT
state.


asenkk:
Wrong branch 'develop'. Should be in 'release/4.0.0'

PR #217, opened 2015-09-16
Fix crashes in the processing of SystemRequest(QUERY_APPS)

Anatoliy-Leshin:
1. Fix crash when processing SystemRequest with QUERY_APPS.
Process tts_name as string, not as array. tts_name in QueryApps json file is "string" according to APPLINK-11731, tts_name in QueryApps json file is "string" according to APPLINK-11731, tts_name in UpdateApplist is "array of strings" according to API.xml.
Fix: APPLINK-15783

2. Fix crash when accessing to non-existing object with Shared-pointer operator "->".
Use device_id from connection_handler.
Fix: APPLINK-15777, APPLINK-15778

### DIFF
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -351,10 +351,8 @@ bool ApplicationManagerImpl::IsAppTypeExistsInFullOrLimited(
       return true;
     }
   }
-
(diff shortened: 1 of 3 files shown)


AGaliuzov:
One more thing. I think you have too send this PR into release/4.0.0 branch

PR #282, opened 2015-12-01
Add checking of SO field.

dtrunov:
Add checking of SO field during creation of request to HMI.
If user calls to nonexistent field from SO, SO creates this field with default value.

Closes-bug: APPLINK-13006


src/components/application_manager/src/message_helper.cc
```diff
@@ -1102,11 +1102,11 @@ smart_objects::SmartObjectList MessageHelper::CreateAddCommandRequestToHMI(
       msg_params[strings::menu_params] = (*i->second)[strings::menu_params];
       msg_params[strings::app_id] = app->app_id();
 
-      if (((*i->second)[strings::cmd_icon].keyExists(strings::value))
+      if (((*i->second).keyExists(strings::cmd_icon))
```
pestOO: Could we move reference to the first and second value of pair at the beginning of the loop?
It will make code more readable.

pestOO:
Reviewed

PR #284, opened 2015-12-01
Hotfix/core dump if make master reset

dtrunov:
Fix core crash during MASTER_RESET
    
Forbid processing message from mobile if Core begins stopping of work.
    
Closes-bug: APPLINK-17367


pestOO:
Reviewed

PR #289, opened 2015-12-02
Fix core crash: assert to double taking mutex

AGritsevich:
SDL crashed when trying got mutex in sub method. Change normal mutex to recursive to fix this problem.
Fixes: [APPLINK-19317](https://adc.luxoft.com/jira/browse/APPLINK-19317)

### DIFF
--- a/src/components/security_manager/src/ssl_context_impl.cc
+++ b/src/components/security_manager/src/ssl_context_impl.cc
@@ -50,8 +50,6 @@ CryptoManagerImpl::SSLContextImpl::SSLContextImpl(SSL *conn, Mode mode)
     bioIn_(BIO_new(BIO_s_mem())),
     bioOut_(BIO_new(BIO_s_mem())),
     bioFilter_(NULL),
-    //recursive mutex
-    bio_locker(true),
     // TODO(EZamakhov): get MTU by parameter (from transport)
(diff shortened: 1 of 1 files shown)


AGritsevich:
Please review: @pestOO , @AGaliuzov

pestOO:
Reviewed

AGaliuzov:
The usage of recursive mutex looks like workaround. Do you now a really reason why SDL lock this mutex twice? Maybe better to fix this instead of masking problem with recursive mutex

anosach-luxoft:
+1 to remove recursive mutex

AGritsevich:
Anyway it's way cheap and fast.

AGritsevich:
Ok, I'll try to find real problem. Thank you for opinion.

AGritsevich:
I removed recursive mutex and just make minimal scope for automutex. I hope you'll like it!
Please review @pestOO, @AGaliuzov, @anosach-luxoft.

anosach-luxoft:
:+1:

AGritsevich:
@anosach-luxoft, I don't think so.

anosach-luxoft:
@AGritsevich, you shouldn't think about it :smile: You have to be sure.

AGritsevich:
I'm sure.

PR #291, opened 2015-12-04
Feature/conditions for sdl to send play tone param via speak request to hmi

dtrunov:
Removed PlayTone notification to HMI.
Moved this parameter to TTS.Speak.


src/components/application_manager/src/commands/mobile/alert_request.cc
```diff
@@ -359,34 +360,22 @@ void AlertRequest::SendSpeakRequest(int32_t app_id) {
   using namespace smart_objects;
   // crate HMI speak request
   SmartObject msg_params = smart_objects::SmartObject(SmartType_Map);
-
-  msg_params[hmi_request::tts_chunks] =
-      smart_objects::SmartObject(SmartType_Array);
-  msg_params[hmi_request::tts_chunks] =
-    (*message_)[strings::msg_params][strings::tts_chunks];
+  if ((*message_)[strings::msg_params].keyExists(strings::tts_chunks) &&
```
AGaliuzov: You check the same value twice, maybe it worth to cache this value?

dtrunov: I applied your suggestion. Please make merging.

src/components/interfaces/HMI_API.xml
```diff
@@ -2199,15 +2199,6 @@
         <description>Must be true if supported</description>
       </param>
     </function>
```
pestOO: xml version?

dtrunov: What did you mean?

PR #303, opened 2015-12-22
Hotfix/remove redundant test main

OHerasym:
Deleted redundant main test files, deleted unused test directory, moved tests to 3rd-party-static & tools/test. 

Related:
             APPLINK-20109


OHerasym:
Please review: @malirod, @AByzhynar , @Kozoriz, @VVeremjova

malirod:
I see two huge commits with the same messages. I think this is not right. Squash them.
I see that deleted 9162 lines, and added only 28. Is this correct? Are you sure that it's right to delete all those files?

AByzhynar:
Reviewed

OHerasym:
@malirod , Yes, all files were in test folder (not used now). Same things are already changed in develop branch, I just do same task in master. Two commits have different commit messages, they have same title. One commit delete files. Second commit add files to tools/test/ & 3rd-party-static/test/.
Is it will be correct squash them?

malirod:
I've looked briefly these commits and I cannot see any logic in them. So I think it's ok to squash. This way we would have one huge "illogical" commit.

OHerasym:
@malirod , Done

malirod:
Reviewed

malirod:
@OHerasym, I have doubts that you've done  the right thing with this PR. As far as I understand the APPLINK-20109, your task was to choose one file from two available and delete it. I don't think that deletion of the big amount of tests is the right way to go.
Even if it's ok to delete those tests (this should be confirmed by someone), I propose to split this PR into two. One will resolve the APPLINK-20109. And another will drop redundant tests(if requred).

VVeremjova:
policy_table_validator should be also moved in tools

OHerasym:
@malirod , @VVeremjova , Added policy_table_validator. Created 3 commits:
 1)Remove reduntant test main files.
2) Add tests in tools & 3rd-party-static.
3) Delete unused folder test

Kozoriz:
Reviewed

VVeremjova:
Reviewed

malirod:
Reviewed

PR #315, opened 2016-01-06
Fix putting NAVI non-media app in BACKGROUND with a second media app

asenkk:
According to the requirements (APPLINK-8531), SDL must support the apps
of the following AppHMIType to be in LIMITED at one and the same time:
MEDIA media, NAVIGATION non-media, COMMUNICATION non-media. But because
of a bug, when activating a second MEDIA media app, the previously activated
NAVIGATION non-media app was put from LIMITED to BACKGROUND. It should
stay in LIMITED.
The fix is taken from CustomSDL/sdl_panasonic/develop(451042b4)
'APPLINK-13597 Buxfix in StateCtrl'.

Fixes: APPLINK-20258

### DIFF
--- a/src/components/application_manager/src/state_controller.cc
+++ b/src/components/application_manager/src/state_controller.cc
@@ -66,29 +66,56 @@ void StateController::HmiLevelConflictResolver::operator ()
   using namespace mobile_apis;
   using namespace helpers;
   DCHECK_OR_RETURN_VOID(state_ctrl_);
-  if (to_resolve == applied_)
-    return;
-  if (Compare<HMILevel::eType, EQ, ONE>(state_->hmi_level(),
-                                        HMILevel::HMI_FULL,
-                                        HMILevel::HMI_LIMITED)) {
-    HmiStatePtr cur_state = to_resolve->RegularHmiState();
-    if (Compare<HMILevel::eType, EQ, ONE>(cur_state->hmi_level(),
-                                          HMILevel::HMI_FULL,
-                                          HMILevel::HMI_LIMITED)) {
-      if (applied_->IsAudioApplication() && state_ctrl_->IsSameAppType(applied_, to_resolve)) {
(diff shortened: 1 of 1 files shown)


asenkk:
@pestOO , @AGaliuzov , @LuxoftAKutsan please review

AGaliuzov:
Seems fine to me

AGaliuzov:
reviewed

src/components/application_manager/src/state_controller.cc
```diff
@@ -66,29 +66,56 @@ void StateController::HmiLevelConflictResolver::operator ()
...
-        LOG4CXX_DEBUG(logger_, "HMI level will be changed for application with appID="
-                      <<to_resolve->app_id());
-        state_ctrl_->SetupRegularHmiState(to_resolve, HMILevel::HMI_LIMITED,
-                     AudioStreamingState::AUDIBLE);
-      } else {
-        LOG4CXX_DEBUG(logger_, "HMI will not be changed level for application with appID="
-                      <<to_resolve->app_id());
-      }
-    }
+  if (to_resolve == applied_) return;
```
pestOO: Add logging for this return case.
And use brackets according to gstyle

pestOO:
Please, update unittests for this changes/fix

LuxoftAKutsan:
Кумшуцув

src/components/application_manager/src/state_controller.cc
asenkk: This code is copy/paste from CustomSDL/sdl_panasonic/develop(451042b). May be it is better to keep it the same for future comparison.
Also this return is about skipping the current app and updating all others. It is called in a cycle over all apps. This is not really a special case - the purpose is to update the other apps.

AGaliuzov:
@pestOO, I suggest to keep this code as close to the PASA implementation as possible.
Since there is the branch with state controller which should be merged in the nearest feature.
And I want to keep this merged as easy as possible. Also this branch contain all required UT.

PR #318, opened 2016-01-08
Fix endpoints data copying during Policy Table Update

asenkk:
Up to now, when performing Policy Table Update, the module_config part
of the policy table was not copied. The change adds copying of this
part. The code is taken from CustomSDL/sdl_panasonic/develop (dd8ade70),
with just a small fix about copying exchange_after_x_ignition_cycles.

Fixes: APPLINK-20261

### DIFF
--- a/src/components/policy/src/policy/policy_table/table_struct/types.cc
+++ b/src/components/policy/src/policy/policy_table/table_struct/types.cc
@@ -489,6 +489,24 @@ ModuleConfig::ModuleConfig(const Json::Value* value__)
     vehicle_model(impl::ValueMember(value__, "vehicle_model")),
     vehicle_year(impl::ValueMember(value__, "vehicle_year")) {
 }
+
+void ModuleConfig::SafeCopyFrom(const ModuleConfig& from) {
+//  device_certificates = from.device_certificates;  // According to the requirements this is optional.
+  exchange_after_x_ignition_cycles = from.exchange_after_x_ignition_cycles;
+  exchange_after_x_kilometers = from.exchange_after_x_kilometers;
+  exchange_after_x_days = from.exchange_after_x_days;
+  timeout_after_x_seconds = from.timeout_after_x_seconds;
+  seconds_between_retries = from.seconds_between_retries;
+  endpoints = from.endpoints;
+  notifications_per_minute_by_priority = from.notifications_per_minute_by_priority;
+
+  vehicle_make.assign_if_valid(from.vehicle_make);
+  vehicle_model.assign_if_valid(from.vehicle_model);
+  vehicle_year.assign_if_valid(from.vehicle_year);
+  certificate .assign_if_valid(from.certificate);
+
+}
+
 Json::Value ModuleConfig::ToJsonValue() const {
   Json::Value result__(Json::objectValue);
   impl::WriteJsonField("device_certificates", device_certificates, &result__);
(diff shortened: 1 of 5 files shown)


asenkk:
@AGaliuzov , @dev-gh , please review

src/components/policy/src/policy/policy_table/table_struct/types.cc
```diff
@@ -489,6 +489,24 @@ ModuleConfig::ModuleConfig(const Json::Value* value__)
...
+
+void ModuleConfig::SafeCopyFrom(const ModuleConfig& from) {
+//  device_certificates = from.device_certificates;  // According to the requirements this is optional.
+  exchange_after_x_ignition_cycles = from.exchange_after_x_ignition_cycles;
+  exchange_after_x_kilometers = from.exchange_after_x_kilometers;
+  exchange_after_x_days = from.exchange_after_x_days;
+  timeout_after_x_seconds = from.timeout_after_x_seconds;
+  seconds_between_retries = from.seconds_between_retries;
+  endpoints = from.endpoints;
+  notifications_per_minute_by_priority = from.notifications_per_minute_by_priority;
```
pestOO: Could we use an ``opearator=`` witn additional ``FiledVerification`` method call?

src/components/policy/src/policy/policy_table/table_struct/types.cc
```diff
@@ -489,6 +489,24 @@ ModuleConfig::ModuleConfig(const Json::Value* value__)
     vehicle_model(impl::ValueMember(value__, "vehicle_model")),
     vehicle_year(impl::ValueMember(value__, "vehicle_year")) {
 }
+
+void ModuleConfig::SafeCopyFrom(const ModuleConfig& from) {
```
pestOO: Could we use an ``opearator=`` witn additional ``FiledVerification`` method call?

src/components/policy/src/policy/src/cache_manager.cc
```diff
@@ -228,8 +228,13 @@ bool CacheManager::ApplyUpdate(const policy_table::Table& update_pt) {
       pt_->policy_table.app_policies_section.apps[iter->first].set_to_null();
       pt_->policy_table.app_policies_section.apps[iter->first].set_to_string("");
     } else if (policy::kDefaultId == (iter->second).get_string()) {
-      pt_->policy_table.app_policies_section.apps[iter->first] =
-          pt_->policy_table.app_policies_section.apps[kDefaultId];
+        policy_table::ApplicationPolicies::const_iterator iter_default =
```
pestOO: It's really hard to read - "the method of the value of the map of the field of the field of field".
Feel free to crate a local const (or not) reference and use it.

AGaliuzov:
@pestOO, I would suggest to avoid the refactoring some stuff in GENIVI until we done the same for PASA. Because in future in will be more easy to merge this repos if they will have the similar code

src/components/policy/src/policy/policy_table/table_struct/types.cc
AGaliuzov: As for me this is more obvious way.
Because it would be really not evident if the `operator=` would avoid the copy of some fileds.
That's why I ve implemented this method for PASA code instead of overloading `operator=`

dev-gh:
Reviewed

src/components/policy/src/policy/policy_table/table_struct/types.cc
pestOO: ``operator=`` must copy _all_ fields of this strucs.
In case it meet requirements - it's better way to use operator.
If not - agree we this implementation.

asenkk: The idea is not to copy all fields. Some of them MUST be omitted, some are optional according to requirements for policy table update.

AGaliuzov:
Run CI

PR #319, opened 2016-01-11
Add count check for ListFiles request.

OHerasym:
SDL must return 1000 file names.

Related:
        APPLINK-14369


OHerasym:
Please review: @malirod, @VVeremjova , @Kozoriz,  @AMikhniev

AMikhniev:
reviewed

VVeremjova:
Add hotfix in the branch name.

malirod:
@OHerasym, please use real links for the tasks. It will more convenient to just click the link rather than opening site manually. Use `[Name](link)` for this.

pestOO:
Agree iwth @malirod with moving it to the ini file.

VVeremjova:
Perhaps this parameter should not be added to ini but it should be taken from mobile api, where it sets

PR #322, opened 2016-01-11
Fix endless SIGSEGV sending

AGritsevich:
Delete some unusefull blocks of code and add exit with error to sig_handler.
It help to prevent endless SIGSEGV. It need, because after getting SIGSEGV code must
do anything, not just logging.
More information by link [stackoverflow](http://stackoverflow.com/questions/6981325/why-signal-handler-goes-to-infinite-loop-sigsegv)
Closes-bug: [APPLINK-20011](https://adc.luxoft.com/jira/browse/APPLINK-20011).

Please review @anosach-luxoft, @pestOO, @AGaliuzov


pestOO:
Look at the first commit of this signal handling - de19d4af, than was a fix de053ce3 which removed it.
The main idea was to _try_ disable logger and than stop the SDL.
Please investigate a possibility to return these original functionality.

malirod:
I see next issues with the current impl of the signal handling:
there is quite limit amount of instructions which can be safely called in the handler. E.g. `new` cannot be used. This is UB (and "ub" sanitizer can confirm this). I am almost sure that logger uses strings, thus threre is `new`. Thus logger cannot be used. Simple `printf` if the way to go. Also I am not sure about `delete`.

Currently the logic of single handling is split between lifecycle and thread. Life cycle subscribes for the special singnals, and on the thread creations there is the code which unsubscribes from those signals. And this is correct. But I propose to collect all this logic in the single place. E.g. in some new `SignalHanlder` It will created in the lifecycle . It will create the additional thread which will handle those special signals and unsubscribe caller thread (main) from them. All treads created further from main thread will inherit flag ignorance. This will encapsulate the signal handling logic.

malirod:
Here is a book with info regarding signals & threads "Advanced Programming in the UNIX Environment, 3rd Edition" (http://poincare.matf.bg.ac.rs/~ivana//courses/ps/sistemi_knjige/pomocno/apue.pdf) . See "12.8. Threads and Signals". There is the sample with separate thread for the signal handling.

AGritsevich:
Please review @anosach-luxoft, @pestOO, @AGaliuzov, @malirod

AGaliuzov:
@AGritsevich please rebase the PR

AGritsevich:
@AGaliuzov, done!

src/appMain/life_cycle.cc
```diff
@@ -359,27 +358,22 @@ namespace {
         break;
       case SIGSEGV:
         LOG4CXX_DEBUG(logger_, "SIGSEGV signal has been caught");
+#ifdef ENABLE_LOG
+        FLUSH_LOGGER();
+#endif
+//abort need to prevent endless sending SIGSEGV
+//http://stackoverflow.com/questions/2663456/how-to-write-a-signal-handler-to-catch-sigsegv
```
anosach-luxoft: indent

anosach-luxoft: You don't need to wrap this call to `ENABLE_LOG` `ifdef`

anosach-luxoft: Maybe call `DEINIT_LOGGER` too?

src/components/connection_handler/src/connection_handler_impl.cc
```diff
@@ -802,6 +802,7 @@ void ConnectionHandlerImpl::CloseSession(ConnectionHandle connection_handle,
       for (;service_list_itr != service_list.rend(); ++service_list_itr) {
         const protocol_handler::ServiceType service_type =
             service_list_itr->service_type;
+
```
anosach-luxoft: I think it's not good practice to add unnecessary files to pull request because of such changes

src/components/include/utils/logger.h
```diff
@@ -63,7 +63,12 @@
     void deinit_logger();
     #define DEINIT_LOGGER() deinit_logger()
 
-    #define LOG4CXX_IS_TRACE_ENABLED(loggerPtr) loggerPtr->isTraceEnabled()
+    //special macros to relief logs from queue
```
anosach-luxoft: Maybe `dump` or `flush` instead of strange `relief`?

src/components/utils/src/logger.cc
```diff
@@ -51,6 +51,15 @@ void deinit_logger () {
   logger::logger_status = logger::LoggerThreadNotCreated;
 }
 
+//Don't destroy logger here!
+//It's just for unloading logger queue
+void flush_logger() {
+  logger::LoggerStatus snatcher = logger::logger_status;
+  logger::logger_status = logger::DeletingLoggerThread;
```
anosach-luxoft: Are you sure you can change logger status without any synchronization?

src/components/include/utils/threads/message_loop_thread.h
```diff
@@ -110,12 +115,20 @@ class MessageLoopThread {
     Handler& handler_;
     // Message queue that is actually owned by MessageLoopThread
     MessageQueue<Message, Queue>& message_queue_;
+   public:
```
anosach-luxoft: You already have `public` section

src/components/include/utils/threads/message_loop_thread.h
```diff
@@ -156,6 +169,12 @@ void MessageLoopThread<Q>::Shutdown() {
   thread_->join();
 }
 
+template<class Q>
+void MessageLoopThread<Q>::WaitEmptyQueue() {
```
anosach-luxoft: Are you sure we need this overhead functionality for all `MessageLoopThread`s?

AGritsevich:
@malirod, @anosach-luxoft  see 8dfe4fe79bc631337efe76585715dfe16cf45a01

AGritsevich:
@malirod, @anosach-luxoft see 426ff466b6d0ea8fe07a95afbedf193d62da3823

AGritsevich:
@anosach-luxoft, @pestOO, @AGaliuzov, @malirod I've checked it on Head Unit and it's works!

pestOO:
Sounds great!
Have you provided these changes to https://github.com/CustomSDL/sdl_panasonic ?

AGritsevich:
@pestOO Yeah, of cause. I've merged changes to panasonic and test it on Head Unit.

AGritsevich:
Please review @anosach-luxoft, @pestOO, @AGaliuzov, @malirod.

pestOO:
Reviewed

AGritsevich:
Please review @anosach-luxoft, @AGaliuzov, @malirod.

malirod:
Please, rebase first

AGritsevich:
@malirod , see 9ac286b5982011f255d1248431f436e4b4cbe3a9

AGritsevich:
Please review @anosach-luxoft, @AGaliuzov, @malirod.

src/components/include/utils/message_queue.h
```diff
@@ -137,6 +143,13 @@ template<typename T, class Q> void MessageQueue<T, Q>::wait() {
   }
 }
 
+template<typename T, class Q> void MessageQueue<T, Q>::waitUntilEmpty() {
```
anosach-luxoft: @AGritsevich why don't you like method `wait` above?

src/components/utils/src/logger.cc
```diff
@@ -51,6 +51,16 @@ void deinit_logger () {
   logger::logger_status = logger::LoggerThreadNotCreated;
 }
 
+// Don't destroy logger here!
+// It's just for unloading logger queue
+void flush_logger() {
+  logger::LoggerStatus old_status = logger::logger_status;
+  // Stop pushing new messages to the log queue
+  logger::logger_status = logger::DeletingLoggerThread;
```
anosach-luxoft: Why `Deleting` status? We're not going to delete logger

src/appMain/life_cycle.cc
```diff
@@ -393,15 +387,11 @@ void sig_handler(int sig) {
...
-  main_thread = pthread_self();
-  // First, register signal handlers
-  if (!::utils::SubscribeToInterruptSignal(&sig_handler) ||
-      !::utils::SubscribeToTerminateSignal(&sig_handler) ||
-      !::utils::SubscribeToFaultSignal(&sig_handler)) {
-    LOG4CXX_FATAL(logger_, "Subscribe to system signals error");
+  // Register signal handlers and wait sys signals
+  // from another thread
+  if (!utils::WaitTerminationSignals(&sig_handler)) {
+      LOG4CXX_WARN(logger_, "Fail to catch system signal!");
```
anosach-luxoft: It's fatal error but not warn

src/appMain/life_cycle.cc
```diff
@@ -362,30 +362,24 @@ bool LifeCycle::InitMessageSystem() {
...
+        LOG4CXX_DEBUG(logger_, "SIGINT signal has been caught");
+        break;
+      case SIGTERM:
+        LOG4CXX_DEBUG(logger_, "SIGTERM signal has been caught");
+        break;
+      case SIGSEGV:
+        LOG4CXX_DEBUG(logger_, "SIGSEGV signal has been caught");
+        FLUSH_LOGGER();
+        // exit need to prevent endless sending SIGSEGV
+        // http://stackoverflow.com/questions/2663456/how-to-write-a-signal-handler-to-catch-sigsegv
```
anosach-luxoft: Don't we need to `DEINIT_LOGGER` here?

src/appMain/life_cycle.cc
```diff
@@ -393,15 +387,11 @@ void sig_handler(int sig) {
...
 void LifeCycle::Run() {
   LOG4CXX_AUTO_TRACE(logger_);
-  main_thread = pthread_self();
-  // First, register signal handlers
-  if (!::utils::SubscribeToInterruptSignal(&sig_handler) ||
-      !::utils::SubscribeToTerminateSignal(&sig_handler) ||
-      !::utils::SubscribeToFaultSignal(&sig_handler)) {
-    LOG4CXX_FATAL(logger_, "Subscribe to system signals error");
+  // Register signal handlers and wait sys signals
+  // from another thread
```
anosach-luxoft: which 
> another thread

?

src/appMain/main.cc
```diff
@@ -125,6 +125,12 @@ int32_t main(int32_t argc, char** argv) {
     profile::Profile::instance()->config_file_name("smartDeviceLink.ini");
   }
 
+  // Unsibscribe once for all threads
+  if ( utils::UnsibscribeFromTermination() ) {
```
anosach-luxoft: Not clear for me why we have an error in case of successful unsubscription. Please, remove spaces in `if` statement

src/components/utils/src/signals_linux.cc
```diff
@@ -35,33 +35,58 @@
...
+    // Catch possible crashes on exit
+    // after e.g. sigint was caught
+    if (sig != SIGSEGV) {
+      CatchSIGSEGV(sig_handler);
+    }
+    sig_handler(sig);
+    return true;
+  }
+  return false;
 }
```
anosach-luxoft: I propose to do not mask `SIGSEGV` signal at all to avoid all this redundant staff with its handling. I think it's not 
necessarily to handle `SIGSEGV` in main thread (IMHO).
And how `sigwait` with following manual `sig_handler` call better than `sigaction`?

src/components/include/utils/message_queue.h
```diff
@@ -96,6 +96,12 @@ template<typename T, class Q = std::queue<T> > class MessageQueue {
     void wait();
 
     /**
+      * \brief waitUntilEmpty message queue
+      * Wait until message queue is empty
+      */
+    void waitUntilEmpty();
```
anosach-luxoft: Please start method name with capital letter

src/components/include/utils/message_queue.h
AGritsevich: @anosach-luxoft , because there different condition to exit.

src/components/utils/src/logger.cc
AGritsevich: @anosach-luxoft , because only this status dropping new messages and don't create new logger. No any action happens to this status.

src/components/include/utils/message_queue.h
anosach-luxoft: Ok, my mistake

src/appMain/life_cycle.cc
AGritsevich: @anosach-luxoft , I think no, because it abnormal finishing of program. And in the end in any cases all memory will be free without DEINIT_LOGGER.

src/appMain/main.cc
AGritsevich: @anosach-luxoft , because this function return result system API function and return 0 - in succces, and other positive numbers in other cases.

anosach-luxoft: so return reverted result

AGritsevich:
@anosach-luxoft, @malirod  updated, see 3f6dc8900e9efe63df8a8758bf258c755d2006c1

anosach-luxoft:
@AGritsevich I still propose to do not mask `SIGSEGV`

anosach-luxoft:
And one more important question: what happens if we receive signal (`SIGINT`, `SIGTERM` or `SIGSEGV`) between `utils::UnsibscribeFromTermination()` and `utils::WaitTerminationSignals()` calls?

AGritsevich:
@anosach-luxoft ,OK, it make sense. See 093ae6f2012accfaec8dfe67ea530b821bb2bbc8

AGritsevich:
@anosach-luxoft , I'm sorry. In previous commit was mistake. See it 5af7bebf3dbc844fe4816ccf7d3b9a1f9fcffd5c

PR #326, opened 2016-01-14
Converting upper bound values

VVeremjova:
Smart Object now returns 64bit integer instead of 32bit.
Added assignment operator for Json value with 64bit (long long int) numbers.
Added getter for Json value.
Added functions for converting int64 to long long int and back
Removed calls asInt64() as duplicate of getter asInt().

### DIFF
--- a/src/components/formatters/test/formatter_json_rpc_test.cc
+++ b/src/components/formatters/test/formatter_json_rpc_test.cc
@@ -30,6 +30,13 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <algorithm>
+#include "formatters/formatter_json_rpc.h"
+#include <string>
+#include <set>
+#include <algorithm>
+#include <json/writer.h>
+#include "gtest/gtest.h"
 #include "formatters/formatter_json_rpc.h"
 #include <string>
 #include <set>
@@ -40,11 +47,26 @@
 
 namespace test {
 namespace components {
-namespace formatters {
+namespace formatters_test {
 
 using namespace NsSmartDeviceLink::NsSmartObjects;
 using namespace NsSmartDeviceLink::NsJSONHandler::Formatters;
 using namespace NsSmartDeviceLink::NsJSONHandler::strings;
+
+void CompactJson(std::string& str) {
+  Json::Value root;
+  Json::Reader reader;
+  reader.parse(str, root);
+  Json::FastWriter writer;
+  str = writer.write(root);
+  std::string::iterator end_pos = std::remove(str.begin(), str.end(), '\n');
+  str.erase(end_pos, str.end());
+}
+
+namespace {
+const int64_t big_64int = 100000000000;
+const std::string str_with_big_int64 = "100000000000";
+}  // namespace
 
 TEST(FormatterJsonRPCTest, CorrectRPCv1Request_ToString_Success) {
   // Create SmartObject
@@ -62,10 +84,10 @@ TEST(FormatterJsonRPCTest, CorrectRPCv1Request_ToString_Success) {
   std::string result;
   // Convert SmartObject to Json string
   EXPECT_TRUE(FormatterJsonRpc::ToString(obj, result));
-  EXPECT_EQ(std::string(
-                "{\n   \"id\" : 4444,\n   \"jsonrpc\" : \"2.0\",\n   "
-                "\"method\" : \"VR.IsReady\"\n}\n"),
-            result);
+  CompactJson(result);
+  const std::string json_string(
+      "{\"id\":4444,\"jsonrpc\":\"2.0\",\"method\":\"VR.IsReady\"}");
+  EXPECT_EQ(json_string, result);
 }
 
 TEST(FormatterJsonRPCTest, CorrectRPCv2Request_ToString_Success) {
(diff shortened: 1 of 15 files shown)


VVeremjova:
Please review : @malirod @anosach-luxoft

This is the same code for Genivi from [CustomSDL/sdl_panasonic/pull/159](https://github.com/CustomSDL/sdl_panasonic/pull/159)

src/components/formatters/test/formatter_json_rpc_test.cc
```diff
@@ -40,12 +47,27 @@
...
 using namespace NsSmartDeviceLink::NsJSONHandler::Formatters;
 using namespace NsSmartDeviceLink::NsJSONHandler::strings;
 
+void CompactJson(std::string& str) {
+  Json::Value root;
+  Json::Reader reader;
+  reader.parse(str, root);
+  Json::FastWriter writer;
+  str = writer.write(root);
+  std::string::iterator end_pos = std::remove(str.begin(), str.end(), '\n');
```
malirod: This code removes all `\n` from the string. But what if this will be in the json string value? I assume that this will cause the corruption of the original string. This is not good. The logic should be next: just cut off last(search from the end of the string till first `}`) `\n` if exists. But since this code was already merged in the another repo, plus this is only for tests at the moment, lets leave as it is.

src/components/smart_objects/include/smart_objects/number_schema_item.h
```diff
@@ -124,20 +124,32 @@ bool TNumberSchemaItem<NumberType>::isValidNumberType(SmartType type) {
   }
 }
 
-template<typename NumberType>
-Errors::eType TNumberSchemaItem<NumberType>::validate(const SmartObject& Object) {
+template <typename InputType, typename OutputType>
+OutputType IntStaticCast(const InputType& value) {
```
malirod: Since input param is assumed to be int, then there is no need in const&. no need to fix.

src/components/smart_objects/src/smart_object.cc
```diff
@@ -298,6 +290,16 @@ bool SmartObject::operator==(const int64_t Value) const {
 }
 
 // =============================================================
+// uint64_t TYPE SUPPORT
```
malirod: These comments are useless. the type is written in the arg type. No need to fix.

src/components/formatters/test/formatter_json_rpc_test.cc
```diff
@@ -40,12 +47,27 @@
...
 namespace test {
 namespace components {
-namespace formatters {
+namespace formatters_test {
 
 using namespace NsSmartDeviceLink::NsSmartObjects;
 using namespace NsSmartDeviceLink::NsJSONHandler::Formatters;
 using namespace NsSmartDeviceLink::NsJSONHandler::strings;
 
+void CompactJson(std::string& str) {
```
malirod: As far as I remember @pestOO, asked to caver this function with test too. I see no test.

src/components/smart_objects/include/smart_objects/number_schema_item.h
malirod: Why this is not in the convert utils?

anosach-luxoft: I think name like `SafeStaticCast` would be better for this function because it could be used not only with `int` type

VVeremjova: @malirod See 47f917e

src/components/smart_objects/src/smart_object.cc
VVeremjova: @malirod These comments are in the whole file, should I remove only this or other too?

malirod: I think you can delete all useless commits.

src/components/formatters/test/formatter_json_rpc_test.cc
VVeremjova: @malirod  See 52ec3c0

src/components/formatters/test/formatter_json_rpc_test.cc
VVeremjova: @malirod Changed in 52ec3c0

src/components/formatters/test/formatter_json_rpc_test.cc
```diff
@@ -40,12 +47,47 @@
...
+  Json::Value root;
+  Json::Reader reader;
+  reader.parse(str, root);
+  Json::FastWriter writer;
+  str = writer.write(root);
+  if (str[str.size() - 1] == '\n') {
+      str.erase(str.size()-1,1);
+  }
+  std::string::iterator end_pos = std::remove(str.begin(), str.end(), '\n');
```
malirod: Why you need these two lines? They remove all endl from string. We have to remove only the last.

src/components/smart_objects/src/smart_object.cc
VVeremjova: @malirod Deleted in 51d8e73

src/components/formatters/test/formatter_json_rpc_test.cc
VVeremjova: @malirod fixed it in 8c60c0c

malirod:
Reviewed

anosach-luxoft:
Reviewed

PR #328, opened 2016-01-14
Fixes issue with RAI execution after connection is dropped

dev-gh:
There is possible rare case, when connection is dropped right after RAI
is received by SDL, but not started yet. SDL has to drop that request,
otherwise app will register, but won't have any valid connection and
will prevent same app re-registration from valid connections.

Fixes: APPLINK-20521


AGaliuzov:
ok to test

PR #332, opened 2016-01-18
Fix sending Policy Table snapshot in OnSystemRequest

asenkk:
Before this change OnSystemRequest notification with PT snapshot and an
update URL, sent by SDL to mobile when the mobile app registers, was
actually empty. The reason was that when sending the notification we
were trying to read PT data from a non existing file, while the actual
data was discarded.
This change fixes the problem by removing reading from file and using
the already built PT data in memory.
It also fixes sending the PTU url in the message (it was also missing).

Fixes: APPLINK-17788

### DIFF
--- a/src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
+++ b/src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
@@ -76,12 +76,10 @@ void OnSystemRequestNotification::Run() {
   }
 
   if (RequestType::PROPRIETARY == request_type) {
-  std::string filename =
-      (*message_)[strings::msg_params][strings::file_name].asString();
-
-  std::vector<uint8_t> binary_data;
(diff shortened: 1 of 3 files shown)


asenkk:
@AGaliuzov , @pestOO , @dev-gh please review

src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
```diff
@@ -76,12 +76,18 @@ void OnSystemRequestNotification::Run() {
   }
 
   if (RequestType::PROPRIETARY == request_type) {
-  std::string filename =
+    /* According to requirements:
```
AGaliuzov: If the implemented behaviour is according to the requirements just remove it this disabled code below

AGaliuzov:
Reviewed

dev-gh:
Reviewed

PR #337, opened 2016-01-19
Remove DCHECKs that seem to always be tripping everyone up

### LINKED ISSUE #167: DCHECK in release/4.0.0 crashes core
In message_helper.cc, the DCHECK() calls at lines [1305 and 1307](https://github.com/smartdevicelink/sdl_core/blob/release/4.0.0/src/components/application_manager/src/message_helper.cc#L1305) crash core on an assert. Whether core crashes at line 1305, or crashes on 1307 depends on which tester app I am using.

If I comment out both of those lines and recompile Core, everything works fine. 

Why is DCHECK being used in the release branch?

justinjdickow:
Fixes #167

### DIFF
--- a/src/components/application_manager/src/message_helper.cc
+++ b/src/components/application_manager/src/message_helper.cc
@@ -1267,9 +1267,7 @@ bool MessageHelper::CreateHMIApplicationStruct(ApplicationConstSharedPtr app,
   }
 
   const SmartObject* app_types = app->app_types();
-  DCHECK_OR_RETURN(app_types, false);
   const SmartObject* ngn_media_screen_name = app->ngn_media_screen_name();
-  DCHECK_OR_RETURN(ngn_media_screen_name, false);
   const connection_handler::DeviceHandle handle = app->device();
   std::string device_name = ApplicationManagerImpl::instance()->GetDeviceName(handle);
 


salengo:
Hello All,

I am new in SDL. I was testing the last "mobileWeather" version and I faced the problem with the DCHECK. I realized that the problem appears when the proxy is created without app_types or media_screen_name. The problem was solved using other constructor in the class "SmartDeviceLinkService" in method startProxy. You can change line :

`proxy = new SyncProxyALM(this, "MobileWeather", true, "330533107");
`

By more complex constructor, like:

```
boolean isMediaApp = true;
Vector<AppHMIType> appType = new Vector<AppHMIType>();
appType.add(AppHMIType.DEFAULT);
proxy = new  SdlProxyALM(this, null, "MobileWeather", null, null, null,  isMediaApp, 
						null, mDesiredAppSdlLanguage, DesiredAppHmiLanguage, appType, 
						"330533107",null, false,false, transport);
```
This change works for me, I can use the SDL_Core version 4.0.2 with the DCHECK with no problem. Other solution could be to change the class SdlProxyALM in the android library to not assign null value by default to appType or media screen name if this is not a correct value for SDL_Core.

PR #346, opened 2016-01-22
Fix validity of Policy Table Snapshot

asenkk:
Up to now policy_table.consumer_friendly_messages->messages were sent
in PTS, which is wrong according to the requirements. Removed them.
Also policy_table.usage_and_error_counts->app_level was invalid.
Fixed this by putting default values to the uninitialized fields.
Now PTS is valid.

Fixes: APPLINK-17684


asenkk:
@dev-gh , @AGaliuzov please review

dev-gh:
Reviewed

asenkk:
@pestOO, this code is taken from CustomSDL/sdl_panasonic develop and in general we don't want to change it in order to facilitate merging.

PR #354, opened 2016-01-25
Fix OnHMIStatusNotificationFromMobile command creation

asenkk:
There was s bug in MobileCommandFactory::CreateCommand(), which was
causing always to create OnHMIStatusNotification command, not
OnHMIStatusNotificationFromMobile, even when the origin of the command
is Command::ORIGIN_MOBILE. Fixing this allows SDLv4 QUERY_APPS to
be handled correctly, in particilar to gray them out when the launcher
app goes in BACKGROUND by sending UpdateAppList with greyOut parameter.

Fixes: APPLINK-17866


asenkk:
@dev-gh , @AGaliuzov , please review

dev-gh:
Reviewed

PR #360, opened 2016-01-27
SDL sends OnHMIStatus(FULL) to non-media app only after end Phone Call if app resumes during active phone call 

dev-gh:
In case default HMI level of app is the same as level assigned during phone call, than no notification is sent to mobile, which is wrong.
With this fix changes sequence of assigning of active states in state controller during application registration, which allow to send notification to mobile side anyway.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 build/
+*.user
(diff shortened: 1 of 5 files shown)


dev-gh:
@dtrunov @AGaliuzov 
Please review

dtrunov:
Reviewed

PR #365, opened 2016-02-01
Fixes double sending of GetUserFriendlyMessage in case of absent message

dev-gh:
In case of absent message in policy DB SDL was returning response twice
- one time without 'messages' and another time with empty 'messages'.

Closes-bug: APPLINK-20985


dev-gh:
@AGaliuzov 
Please review

PR #368, opened 2016-02-02
Fix merging updated PreloadedPT into existing database

asenkk:
There was a problem when the preloaded PT is changed between SDL
restarts - tables wer not merged. Added call to MergePreloadPT() to
fix this. But even without changing PPT, after SDL restart we had
errors about PT validity. The reason was that in app_policies tables
we had invalid fields (according to Policy Table Data Dictionary).
Removed them and fixed validation.

Fixes: APPLINK-17753

### DIFF
--- a/src/components/policy/src/policy/policy_table/table_struct/types.cc
+++ b/src/components/policy/src/policy/policy_table/table_struct/types.cc
@@ -26,107 +26,55 @@ namespace policy_table_interface_base {
   PolicyBase::PolicyBase()
     : CompositeType(kUninitialized) {
   }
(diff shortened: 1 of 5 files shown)


asenkk:
@AGaliuzov , @dev-gh please review

dev-gh:
I think it would be better if you can split changes for two commits next time since there were two issues.
Reviewed.

asenkk:
@dev-gh , the two issues are connected - without fixing the second issue, the first one (for which we have a ticket) can't be fixed. So the second one is more like a sub-issue. But in general I agree with you.

Issue #374, opened 2016-02-03
IviInfoUpdated Only Returns Navi Applications


JackLivio:
I was testing interior vehicle data notifications and I can only receive vehicle data notifications in SPT if the HMI type is set to NAVIGATION.

[This is the function in question] (https://github.com/smartdevicelink/sdl_core/blob/develop/src/components/application_manager/src/application_manager_impl.cc#L310).

[IviInfoUpdated will only return apps that are of type "navi"](https://github.com/smartdevicelink/sdl_core/blob/develop/src/components/application_manager/src/application_manager_impl.cc#L324). Instead, shouldn't this function be looking for apps subscribed to vehicle data?

PR #375, opened 2016-02-03
Fixed IviInfoUpdated App Query

JackLivio:
The query in IviInfoUpdated was only returing connected navigation apps. It now returns apps that are subscribed to vehicle info.


JackLivio:
#374

PR #382, opened 2016-02-09
Fix return info in AlertManeuver, AlertManeuver, ShowConstantTBT

OHerasym:
Fix return info response in AlertManeuver
Add return info in ReadDID response
Add return info in ShowConstantTBT response
Related:
[APPLINK-14569](https://adc.luxoft.com/jira/browse/APPLINK-14569)

### DIFF
--- a/src/components/application_manager/src/commands/mobile/read_did_request.cc
+++ b/src/components/application_manager/src/commands/mobile/read_did_request.cc
@@ -1,6 +1,5 @@
 /*
-
- Copyright (c) 2013, Ford Motor Company
+ Copyright (c) 2016, Ford Motor Company
  All rights reserved.
 
  Redistribution and use in source and binary forms, with or without
@@ -31,6 +30,7 @@
  POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include <string>
 #include "application_manager/commands/mobile/read_did_request.h"
 #include "application_manager/application_manager_impl.h"
 #include "application_manager/application_impl.h"
@@ -101,7 +101,11 @@ void ReadDIDRequest::on_event(const event_engine::Event& event) {
 
       bool result = mobile_apis::Result::SUCCESS == result_code;
 
-      SendResponse(result, result_code, NULL, &(message[strings::msg_params]));
+      const std::string return_info =
+          message[strings::msg_params][hmi_response::message].asString();
+
+      SendResponse(result, result_code, return_info.c_str(),
+                   &(message[strings::msg_params]));
       break;
     }
     default: {
--- a/src/components/application_manager/src/commands/mobile/show_constant_tbt_request.cc
+++ b/src/components/application_manager/src/commands/mobile/show_constant_tbt_request.cc
@@ -31,7 +31,7 @@
  POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include <string.h>
+#include <cstring>
 #include "application_manager/commands/mobile/show_constant_tbt_request.h"
 #include "application_manager/application_manager_impl.h"
 #include "application_manager/application_impl.h"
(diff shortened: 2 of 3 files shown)


OHerasym:
Please review: @malirod, @anosach-luxoft , @AByzhynar , @VVeremjova , @Kozoriz, @AMikhniev

AByzhynar:
Reviewed

AMikhniev:
Reviewed

Kozoriz:
reviewed

anosach-luxoft:
Reviewed

src/components/application_manager/src/commands/mobile/show_constant_tbt_request.cc
```diff
@@ -181,6 +181,7 @@ void ShowConstantTBTRequest::on_event(const event_engine::Event& event) {
   switch (event.id()) {
     case hmi_apis::FunctionID::Navigation_ShowConstantTBT: {
       LOG4CXX_INFO(logger_, "Received Navigation_ShowConstantTBT event");
+      const char* return_info = NULL;
```
anosach-luxoft: @OHerasym please use `std::string` here too

src/components/application_manager/src/commands/mobile/read_did_request.cc
```diff
@@ -101,7 +101,11 @@ void ReadDIDRequest::on_event(const event_engine::Event& event) {
 
       bool result = mobile_apis::Result::SUCCESS == result_code;
 
-      SendResponse(result, result_code, NULL, &(message[strings::msg_params]));
+      std::string return_info =
```
malirod: @OHerasym, in this case return_info  can be const

src/components/application_manager/src/commands/mobile/show_constant_tbt_request.cc
malirod: @OHerasym, don't be afraid to use string be value. Even non-const. By the way, assigning something to the **const** pointer looks very strange.

OHerasym: @malirod , info must be changed only in success case, otherwise NULL, if there will std::string, there will be additional check, so why I can't use const char*?

malirod: @OHerasym,
- assignment to `const char* return_info` looks wring. remove const at least
- why you need brackets around the `message[strings::msg_params][hmi_response::message].asString()`?

OHerasym: @malirod , f21540d, added const, deleted brackets. cant change const char* to char* , because c_str() return const char*.

anosach-luxoft: @OHerasym I think it would be better to use `std::string` as I proposed before and check it for empty before passing to response. For example:
```
SendResponse(result, result_code, return_info.empty() ? 0 : return_info.c_str(), &(message[strings::msg_params]));
```
It's bad idea to store raw pointer to `std::string::c_str()` beacause string could be deleted any time after some logic changes and you'll get invalid pointer

malirod: @anosach-luxoft, actually I like the existing approach. Your variant with additional string allocation, ternary operator looks worse to me. The pointer to internals doesn't looks to scary to me since we know for sure that the massage will not change. I don't like most the assignment to the const pointer. But I can leave with this fact.

OHerasym:
@anosach-luxoft , @malirod , bff0216 , Changed const char* to std::string

OHerasym:
@malirod , 5dff0c9 , Deleted empty string

malirod:
Reviewed

PR #383, opened 2016-02-09
Increase Slider mobile request timeout to receive HMI response

asenkk:
Until now we were setting generic command timeout for Slider request
equal to Slider's own timeout immediatelly after receiving the command.
Then we send the command to HMI and wait for HMI response. If we don't
do anything in HMI it sends TIMED_OUT response to SDL, but this response
comes AFTER the generic command timeout has expired (because it was
started a litle bit earlier). The generic timeout triggers GENERIC_ERROR
response and HMI's TIMED_OUT response is ignored.
The fix is to set generic command timeout bigger that slider timeout.

Fixes: APPLINK-20738

### DIFF
--- a/src/components/application_manager/src/commands/mobile/slider_request.cc
+++ b/src/components/application_manager/src/commands/mobile/slider_request.cc
@@ -54,8 +54,12 @@ bool SliderRequest::Init() {
   /* Timeout in milliseconds.
      If omitted a standard value of 10000 milliseconds is used.*/
   if ((*message_)[strings::msg_params].keyExists(strings::timeout)) {
-    default_timeout_ =
-        (*message_)[strings::msg_params][strings::timeout].asUInt();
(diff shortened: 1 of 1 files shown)


asenkk:
@AGaliuzov , @dev-gh please review

asenkk:
@AGaliuzov, we depend on an external component - the HMI. We can't predict what delays are there. The only thing we can do is to give it some arbitrary, but reasonable extra time to compensate the inevitable communication and processing delays (above the command specific delay).

dev-gh:
As fix this is okay, but I would propose to discuss that with requirement team than.
Reviewed.

asenkk:
@dev-gh , created requirements question : APPLINK-21499 "Requirements advice needed about HMI communication timeout period".

asenkk:
Identical to https://github.com/CustomSDL/sdl_panasonic/pull/214

Issue #391, opened 2016-02-15
OnButtonEventNotification::Run() Skips 'OK' Button Event Notification 


JackLivio:
[OnButtonEventNotification::Run() Skips 'OK' Button Event Notification if the app is in full]
(https://github.com/smartdevicelink/sdl_core/blob/master/src/components/application_manager/src/commands/mobile/on_button_event_notification.cc#L118).

Proposed fix:

```
    if ((static_cast<uint32_t>(mobile_apis::ButtonName::OK) == btn_id) &&
        (!subscribed_app->IsFullscreen())) { //If app is NOT in fullscreen, continue
      continue;
    }
```

PR #392, opened 2016-02-15
Add NOT to checking if app is in full for OK button notification

JackLivio:
Issue #391


anosach-luxoft:
@JackLivio are you sure you're following actual requirements with your fix?

JackLivio:
@joeygrover @justinjdickow Is there a requirement we can site that shows this is the correct behavior? This issue came up when trying to pause/play music on Pandora's mobile app. Core was not constructing on button event notifications for the OK button when the app was in full due to previously implemented logic, which i assume was a typo.

Take the comment above the if statement into consideration.

```
//Send ButtonEvent notification for OK button only in HMI_FULL mode
```

Issue #401, opened 2016-03-01
PolicyHandler::SendMessageToSDK using unimplemented container


JackLivio:
In the Master Branch, [PolicyHandler::SendMessageToSDK](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/application_manager/src/policies/policy_handler.cc#L728-L763) is using an unimplemented container 'last_used_app_ids_'. 

SendMessageToSDK() will always return false since last_used_app_ids_ is always empty. 

[The develop branch uses GetAppIdForSending() instead of 'last_used_app_ids_'](https://github.com/smartdevicelink/sdl_core/blob/develop/src/components/application_manager/src/policies/policy_handler.cc#L749). I think master hasn't been updated with these changes yet, but this is a blocking issue for some developers at the moment and needs to be updated asap.

madhuy:
need this change as well. other url been not sent to Mobile https://github.com/smartdevicelink/sdl_core/blob/develop/src/components/application_manager/src/message_helper/message_helper.cc#L2093

FStoikov:
Jira ticket created for current case: https://adc.luxoft.com/jira/browse/FORDTCN-2867

FStoikov:
Jira ticket created for current case: https://adc.luxoft.com/jira/browse/FORDTCN-2867

PR #404, opened 2016-03-03
Fix for sending policy table snapshot (Genivi implementation)

JackLivio:
Policy handler was checking an unimplemented contained last_used_app_ids_. Replaced with 'GetAppIdForSending() to properly find app id to send policy snapshot to.

Also in message_helper.cc, I fixed the url paremeter to what actually corresponds to onSystemRequests in the mobile_api.xml


JackLivio:
#401

PR #425, opened 2016-03-22
Added validator in processing of SystemRequest(QUERY_APPS)

okoreniak:
Added class QueryAppsDataValidatorTemplate<SmartObject> to validate request data.

PR #433, opened 2016-03-23
Hotfix/fixed validator in system request query apps processing

okoreniak:
Please review:
@Anatoliy-Leshin
@AByzhynar
@LuxoftAKutsan
@malirod
@AGaliuzov
@dev-gh

[Previous request](https://github.com/smartdevicelink/sdl_core/pull/425)
[APPLINK-17836](https://adc.luxoft.com/jira/browse/APPLINK-17836)

### DIFF
--- a/src/components/application_manager/src/commands/mobile/system_request.cc
+++ b/src/components/application_manager/src/commands/mobile/system_request.cc
@@ -33,7 +33,8 @@ Copyright (c) 2013, Ford Motor Company
 
 #include <vector>
 #include <string>
-#include <stdio.h>
+#include <map>
+#include <set>
 #include "application_manager/commands/mobile/system_request.h"
 #include "application_manager/application_manager_impl.h"
 #include "application_manager/application_impl.h"
@@ -43,8 +44,349 @@ Copyright (c) 2013, Ford Motor Company
 #include "utils/file_system.h"
 #include "formatters/CFormatterJsonBase.hpp"
 #include "json/json.h"
+#include "utils/helpers.h"
 
 namespace application_manager {
+
+namespace {
+
+CREATE_LOGGERPTR_LOCAL(logger_, "ApplicationManager")
+
+const char* kQueryAppsValidationFailedPrefix =
+    ":QUERY_APPS_VALIDATION_FAILED: ";
+
+const unsigned int kVrSynonymLengthMax = 40U;
+const unsigned int kVrSynonymLengthMin = 1U;
+const unsigned int kTtsNameLengthMax = 500U;
+const unsigned int kVrArraySizeMax = 100U;
+const unsigned int kVrArraySizeMin = 1U;
+const unsigned int kUrlSchemaLengthMax = 255U;
+const unsigned int kPackageNameLengthMax = 255U;
+const unsigned int kAppIdLengthMax = 40U;
+const unsigned int kAppNameLengthMax = 100U;
+const unsigned int kLanguageArraySizeMax = 100U;
+
+class QueryAppsDataValidator {
+ public:
+  typedef std::set<std::string> SynonymsSet;
+  typedef std::map<std::string, SynonymsSet> SynonymsMap;
+
+  QueryAppsDataValidator(const smart_objects::SmartObject& object,
+                         const ApplicationManagerImpl& manager)
+      : data_(object), manager_(manager) {}
+
+  bool Validate() const {
+    LOG4CXX_AUTO_TRACE(logger_);
+    if (!data_.isValid()) {
+      LOG4CXX_ERROR(logger_,
+                    kQueryAppsValidationFailedPrefix
+                        << "QueryApps response is not valid.");
+      return false;
+    }
+    if (!HasResponseKey()) {
+      return false;
+    }
+    return ValidateAppDataAndOsAndLanguagesData();
+  }
+
+ private:
+  bool HasResponseKey() const {
+    if (!data_.keyExists(json::response)) {
+      LOG4CXX_WARN(logger_,
+                   kQueryAppsValidationFailedPrefix
+                       << "QueryApps response does not contain '"
+                       << json::response << "' parameter.");
+      return false;
+    }
+    return true;
+  }
+
+  bool ValidateAppDataAndOsAndLanguagesData() const {
+    const smart_objects::SmartArray* objects_array =
+        data_[json::response].asArray();
+    if (!objects_array) {
+      LOG4CXX_WARN(logger_,
+                   kQueryAppsValidationFailedPrefix
+                       << "QueryApps response is not array.");
+      return false;
+    }
+    const std::size_t arr_size(objects_array->size());
+    SynonymsMap synonyms_map;
+    for (std::size_t idx = 0; idx < arr_size; ++idx) {
+      const smart_objects::SmartObject& app_data = (*objects_array)[idx];
+
+      if (!app_data.isValid()) {
+        LOG4CXX_WARN(logger_,
+                     kQueryAppsValidationFailedPrefix
+                         << "Wrong application data in json file.");
+        return false;
+      }
+      std::set<std::string> app_ids_set;
+      if (!ValidateAppIdAndAppName(app_data, app_ids_set)) {
+        return false;
+      }
+      // Verify os and dependent languages data
+      std::string os_type;
+      if (app_data.keyExists(json::ios)) {
+        os_type = json::ios;
+        if (!app_data[os_type].keyExists(json::urlScheme)) {
+          LOG4CXX_WARN(logger_,
+                       kQueryAppsValidationFailedPrefix
+                           << "Can't find URL scheme in json file.");
+          return false;
+        }
+        if (app_data[os_type][json::urlScheme].asString().length() >
+            kUrlSchemaLengthMax) {
+          LOG4CXX_WARN(
+              logger_,
+              kQueryAppsValidationFailedPrefix
+                  << "An urlscheme length exceeds maximum allowed ["
+                  << app_data[os_type][json::urlScheme].asString().length()
+                  << "]>[" << kUrlSchemaLengthMax << "]");
+          return false;
+        }
+      }
+      if (os_type.empty()) {
+        if (app_data.keyExists(json::android)) {
+          os_type = json::android;
+          if (!app_data[os_type].keyExists(json::packageName)) {
+            LOG4CXX_WARN(logger_,
+                         kQueryAppsValidationFailedPrefix
+                             << "Can't find package name in json file.");
+            return false;
+          }
+          if (app_data[json::android][json::packageName].asString().length() >
+              kPackageNameLengthMax) {
+            LOG4CXX_WARN(logger_,
+                         kQueryAppsValidationFailedPrefix
+                             << "Package name length ["
+                             << app_data[json::android][json::packageName]
+                                    .asString()
+                                    .length() << "] exceeds max length ["
+                             << kPackageNameLengthMax << "]in json file.");
+            return false;
+          }
+        }
+      }
+
+      if (os_type.empty()) {
+        LOG4CXX_WARN(logger_,
+                     kQueryAppsValidationFailedPrefix
+                         << "Can't find mobile OS type in json file.");
+        return false;
+      }
+
+      // Languages verification
+      if (!app_data[os_type].keyExists(json::languages)) {
+        LOG4CXX_WARN(logger_,
+                     kQueryAppsValidationFailedPrefix
+                         << "'languages' doesn't exist");
+        return false;
+      }
+      if (!ValidateLanguages(app_data[os_type][json::languages],
+                             synonyms_map)) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  bool ValidateAppIdAndAppName(const smart_objects::SmartObject& app_data,
+                               std::set<std::string>& app_ids_set) const {
+    // Verify appid
+    if (!app_data.keyExists(json::appId)) {
+      LOG4CXX_WARN(logger_,
+                   kQueryAppsValidationFailedPrefix
+                       << "Can't find app ID in json file.");
+      return false;
+    }
+    // Verify appid length
+    const std::string app_id(app_data[json::appId].asString());
+    if (app_id.length() > kAppIdLengthMax) {
+      LOG4CXX_WARN(logger_,
+                   kQueryAppsValidationFailedPrefix
+                       << "An Object ID length exceeds maximum allowed ["
+                       << app_id.length() << "]>[" << kAppIdLengthMax << "]");
+      return false;
+    }
+
+    // Verify that appid is unique
+    if (app_ids_set.find(app_id) != app_ids_set.end()) {
+      LOG4CXX_WARN(logger_,
+                   kQueryAppsValidationFailedPrefix
+                       << "An Object ID is not unigue [" << app_id << "]");
+      return false;
+    }
+    app_ids_set.insert(app_id);
+
+    // Verify that app is not registered yet
+    ApplicationSharedPtr registered_app =
+        manager_.application_by_policy_id(app_id);
+    if (registered_app) {
+      LOG4CXX_WARN(logger_,
+                   kQueryAppsValidationFailedPrefix
+                       << "Application with the same id: " << app_id
+                       << " is registered already.");
+      return false;
+    }
+    // Verify app name exist
+    if (!app_data.keyExists(json::name)) {
+      LOG4CXX_WARN(logger_,
+                   kQueryAppsValidationFailedPrefix
+                       << "Can't find app name in json file.");
+      return false;
+    }
+    // And app name length
+    const std::string appName(app_data[json::name].asString());
+    if (appName.length() > kAppNameLengthMax) {
+      LOG4CXX_WARN(logger_,
+                   kQueryAppsValidationFailedPrefix
+                       << "Name of application exceeds maximum allowed ["
+                       << appName.length() << "]>[" << kAppNameLengthMax
+                       << "].");
+      return false;
+    }
+    return true;
+  }
+
+  bool ValidateLanguages(const smart_objects::SmartObject& languages,
+                         SynonymsMap& synonyms_map) const {
+    bool default_language_found = false;
+    const size_t languages_array_size = languages.length();
+    if (languages_array_size > kLanguageArraySizeMax) {
+      LOG4CXX_WARN(logger_,
+                   kQueryAppsValidationFailedPrefix
+                       << "'languages' array exceeds max size ["
+                       << languages_array_size << "]>[" << kLanguageArraySizeMax
+                       << "]");
+      return false;
+    }
+    // Every language has ttsname string and vrsynonyms array
+    for (size_t idx = 0; idx < languages_array_size; ++idx) {
+      const smart_objects::SmartObject& language = languages.getElement(idx);
+      std::set<std::string> keys = language.enumerate();
+      for (std::set<std::string>::const_iterator iter = keys.begin();
+           iter != keys.end();
+           ++iter) {
+        // Verify default language defined
+        if (!(*iter).compare(json::default_)) {
+          default_language_found = true;
+        }
+        // Add set for synonyms' duplicates validation
+        if (synonyms_map.find(*iter) == synonyms_map.end()) {
+          synonyms_map[*iter] = SynonymsSet();
+        }
+        // ttsName verification
+        if (!language[*iter].keyExists(json::ttsName)) {
+          LOG4CXX_WARN(logger_,
+                       kQueryAppsValidationFailedPrefix
+                           << "'languages.ttsName' doesn't exist");
+          return false;
+        }
+        const smart_objects::SmartObject& ttsNameObject =
+            language[*iter][json::ttsName];
+        // ttsName is string
+        if (smart_objects::SmartType_String == ttsNameObject.getType()) {
+          const std::string ttsName = language[*iter][json::ttsName].asString();
+          if (ttsName.length() > kTtsNameLengthMax) {
+            LOG4CXX_WARN(logger_,
+                         kQueryAppsValidationFailedPrefix
+                             << "ttsName string exceeds max length ["
+                             << ttsName.length() << "]>[" << kTtsNameLengthMax
+                             << "]");
+            return false;
+          }
+        } else {
+          LOG4CXX_WARN(logger_,
+                       kQueryAppsValidationFailedPrefix
+                           << "ttsName is not the string type.");
+          return false;
+        }
+        if (!language[*iter].keyExists(json::vrSynonyms)) {
+          LOG4CXX_WARN(logger_,
+                       kQueryAppsValidationFailedPrefix
+                           << "'languages.vrSynonyms' doesn't exist");
+          return false;
+        }
+        const smart_objects::SmartArray* synonyms_array =
+            language[*iter][json::vrSynonyms].asArray();
+        if (!synonyms_array) {
+          LOG4CXX_WARN(logger_,
+                       kQueryAppsValidationFailedPrefix
+                           << "vrSynonyms is not array.");
+          return false;
+        }
+        const size_t synonyms_array_size = synonyms_array->size();
+        if (synonyms_array_size < kVrArraySizeMin) {
+          LOG4CXX_WARN(logger_,
+                       kQueryAppsValidationFailedPrefix
+                           << "vrSynomyms array has [" << synonyms_array_size
+                           << "] size < allowed min size [" << kVrArraySizeMin
+                           << "]");
+          return false;
+        }
+        if (synonyms_array_size > kVrArraySizeMax) {
+          LOG4CXX_WARN(logger_,
+                       kQueryAppsValidationFailedPrefix
+                           << "vrSynomyms array size [" << synonyms_array_size
+                           << "] exceeds maximum allowed size ["
+                           << kVrArraySizeMax << "]");
+          return false;
+        }
+        for (std::size_t idx = 0; idx < synonyms_array_size; ++idx) {
+          const smart_objects::SmartObject& synonym = (*synonyms_array)[idx];
+          const std::string vrSynonym = synonym.asString();
+          if (vrSynonym.length() > kVrSynonymLengthMax) {
+            LOG4CXX_WARN(logger_,
+                         kQueryAppsValidationFailedPrefix
+                             << "vrSYnomym item [" << idx
+                             << "] exceeds max length [" << vrSynonym.length()
+                             << "]>[" << kVrSynonymLengthMax << "]");
+            return false;
+          }
+          if (vrSynonym.length() < kVrSynonymLengthMin) {
+            LOG4CXX_WARN(logger_,
+                         kQueryAppsValidationFailedPrefix
+                             << "vrSYnomym item [" << idx << "] length ["
+                             << vrSynonym.length()
+                             << "] is less then min length ["
+                             << kVrSynonymLengthMin << "] allowed.");
+            return false;
+          }
+          // Verify duplicates
+          SynonymsMap::iterator synonyms_map_iter = synonyms_map.find(*iter);
+          SynonymsSet* synonyms_set = &(synonyms_map_iter->second);
+          if (synonyms_map_iter != synonyms_map.end()) {
+            if ((*synonyms_map_iter).second.find(vrSynonym) ==
+                synonyms_set->end()) {
+              synonyms_set->insert(vrSynonym);
+            } else {
+              LOG4CXX_WARN(logger_,
+                           kQueryAppsValidationFailedPrefix
+                               << "vrSYnomym item already defined ["
+                               << vrSynonym.c_str() << "] for language ["
+                               << *iter << "]");
+              return false;
+            }
+          }
+        }
+      }
+    }
+    if (!default_language_found) {
+      LOG4CXX_WARN(logger_,
+                   kQueryAppsValidationFailedPrefix
+                       << " 'languages'.default' doesn't exist");
+      return false;
+    }
+    return true;
+  }
+
+  const smart_objects::SmartObject& data_;
+  const ApplicationManagerImpl& manager_;
+
+  DISALLOW_COPY_AND_ASSIGN(QueryAppsDataValidator);
+};
+}
 
 namespace commands {
 
@@ -54,11 +396,9 @@ const std::string kSYNC = "SYNC";
 const std::string kIVSU = "IVSU";
 
 SystemRequest::SystemRequest(const MessageSharedPtr& message)
-    : CommandRequestImpl(message) {
-}
+    : CommandRequestImpl(message) {}
 
-SystemRequest::~SystemRequest() {
-}
+SystemRequest::~SystemRequest() {}
 
 void SystemRequest::Run() {
   LOG4CXX_AUTO_TRACE(logger_);
@@ -77,7 +417,7 @@ void SystemRequest::Run() {
           (*message_)[strings::msg_params][strings::request_type].asInt());
 
   if (!policy::PolicyHandler::instance()->IsRequestTypeAllowed(
-           application->mobile_app_id(), request_type)) {
+          application->mobile_app_id(), request_type)) {
     SendResponse(false, mobile_apis::Result::DISALLOWED);
     return;
   }
@@ -89,15 +429,14 @@ void SystemRequest::Run() {
     file_name = kSYNC;
   }
 
-  bool is_system_file =
-      std::string::npos != file_name.find(kSYNC) ||
-      std::string::npos != file_name.find(kIVSU);
+  bool is_system_file = std::string::npos != file_name.find(kSYNC) ||
+                        std::string::npos != file_name.find(kIVSU);
 
   // to avoid override existing file
   if (is_system_file) {
     const uint8_t max_size = 255;
     char buf[max_size] = {'\0'};
-    snprintf(buf, sizeof(buf)/sizeof(buf[0]), "%d%s", index++, file_name.c_str());
+    snprintf(buf, max_size - 1, "%d%s", index++, file_name.c_str());
     file_name = buf;
   }
 
@@ -118,9 +457,10 @@ void SystemRequest::Run() {
   file_dst_path += file_name;
 
   if ((*message_)[strings::params].keyExists(strings::binary_data)) {
-    LOG4CXX_DEBUG(logger_, "Binary data is present. Trying to save it to: "
-                  << binary_data_folder);
-    if (mobile_apis::Result::SUCCESS  !=
+    LOG4CXX_DEBUG(
+        logger_,
+        "Binary data is present. Trying to save it to: " << binary_data_folder);
+    if (mobile_apis::Result::SUCCESS !=
         (ApplicationManagerImpl::instance()->SaveBinary(
             binary_data, binary_data_folder, file_name, 0))) {
       LOG4CXX_DEBUG(logger_, "Binary data can't be saved.");
@@ -131,16 +471,28 @@ void SystemRequest::Run() {
     std::string app_full_file_path = binary_data_folder;
     app_full_file_path += file_name;
 
-    LOG4CXX_DEBUG(logger_, "Binary data is not present. Trying to find file "
-                  << file_name << " within previously saved app file in "
-                  << binary_data_folder);
+    LOG4CXX_DEBUG(logger_,
+                  "Binary data is not present. Trying to find file "
+                      << file_name << " within previously saved app file in "
+                      << binary_data_folder);
 
     const AppFile* file = application->GetFile(app_full_file_path);
     if (!file || !file->is_download_complete ||
         !file_system::MoveFile(app_full_file_path, file_dst_path)) {
       LOG4CXX_DEBUG(logger_, "Binary data not found.");
-      SendResponse(false, mobile_apis::Result::REJECTED);
-      return;
+
+      std::string origin_file_name;
+      if ((*message_)[strings::msg_params].keyExists(strings::file_name)) {
+        origin_file_name =
+            (*message_)[strings::msg_params][strings::file_name].asString();
+      }
+      if (!(mobile_apis::RequestType::HTTP == request_type &&
+            0 == origin_file_name.compare(kIVSU))) {
+        LOG4CXX_DEBUG(logger_, "Binary data required. Reject");
+        SendResponse(false, mobile_apis::Result::REJECTED);
+        return;
+      }
+      LOG4CXX_DEBUG(logger_, "IVSU does not require binary data. Continue");
     }
     processing_file_ = file_dst_path;
   }
@@ -160,9 +512,8 @@ void SystemRequest::Run() {
     }
 
     CFormatterJsonBase::jsonValueToObj(root, sm_object);
-
     if (!ValidateQueryAppData(sm_object)) {
-      SendResponse(false, mobile_apis::Result::INVALID_DATA);
+      SendResponse(false, mobile_apis::Result::GENERIC_ERROR);
       return;
     }
 
@@ -172,10 +523,10 @@ void SystemRequest::Run() {
     return;
   }
 
-  smart_objects::SmartObject msg_params = smart_objects::SmartObject(
-      smart_objects::SmartType_Map);
+  smart_objects::SmartObject msg_params =
+      smart_objects::SmartObject(smart_objects::SmartType_Map);
   if (std::string::npos != file_name.find(kIVSU)) {
-    msg_params[strings::file_name] = file_name.c_str();
+    msg_params[strings::file_name] = file_name;
   } else {
     msg_params[strings::file_name] = file_dst_path;
   }
@@ -186,12 +537,14 @@ void SystemRequest::Run() {
   msg_params[strings::request_type] =
       (*message_)[strings::msg_params][strings::request_type];
   SendHMIRequest(hmi_apis::FunctionID::BasicCommunication_SystemRequest,
-                 &msg_params, true);
-
+                 &msg_params,
+                 true);
 }
 
 void SystemRequest::on_event(const event_engine::Event& event) {
-  LOG4CXX_INFO(logger_, "AddSubMenuRequest::on_event");
+  LOG4CXX_AUTO_TRACE(logger_);
+  using namespace helpers;
+
   const smart_objects::SmartObject& message = event.smart_object();
 
   switch (event.id()) {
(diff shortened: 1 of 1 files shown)


src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+        continue;
+      }
+
+      // Languages verification
+      if (!app_data[os_type].keyExists(json::languages)) {
+        LOG4CXX_WARN(logger_, error_prefix_ << "\"languages\" not exists");
+        result = false;
+        continue;
+      }
+      if (!ValidateLanduages(app_data[os_type][json::languages]))
```
dev-gh: @okoreniak 
Typo in method name

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+          LOG4CXX_WARN(logger_,
+                       error_prefix_
+                           << "\"languages\".\"\ttsName\" doesn't exist");
+          result = false;
+        } else {
+          const smart_objects::SmartObject& ttsNameObject =
+              language[*iter][json::ttsName];
+          // ttsName is string
+          if (smart_objects::SmartType_String == ttsNameObject.getType()) {
+            std::string ttsName =
```
AByzhynar: @okoreniak Shouldn't be const?

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+static const char* kQeeryAppsValidationFailedPrefix =
+    ":QUERY_APPS_VALIDATION_FAILED: ";
+
+template <typename ObjectType>
+class QueryAppsDataValidatorTemplate {
+ public:
+  typedef std::set<std::string> SynonymsSetType;
+  typedef std::map<std::string, SynonymsSetType> SynonymsMapType;
+
+  QueryAppsDataValidatorTemplate(const ObjectType& o)
```
AByzhynar: @okoreniak We have to avoid using such unclear names like `o`. But in this particular case may be it's normal. I'm not sure.

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +57,366 @@ uint32_t SystemRequest::index = 0;
...
+};
+
+template <typename ObjectType>
+class QueryAppsDataValidatorTemplate {
+ public:
+  typedef std::set<utils::custom_string::CustomString, comparer>
+      synonyms_set_type;
+  typedef std::map<std::string, synonyms_set_type> synonyms_map_type;
+
+  QueryAppsDataValidatorTemplate(const ObjectType& o,
```
AByzhynar: @okoreniak We have to avoid using such unclear names like o. But in this particular case may be it's normal. I'm not sure.

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +57,366 @@ uint32_t SystemRequest::index = 0;
...
+                                 const char* error_message = "warning")
+      : data_(o)
+      , error_message_(error_message)
+      , max_vr_synonym_length_(40)
+      , min_vr_synonym_length_(1)
+      , max_tts_name_length_(500)
+      , max_vr_array_size_(100) {}
+
+  bool Validate() const {
```
AByzhynar: @okoreniak May be move all methods definitions out of the class? Class interface will be more readable.

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -299,8 +289,8 @@ class QueryAppsDataValidatorTemplate {
               language[*iter][json::ttsName];
           // ttsName is string
           if (smart_objects::SmartType_String == ttsNameObject.getType()) {
-            utils::custom_string::CustomString ttsName =
-                language[*iter][json::ttsName].asCustomString();
+            std::string ttsName =
```
AByzhynar: @okoreniak Why not const?

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -346,8 +336,8 @@ class QueryAppsDataValidatorTemplate {
             }
             for (std::size_t idx = 0; idx < syn_size; ++idx) {
               const smart_objects::SmartObject& synonym = (*syn_array)[idx];
-              const utils::custom_string::CustomString vrSynonym =
-                  synonym.asCustomString();
+              std::string vrSynonym =
```
AByzhynar: @okoreniak The same as above. Can you make it `const`? Or you are changing its value afterwards?

okoreniak: Fixed. [81c7cf2](https://github.com/smartdevicelink/sdl_core/pull/433/commits/81c7cf2d0e1ee43ef5fa0f443165853d1bbe6484)

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: Fixed [b4b20ec](https://github.com/smartdevicelink/sdl_core/pull/433/commits/b4b20ec4191ad5a819381529befa59bf2fa24917)

dev-gh:
Reviewed

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: Fixed [81c7cf2d0e](https://github.com/smartdevicelink/sdl_core/pull/433/commits/81c7cf2d0e1ee43ef5fa0f443165853d1bbe6484)

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: It's a template, so methods defined in the template definition.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: Fixed all std::string variables to be const. [81c7cf2d0e1](https://github.com/smartdevicelink/sdl_core/pull/433/commits/81c7cf2d0e1ee43ef5fa0f443165853d1bbe6484)

AByzhynar:
Reviewed

Anatoliy-Leshin:
Sorry, have no time to review today and tomorrow.

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -224,63 +592,23 @@ void SystemRequest::on_event(const event_engine::Event& event) {
...
-    if (!app_data.keyExists(json::name)) {
-      LOG4CXX_ERROR(logger_, "Can't find app name in json file.");
-      continue;
-    }
-
-    return true;
-  }
 
-  return false;
+  QueryAppsDataValidatorTemplate<smart_objects::SmartObject> validator(data);
```
Kozoriz: @okoreniak , for what template there is? We can validate only SmartObjects. Or no?

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+      }
+
+      // Languages verification
+      if (!app_data[os_type].keyExists(json::languages)) {
+        LOG4CXX_WARN(logger_, error_prefix_ << "\"languages\" not exists");
+        result = false;
+        continue;
+      }
+      if (!ValidateLanguages(app_data[os_type][json::languages]))
+        result = false;
```
Kozoriz: @okoreniak , can we change this to : `result = ValidateLanguages(app_data[os_type][json::languages])` ??

Kozoriz: or to :` result = result && ValidateLanguages(app_data[os_type][json::languages])`

src/components/application_manager/src/commands/mobile/system_request.cc
VVeremjova: I also don't see any reason to use template class

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+    }
+
+    // Verify that appid is unique
+    if (app_ids_map_.find(app_id) != app_ids_map_.end()) {
+      LOG4CXX_WARN(logger_,
+                   error_prefix_ << "An Object ID is not unigue [" << app_id
+                                 << "]");
+      result = false;
+    } else {
+      app_ids_map_[app_id] = 1;
```
VVeremjova: Why do you set 1? Why do you need map if you use only `app_id`?

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: Template is good in case when you use a simple class in one place without splitting it to header and implementation, that doesn't produce new files.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: no, it'll brake logic cause result is cumulative after all verifications, for example ValidateLanguages() may return true for some data and produce error.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: Actually std::set maybe more suitable in this case with little changes in code. Current implementation uses map so code is valid for this case.

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +57,366 @@ uint32_t SystemRequest::index = 0;
 const std::string kSYNC = "SYNC";
 const std::string kIVSU = "IVSU";
 
+#define QUERY_APPS_VALIDATION_FAILED ":QUERY_APPS_VALIDATION_FAILED:"
+
+struct comparer {
```
AByzhynar: @okoreniak Type names start with a capital letter and have a capital letter for each new word, with no underscores: MyExcitingClass, MyExcitingEnum.

okoreniak:
7 days passed.
@Anatoliy-Leshin @LuxoftAKutsan @malirod @AGaliuzov - please review to finish request/task.

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
 const std::string kSYNC = "SYNC";
 const std::string kIVSU = "IVSU";
 
+static const char* kQeeryAppsValidationFailedPrefix =
+    ":QUERY_APPS_VALIDATION_FAILED: ";
+
+template <typename ObjectType>
+class QueryAppsDataValidatorTemplate {
```
malirod: @okoreniak, on my opinion the "Template" suffix is redundant.

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+                                  << json::response << "' parameter.");
+      return false;
+    }
+    return true;
+  }
+
+  bool ValidateAppDataAndOsAndLanguagesData() const {
+    const smart_objects::SmartArray* obj_array =
+        data_[json::response].asArray();
+    if (NULL == obj_array) {
```
malirod: @okoreniak, this can be just `if(obj array)`

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+      LOG4CXX_ERROR(logger_,
+                    error_prefix_ << "QueryApps response is not array.");
+      return false;
+    }
+    bool result = true;
+    const std::size_t arr_size(obj_array->size());
```
malirod: @okoreniak, why such non-usual init of the integral var?

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+    for (std::size_t idx = 0; idx < arr_size; ++idx) {
+      const smart_objects::SmartObject& app_data = (*obj_array)[idx];
+
+      if (!app_data.isValid()) {
+        LOG4CXX_WARN(logger_,
+                     error_prefix_ << "Wrong application data in json file.");
+        result = false;
```
malirod: @okoreniak, can we just return false from the loop? Why we need this result var? Do we need to proceed with further checks if we have already false?

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+    bool result = true;
+    // Verify appid lenght
+    const std::string app_id(app_data[json::appId].asString());
+    if (app_id.length() > max_app_id_length_) {
+      LOG4CXX_WARN(logger_,
+                   error_prefix_
+                       << "An Object ID length exceeds maximum allowed ["
+                       << app_id.length() << "]>[" << max_app_id_length_
+                       << "]");
+      result = false;
```
malirod: @okoreniak, same case as above. Why to just return false immediately.

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+    }
+
+    // Verify that appid is unique
```
malirod: @okoreniak, as you mentioned below the `set` is more suitable. So I propose to use it.
Also I propose to pass `app_ids_map_` as method param. And create it before the loop on the caller side. There is only one place where this method is used, so I don't see the point the private member. Also this will make redundant the call of clear. 
By the way, rename it. Give more meaningful name.E.g. `validated_app_ids` or `processed_app_ids` or something better. The name should make the comment on line 217 redundant(so it will be removed).

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+    }
+
+    // Verify that app is not registered yet
+    ApplicationSharedPtr registered_app =
+        ApplicationManagerImpl::instance()->application_by_policy_id(app_id);
+    if (registered_app) {
+      LOG4CXX_DEBUG(logger_,
+                    error_prefix_ << "Application with the same id: " << app_id
+                                  << " is registered already.");
+      result = false;
```
malirod: @okoreniak, am I right that this param is optional, that is why we don't return and log just debug?

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+    // And app name length
+    const std::string appName(
+        app_data[json::name].asString());
+    if (appName.length() > max_app_name_length_) {
+      LOG4CXX_WARN(logger_,
+                   error_prefix_
+                       << "Name of application exceeds maximum allowed ["
+                       << appName.length() << "]>[" << max_app_name_length_
+                       << "].");
+      result = false;
```
malirod: @okoreniakm why not just to return?

malirod: @okoreniak, imagine there will be additional checks later on. According the check before we should break validation, but current code assumes that this is last check in the method..

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+      std::set<std::string> keys = language.enumerate();
+      for (std::set<std::string>::const_iterator iter = keys.begin();
+           iter != keys.end();
+           ++iter) {
+        // Verify default language defined
+        if (!(*iter).compare(json::default_)) {
+          default_language_found = true;
+        }
+        // Add set for synonyms' duplicates validation
+        if (synonyms_map_.find(*iter) == synonyms_map_.end())
```
malirod: @okoreniak, same case as with `app_ids_map_`. Give the better name and pass as arg instead of private member.

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+
+        if (!language[*iter].keyExists(json::vrSynonyms)) {
+          LOG4CXX_WARN(logger_,
+                       error_prefix_
+                           << "\"languages\".\"\vrSynonyms\" doesn't exist");
+          result = false;
+        } else {
+          const smart_objects::SmartArray* syn_array =
+              language[*iter][json::vrSynonyms].asArray();
+          if (NULL == syn_array) {
```
malirod: @okoreniak, please why full words. `syn_array` -> `synonyms_array`, `syn_size ` -> `synonyms_size`

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+            LOG4CXX_WARN(logger_, error_prefix_ << "vrSynonyms is not array.");
+            result = false;
+          } else {
+            const size_t syn_size = syn_array->size();
+            if (syn_size < min_vr_array_size_) {
+              LOG4CXX_WARN(logger_,
+                           error_prefix_ << "vrSynomyms array has [" << syn_size
+                                         << "] size < allowed min size ["
+                                         << min_vr_array_size_ << "]");
+              result = false;
```
malirod: @okoreniak, why we need the result var? Why not to just return?

src/components/application_manager/src/commands/mobile/system_request.cc
malirod: @okoreniak, since this validator can work with SmartObject **only**, I see no reason in template. I see no harm in additional file and splitting declaration and impl. Also templates makes more pressure on the compiler. So I propose to turn into simple class.

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+                                           << "] length [" << vrSynonym.length()
+                                           << "] is less then min length ["
+                                           << min_vr_synonym_length_
+                                           << "] allowed.");
+                result = false;
+              }
+              // Verify duplicates
+              SynonymsMapType::iterator synonyms_map_iter =
+                  synonyms_map_.find(*iter);
+              DCHECK(synonyms_map_iter != synonyms_map_.end());
```
malirod: @okoreniak, why there is dcheck? Why not just to return false?

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
 const std::string kSYNC = "SYNC";
 const std::string kIVSU = "IVSU";
 
+static const char* kQeeryAppsValidationFailedPrefix =
+    ":QUERY_APPS_VALIDATION_FAILED: ";
+
+template <typename ObjectType>
+class QueryAppsDataValidatorTemplate {
+ public:
```
malirod: @okoreniak, since this class is used only in this cc file then do next
- extract all consts and the validator itsellf to the anon namespace
- I don't see the point in `error_prefix_` member. Turn the kQeeryAppsValidationFailedPrefix  into the const string (same way as it's done for other consts here) and use it directly.
- double check how the private consts should be named. can we make those vars static and init them where they are declared?

anosach-luxoft:
@okoreniak please link JIRA ticket

anosach-luxoft:
@okoreniak are you sure this changes should be delivered to **master** branch?

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +57,366 @@ uint32_t SystemRequest::index = 0;
...
+                  const utils::custom_string::CustomString y) {
+    return x.AsMBString().compare(y.AsMBString()) < 0;
+  }
+};
+
+template <typename ObjectType>
+class QueryAppsDataValidatorTemplate {
+ public:
+  typedef std::set<utils::custom_string::CustomString, comparer>
+      synonyms_set_type;
```
LuxoftAKutsan: Please remove word type from type name and make type CamelCase (according google style)

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +57,366 @@ uint32_t SystemRequest::index = 0;
...
+
+struct comparer {
+ public:
+  bool operator()(const utils::custom_string::CustomString x,
+                  const utils::custom_string::CustomString y) {
+    return x.AsMBString().compare(y.AsMBString()) < 0;
+  }
+};
+
+template <typename ObjectType>
```
LuxoftAKutsan: @okoreniak  I can't see why template is required. Please remove it.

src/components/application_manager/src/commands/mobile/system_request.cc
LuxoftAKutsan: @okoreniak  Please reread googcle coding style. Types should started from capital letter

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +57,366 @@ uint32_t SystemRequest::index = 0;
 const std::string kSYNC = "SYNC";
 const std::string kIVSU = "IVSU";
 
+#define QUERY_APPS_VALIDATION_FAILED ":QUERY_APPS_VALIDATION_FAILED:"
```
LuxoftAKutsan: @okoreniak use  static string except github.

src/components/application_manager/src/commands/mobile/system_request.cc
LuxoftAKutsan: @okoreniak  remove template.

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +57,366 @@ uint32_t SystemRequest::index = 0;
...
+
+  bool Validate() const {
+    if (!data_.isValid()) {
+      LOG4CXX_ERROR(logger_,
+                    QUERY_APPS_VALIDATION_FAILED
+                        << "QueryApps response is not valid.");
+      return false;
+    }
+    if (!hasResponseKey())
+      return false;
```
LuxoftAKutsan: @okoreniak  google coding style.

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +57,366 @@ uint32_t SystemRequest::index = 0;
...
+    if (app_ids_map_.size())
+      app_ids_map_.clear();
+    if (synonyms_map_.size())
+      synonyms_map_.clear();
+
+    return validateAppDataAndOsAndLanguagesData();
+  }
+
+ private:
+  bool hasResponseKey() const {
```
LuxoftAKutsan: @okoreniak  Methods should be started from capital letter.

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +57,366 @@ uint32_t SystemRequest::index = 0;
...
+                        << "QueryApps response does not contain '"
+                        << json::response << "' parameter.");
+      return false;
+    }
+    return true;
+  }
+
+  bool validateAppDataAndOsAndLanguagesData() const {
+    smart_objects::SmartArray* obj_array = data_[json::response].asArray();
+    if (NULL == obj_array) {
```
LuxoftAKutsan: @okoreniak i would prefer   `if (!obj_array) `

okoreniak:
@anosach-luxoft 
- added link to issue
- Galiuzov  asked to send it to master branch [here](https://github.com/smartdevicelink/sdl_core/pull/425#issuecomment-199864326)

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +57,366 @@ uint32_t SystemRequest::index = 0;
...
+    if (!app_data.keyExists(json::appId)) {
+      LOG4CXX_WARN(logger_,
+                   QUERY_APPS_VALIDATION_FAILED
+                       << "Can't find app ID in json file.");
+      return false;
+    }
+
+    // verify appid lenght
+    const std::string app_id(app_data[json::appId].asString());
+    if (app_id.length() > 100) {
```
LuxoftAKutsan: @okoreniak  magic number

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +57,366 @@ uint32_t SystemRequest::index = 0;
...
+    }
+
+    // verify that appid is unique
+    if (app_ids_map_.find(app_id) != app_ids_map_.end()) {
+      LOG4CXX_WARN(logger_,
+                   QUERY_APPS_VALIDATION_FAILED
+                       << "An Object ID is not unigue [" << app_id << "]");
+      return false;
+    } else {
+      app_ids_map_[app_id] = 1;
```
LuxoftAKutsan: @okoreniak  can't understand logic here

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +57,366 @@ uint32_t SystemRequest::index = 0;
...
+    }
+
+    // verify that app is not registered yet
+    ApplicationSharedPtr registered_app =
+        ApplicationManagerImpl::instance()->application_by_policy_id(app_id);
```
LuxoftAKutsan: Please try to avoid singleton Usage. 
You can pass ApplicationManagerImpl& as parameter to validators constructor and hold it.

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +57,366 @@ uint32_t SystemRequest::index = 0;
...
+    if (!app_data.keyExists(json::name)) {
+      LOG4CXX_WARN(logger_,
+                   QUERY_APPS_VALIDATION_FAILED
+                       << "Can't find app name in json file.");
+      return false;
+    }
+    // and app name length
+    const custom_str::CustomString appName(
+        app_data[json::name].asCustomString());
+    if (appName.length() > 100) {
```
LuxoftAKutsan: @okoreniak  Please avoid all magik numbers

src/components/application_manager/src/commands/mobile/system_request.cc
LuxoftAKutsan: @okoreniak . I agree with guys, please remove template, extract this class to separate file ( in application manger folder and add unit test for it.

LuxoftAKutsan:
@Anatoliy-Leshin
@AByzhynar
@LuxoftAKutsan
@malirod
@AGaliuzov
@dev-gh

I propose to squash all commits and repeat review if this changes, because it is hard for me to review it. @okoreniak  you can add no updated review only people whi didn't wrote  "reviewed" comment

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: it's a valid way to initialize an integral values. when it's used without parameters it guaranties that variables get default values, for example int i(); guaranties that i will be initialized by 0 value.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: There wasn't clear requirement about the validation strategy. I used approach with a deep validation.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: Validating class tries to validate as deep as it is able.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: agree.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: Validating code tries to validate as much as it may.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @malirod  agree.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @malirod  agree

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @malirod  Algorithm tries to validate as much as it may.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @malirod  This dcheck guaranties that code wasn't broken and iterator is valid.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @malirod  I agree.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: ok

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: ok

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: obsolete code, please ignore for now.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @LuxoftAKutsan  it's obsolete code, please ignore it.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @LuxoftAKutsan  obsolete code, please ignore

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @LuxoftAKutsan  obsolete code, please ignore.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @LuxoftAKutsan  obsolete code, number moved to private variables.

src/components/application_manager/src/commands/mobile/system_request.cc
malirod: @okoreniak, I still don't get the point in this dcheck. Why there are no similar dchecs in other places?

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @LuxoftAKutsan  map is used to keep app_ids. if map find() didn't find id than else part inserts app_id as key in map and associates 1 with it.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @LuxoftAKutsan ok

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @LuxoftAKutsan all magic numbers were moved to private variables. [528f0bd9e015bd](https://github.com/smartdevicelink/sdl_core/pull/433/commits/528f0bd9e015bd074e5879a6513d0f3302bcef9d)

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @malirod  This is only one place where this dcheck may abort application in case code was updated in some unpredicted way. So if it happens than code should be verified on latest changes to fix this dcheck issue.

src/components/application_manager/src/commands/mobile/system_request.cc
AByzhynar: @malirod Perhaps you did mean `if(!obj_array)`

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+
+  bool ValidateAppIdAndAppName(
+      const smart_objects::SmartObject& app_data) const {
+    // Verify appid
+    if (!app_data.keyExists(json::appId)) {
+      LOG4CXX_WARN(logger_, error_prefix_ << "Can't find app ID in json file.");
+      return false;
+    }
+    bool result = true;
+    // Verify appid lenght
```
AByzhynar: @okoreniak typo

Kozoriz:
@okoreniak , please make this pull faster, because this is very important pull which closes 2 defects for current SOW. Need to merge this pull TODAY.

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -53,12 +56,360 @@ uint32_t SystemRequest::index = 0;
...
+};
+
 SystemRequest::SystemRequest(const MessageSharedPtr& message)
-    : CommandRequestImpl(message) {
-}
+    : CommandRequestImpl(message) {}
 
-SystemRequest::~SystemRequest() {
-}
+SystemRequest::~SystemRequest() {}
```
AByzhynar: @okoreniak What do we need this empty destructor for?

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @LuxoftAKutsan  done

okoreniak:
All notices mainly fixed and commited. Commits squashed.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @AByzhynar  It's legacy code

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -33,6 +33,8 @@ Copyright (c) 2013, Ford Motor Company
 
 #include <vector>
 #include <string>
+#include <map>
+#include <set>
 #include <stdio.h>
```
LuxoftAKutsan: @okoreniak  c headers should be before c++ headers

okoreniak:
@Anatoliy-Leshin @LuxoftAKutsan @malirod @AGaliuzov All notices mainly fixed and committed. Commits squashed. Please review.

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -43,9 +45,356 @@ Copyright (c) 2013, Ford Motor Company
...
 #include "json/json.h"
+#include "utils/helpers.h"
 
 namespace application_manager {
 
+namespace {
+
+CREATE_LOGGERPTR_LOCAL(logger_, "ApplicationManager")
+
+using namespace application_manager::strings;
```
LuxoftAKutsan: @okoreniak  I think it is very bad idea to add this using. There to many redundant names. Also agccording GStyle using is not allowed

src/components/application_manager/src/commands/mobile/system_request.cc
LuxoftAKutsan: @okoreniak  Please don't forget to add name of person whom yo address your message.

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -43,9 +45,356 @@ Copyright (c) 2013, Ford Motor Company
...
+                   error_prefix_ << "An Object ID is not unigue [" << app_id
+                                 << "]");
+      result = false;
+    } else {
+      app_ids_set.insert(app_id);
+    }
+
+    // Verify that app is not registered yet
+    ApplicationSharedPtr registered_app =
+        ApplicationManagerImpl::instance()->application_by_policy_id(app_id);
```
LuxoftAKutsan: @okoreniak  Please try to avoid singleton usage, You can Pass ApplicationManagerImpl as parameter to constructor of this class

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -43,9 +45,356 @@ Copyright (c) 2013, Ford Motor Company
...
+                         SynonymsMap& synonyms_map) const {
+    bool result = true;
+    bool default_language_found = false;
+    const size_t languages_array_size = languages.length();
+    if (languages_array_size > max_languages_array_size_) {
+      LOG4CXX_WARN(logger_,
+                   error_prefix_ << "\"languages\" array exceeds max size ["
+                                 << languages_array_size << "]>["
+                                 << max_languages_array_size_ << "]");
+      return result = false;
```
LuxoftAKutsan: @okoreniak  can't figure out why you write
```
return result = false;
```
but not 
```
return false;
```
 ?

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -43,9 +45,356 @@ Copyright (c) 2013, Ford Motor Company
...
+          default_language_found = true;
+        }
+        // Add set for synonyms' duplicates validation
+        if (synonyms_map.find(*iter) == synonyms_map.end())
+          synonyms_map[*iter] = SynonymsSet();
+        // ttsName verification
+        if (!language[*iter].keyExists(json::ttsName)) {
+          LOG4CXX_WARN(logger_,
+                       error_prefix_
+                           << "\"languages\".\"\ttsName\" doesn't exist");
```
LuxoftAKutsan: @okoreniak  You can use single quotes inside logs to avoid screening

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @LuxoftAKutsan  leftover after updating code to return false instead of result. Fixed for next commit.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @LuxoftAKutsan  Fixed for next commit.

LuxoftAKutsan: @okoreniak in future please insert hash of commit in answer

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @LuxoftAKutsan  Fixed for new commit.

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @LuxoftAKutsan  Fixed [a33b42bf98326](https://github.com/smartdevicelink/sdl_core/pull/433/commits/a33b42bf98326c09a7b4de558687c820bdca150b)

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @LuxoftAKutsan  Fixed [a33b42bf98326c09a7](https://github.com/smartdevicelink/sdl_core/pull/433/commits/a33b42bf98326c09a7b4de558687c820bdca150b)

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @LuxoftAKutsan  Fixed. [a33b42bf98326c09a7b](https://github.com/smartdevicelink/sdl_core/pull/433/commits/a33b42bf98326c09a7b4de558687c820bdca150b)

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @LuxoftAKutsan  Fixed [a33b42bf98326c09a7](https://github.com/smartdevicelink/sdl_core/pull/433/commits/a33b42bf98326c09a7b4de558687c820bdca150b)

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @LuxoftAKutsan Fixed [a33b42bf98326c09a](https://github.com/smartdevicelink/sdl_core/pull/433/commits/a33b42bf98326c09a7b4de558687c820bdca150b)

LuxoftAKutsan:
Reviewed. 
Please don't forget to squash commit with fixes before merge

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -224,63 +598,30 @@ void SystemRequest::on_event(const event_engine::Event& event) {
...
-    }
-
-    if (!app_data.keyExists(json::name)) {
-      LOG4CXX_ERROR(logger_, "Can't find app name in json file.");
-      continue;
-    }
+  ApplicationManagerImpl* manager = ApplicationManagerImpl::instance();
 
-    return true;
+  if (!manager) {
```
malirod: @okoreniak, I think dcheck here is much better.

malirod:
Reviewed

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -43,9 +44,350 @@ Copyright (c) 2013, Ford Motor Company
...
+                             << vrSynonym.length()
+                             << "] is less then min length ["
+                             << kVrSynonymLengthMin << "] allowed.");
+            return false;
+          }
+          // Verify duplicates
+          SynonymsMap::iterator synonyms_map_iter = synonyms_map.find(*iter);
+          SynonymsSet* synonyms_set = &(synonyms_map_iter->second);
+          if (synonyms_map_iter != synonyms_map.end()) {
+            if ((*synonyms_map_iter).second.find(vrSynonym) ==
```
anosach-luxoft: @okoreniak, you don't have to check set for key existence. In case of key already present in set `insert()` method returns false

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -43,9 +44,350 @@ Copyright (c) 2013, Ford Motor Company
...
+                   kQueryAppsValidationFailedPrefix
+                       << "Name of application exceeds maximum allowed ["
+                       << appName.length() << "]>[" << kAppNameLengthMax
+                       << "].");
+      return false;
+    }
+    return true;
+  }
+
+  bool ValidateLanguages(const smart_objects::SmartObject& languages,
```
anosach-luxoft: @okoreniak this method looks too terrible. Is it possible to break it to few separate methods?

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @anosach-luxoft Fixed. [c3a6aa3b](https://github.com/smartdevicelink/sdl_core/pull/433/commits/c3a6aa3b334a4b975eedf6e21251916085118d24)

src/components/application_manager/src/commands/mobile/system_request.cc
okoreniak: @anosach-luxoft Fixed. [c3a6aa3b3](https://github.com/smartdevicelink/sdl_core/pull/433/commits/c3a6aa3b334a4b975eedf6e21251916085118d24)

PR #453, opened 2016-04-01
Feature/add_implementation_WayPoints

OHerasym:
Add implementation SubscribeWayPoints
Add implementation UnsubscribeWayPoints
Add implementation GetWayPoints
Add implementation OnWayPointsChange notification

CRQ: [SubscribeWayPoints](https://adc.luxoft.com/jira/browse/APPLINK-21629)
[UnsubscribeWayPoints](https://adc.luxoft.com/jira/browse/APPLINK-21641)
[GetWayPoints](https://adc.luxoft.com/jira/browse/APPLINK-21610)
[OnWayPointsChange](https://adc.luxoft.com/jira/browse/APPLINK-21351)

### DIFF
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -2500,11 +2500,17 @@ void ApplicationManagerImpl::UnregisterApplication(
                             << is_resuming
                             << "; is_unexpected_disconnect = "
                             << is_unexpected_disconnect);
+  size_t subscribed_for_way_points_app_count = 0;
 
   // SDL sends UnsubscribeWayPoints only for last application
-  if (subscribed_way_points_apps_list_.size() == 1) {
+  {
+    sync_primitives::AutoLock lock(subscribed_way_points_apps_lock_);
+    subscribed_for_way_points_app_count =
+      subscribed_way_points_apps_list_.size();
+  }
+  if (1 == subscribed_for_way_points_app_count) {
     LOG4CXX_ERROR(logger_, "Send UnsubscribeWayPoints");
-    MessageHelper::SendUnSubscribedWayPoints();
+    MessageHelper::SendUnsubscribedWayPoints();
   }
 
   NaviServiceStatusMap::iterator it = navi_service_status_.find(app_id);
@@ -3466,6 +3472,7 @@ void ApplicationManagerImpl::ClearTTSGlobalPropertiesList() {
 bool ApplicationManagerImpl::IsAppSubscribedForWayPoints(
     const uint32_t app_id) const {
   LOG4CXX_AUTO_TRACE(logger_);
+  sync_primitives::AutoLock lock(subscribed_way_points_apps_lock_);
   if (subscribed_way_points_apps_list_.find(app_id) ==
       subscribed_way_points_apps_list_.end()) {
     return false;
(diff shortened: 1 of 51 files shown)


src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -3447,6 +3453,42 @@ void ApplicationManagerImpl::ClearTTSGlobalPropertiesList() {
...
+
+void ApplicationManagerImpl::UnsubscribeAppFromWayPoints(const uint32_t app_id) {
+  LOG4CXX_AUTO_TRACE(logger_);
+  sync_primitives::AutoLock lock(subscribed_way_points_apps_lock_);
+  subscribed_way_points_apps_list_.erase(app_id);
+}
+
+bool ApplicationManagerImpl::IsAnyAppSubscribedForWayPoints() const {
+  LOG4CXX_AUTO_TRACE(logger_);
+  if (subscribed_way_points_apps_list_.empty()) {
```
anosach-luxoft: @OHerasym just `return !subscribed_way_points_apps_list_.empty()` :smile:

AByzhynar:
Reviewed. but please squash commits into some less than 11.

AGaliuzov:
@OHerasym you have listed 4 CRQ with one pull request.
For me it makes more sense to have 4 small pull request instead

OHerasym:
@AGaliuzov , yes, but all CRQs are related. In Subscribe implementation I need Unsubscribe, and for OnWayPointChange needs GetWayPoints and Subscribe.

OHerasym:
@anosach-luxoft , 771fcbb, Added autolock & copyrights in headers

src/components/application_manager/src/commands/mobile/subscribe_way_points_request.cc
```diff
@@ -0,0 +1,75 @@
...
+  if (application_manager::ApplicationManagerImpl::instance()
+          ->IsAppSubscribedForWayPoints(app->app_id())) {
+    SendResponse(false, mobile_apis::Result::IGNORED);
+    return;
+  }
+
+  if (application_manager::ApplicationManagerImpl::instance()
+          ->IsAnyAppSubscribedForWayPoints()) {
+    application_manager::ApplicationManagerImpl::instance()
+        ->SubscribeAppForWayPoints(app->app_id());
```
AGaliuzov: I was thinking about the ability to have the subscirprions with the application itself.
Right now I can see that you got to app manager to get an application and then go to the appmanager again to get the subscribtions which looks like overhead for me

src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -2500,6 +2500,18 @@ void ApplicationManagerImpl::UnregisterApplication(
                             << is_resuming
                             << "; is_unexpected_disconnect = "
                             << is_unexpected_disconnect);
+  bool send_unsubscribed_way_points;
```
anosach-luxoft: @OHerasym you can just do next way:

```
size_t size = 0;
{
    sync_primitives::AutoLock lock(subscribed_way_points_apps_lock_);
    size = subscribed_way_points_apps_list_.size();
}
if (1 == size) {
    LOG4CXX_ERROR(logger_, "Send UnsubscribeWayPoints");
    MessageHelper::SendUnsubscribedWayPoints();
}
```

I think it would be better

src/components/application_manager/src/commands/mobile/subscribe_way_points_request.cc
OHerasym: @AGaliuzov , why it's overhead? for example we need send Unsubscribe only for last application, in case when application instance will have subscriptions, we need to check all applications each time, when we need send notifications, we again need check all applications.

src/components/application_manager/src/application_manager_impl.cc
OHerasym: @anosach-luxoft , f54e61f , Done

anosach-luxoft:
Reviewed

PR #454, opened 2016-04-01
Remove POSIX timer usage

anosach-luxoft:
Roll back previous timer implementation based on `pthread`s because of issues with native POSIX timer. Also, restored implementation had been refactored and simplified.

@AGritsevich, @AGaliuzov, @Kozoriz @Kozoriz, @OHerasym, @pestOO, @malirod, @LuxoftAKutsan, @AByzhynar please review

### DIFF
--- a/src/components/application_manager/src/request_controller.cc
+++ b/src/components/application_manager/src/request_controller.cc
@@ -56,7 +56,6 @@ RequestController::RequestController()
       is_low_voltage_(false) {
   LOG4CXX_AUTO_TRACE(logger_);
   InitializeThreadpool();
-  timer_.Start(default_sleep_time_, true);
 }
 
 RequestController::~RequestController() {
@@ -498,10 +497,6 @@ void RequestController::UpdateTimer() {
                    << front->timeout_msec()/date_time::DateTime::MILLISECONDS_IN_SECOND);
       timer_.Start(0u, true);
     }
-  } else {
-    LOG4CXX_DEBUG(logger_, "Sleep for default sleep time "
-                  << default_sleep_time_ << " milliseconds.");
-    timer_.Start(default_sleep_time_, true);
   }
 }
 
--- a/src/components/policy/src/policy/usage_statistics/src/counter.cc
+++ b/src/components/policy/src/policy/usage_statistics/src/counter.cc
@@ -103,7 +103,7 @@ AppStopwatchImpl::AppStopwatchImpl(
 
 void AppStopwatchImpl::Start(AppStopwatchId stopwatch_type) {
   stopwatch_type_ = stopwatch_type;
-  timer_.Start(time_out_ * date_time::DateTime::MILLISECONDS_IN_SECOND, true);
+  timer_.Start(time_out_ * date_time::DateTime::MILLISECONDS_IN_SECOND, false);
 }
 
 void AppStopwatchImpl::Switch(AppStopwatchId stopwatch_type) {
--- a/src/components/utils/include/utils/timer.h
+++ b/src/components/utils/include/utils/timer.h
@@ -29,105 +29,156 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
-
 #ifndef SRC_COMPONENTS_UTILS_INCLUDE_UTILS_TIMER_H_
 #define SRC_COMPONENTS_UTILS_INCLUDE_UTILS_TIMER_H_
 
+#include <string>
 #include <stdint.h>
 
-#ifdef __QNX__
-#include <sys/siginfo.h>
-#else
-#include <signal.h>
-typedef sigval_t sigval;
-#endif
-
-#include <time.h>
-#include <string>
-
+#include "utils/macro.h"
 #include "utils/lock.h"
 #include "utils/timer_task.h"
+#include "utils/threads/thread.h"
+#include "utils/threads/thread_delegate.h"
 
 namespace timer {
+
 typedef uint32_t Milliseconds;
+
 /**
- * @brief The Timer is class for calling any method after out of internal time.
- * Time setups in ::Start(uint,bool) method and starts time out steps.
- * User can call Start, Stop, IsRunning, GetTimeout.
+ * @brief Timer calls custom callback function after
+ * specified timeout has been elapsed
  */
 class Timer {
  public:
   /**
-   * @brief constructor
-   * @param name for indentify of current timer
-   * @param task_for_tracking is SPtr to trackable task
+   * @brief Constructor
+   * Does not starts timer
+   * @param name Timer name for identity
+   * @param task Task for tracking
    */
-  Timer(const std::string& name, const TimerTask* task_for_tracking);
+  Timer(const std::string& name, TimerTask* task);
 
   /**
-   * @brief destructor - if timer running : call stop in the body
+   * @brief Destructor
+   * Stops timer if it's running
    */
   ~Timer();
+  
   /**
-   * @brief starts timer with new timeout
-   * @param timeout - time to call method from trackable class
-   * @param repeatable - should timer repeat after calling callback, or no
+   * @brief Starts timer with specified timeout
+   * @param timeout Timer timeout
+   * @param single_shot Shows needs to restart timer after timeout
    */
-  void Start(const Milliseconds timeout, const bool repeatable);
+  void Start(const Milliseconds timeout, const bool single_shot);
+
   /**
-   * @brief stops timer without calling callback
+   * @brief Stops timer if it's running
    */
   void Stop();
+
   /**
-   * @brief method which know about Timer state: is running or not
-   * @return true when timer runned, false when timer stand
+   * @brief Gets current timer status
+   * @return True in case of timer is running, false otherwise
    */
-  bool IsRunning() const;
+  bool is_running() const;
+
   /**
-   * @brief GetTimeout
-   * @return returns timeout
+   * @brief Gets current timer timeout
+   * @return Current timeout in milliseconds.
+   * Null if timer has not been started
    */
-  Milliseconds GetTimeout() const;
+  Milliseconds timeout() const;
 
  private:
+  /**
+   * @brief Delegate for timer thread
+   */
+  class TimerDelegate : public threads::ThreadDelegate {
+   public:
+    /**
+     * @brief Constructor
+     * @param timer Timer instance pointer for callback calling
+     */
+    explicit TimerDelegate(const Timer* timer);
+
+    /**
+     * @brief Sets timer timeout
+     * Thread-safe method
+     * @param timeout Timeout in milliseconds to be set
+     */
+    void set_timeout(const Milliseconds timeout);
+
+    /**
+     * @brief Gets timer timeout
+     * Thread-safe method
+     * @return Timer timeout
+     */
+    Milliseconds timeout() const;
+
+    /**
+     * @brief Sets timer delegate stop flag
+     * Thread-safe method
+     * @param stop_flag Bool flag to be set
+     */
+    void set_stop_flag(const bool stop_flag);
+
+    /**
+     * @brief Gets timer delegate stop flag
+     * Thread-safe method
+     * @return Delegate stop flag
+     */
+    bool stop_flag() const;
+
+    void threadMain() OVERRIDE;
+    void exitThreadMain() OVERRIDE;
+
+   private:
+    const Timer* timer_;
+
+    /*
+     * Params lock used to protect timeout_ and stop_flag_ variables
+     */
+    mutable sync_primitives::Lock params_lock_;
+    Milliseconds timeout_;
+    bool stop_flag_;
+
+    /*
+     * State lock used to protect condition variable
+     */
+    sync_primitives::Lock state_lock_;
+    sync_primitives::ConditionalVariable termination_condition_;
+
+    DISALLOW_COPY_AND_ASSIGN(TimerDelegate);
+  };
+
+  /**
+   * @brief Callback called on timeout
+   */
+  void OnTimeout() const;
+
   const std::string name_;
-  const TimerTask* task_;
-  bool repeatable_;
-  uint32_t timeout_ms_;
-  bool is_running_;
-  timer_t timer_;
-  mutable sync_primitives::Lock lock_;
-  sync_primitives::Lock task_lock_;
 
-  /**
-   * @brief method called from friend handler_wrapper and call run() from task.
+  /*
+   * Task lock used to protect task from deleting during execution
    */
-  void OnTimeout();
+  mutable sync_primitives::Lock task_lock_;
+  TimerTask* task_;
 
-  /**
-   * @brief method for setting correct timeout.
-   * @param timeout - if it`s value = 0, timeout will be setted to 1
-   * There would be no way to stop thread if timeout in lopper will be 0
-   * and if we puts to timer_create zero timeout then we get sys error(22)
-   */
-  void SetTimeoutUnsafe(const Milliseconds timeout);
+  mutable TimerDelegate delegate_;
+  threads::Thread* thread_;
 
-  /**
-   * @brief startUnsafe, stopUnsafe - methods used for correct synchronization
-   *  and must be used only with sync_primitive (auto_lock e.g.)
-   * @return true if start/stop successfull, false if unsuccessfull
+  /*
+   * We should not protect this variable with any
+   * synchronization primitives in current implementation
+   * because we use it only in two places, that cannot
+   * be invoked simultaneously
    */
-  void StartUnsafe();
-  bool StopUnsafe();
-
-  /**
-   * @brief alone function which sends to posix_timer as callee
-   * @param signal_value - structure with parameters of posix_timer callee
-   */
-  friend void HandlePosixTimer(sigval signal_value);
+  bool single_shot_;
 
   DISALLOW_COPY_AND_ASSIGN(Timer);
 };
+
 }  // namespace timer
 
 #endif  // SRC_COMPONENTS_UTILS_INCLUDE_UTILS_TIMER_H_
--- a/src/components/utils/src/threads/posix_thread.cc
+++ b/src/components/utils/src/threads/posix_thread.cc
@@ -151,6 +151,10 @@ PlatformThreadHandle Thread::CurrentId() {
   return pthread_self();
 }
 
+bool Thread::IsCurrentThread() const {
+  return pthread_equal(CurrentId(), thread_handle());
+}
+
 bool Thread::start(const ThreadOptions& options) {
   LOG4CXX_AUTO_TRACE(logger_);
 
@@ -258,7 +262,7 @@ void Thread::stop() {
 
 void Thread::join() {
   LOG4CXX_AUTO_TRACE(logger_);
-  DCHECK(!pthread_equal(pthread_self(), handle_));
+  DCHECK_OR_RETURN_VOID(!IsCurrentThread());
 
   stop();
 
--- a/src/components/utils/src/timer.cc
+++ b/src/components/utils/src/timer.cc
@@ -29,190 +29,134 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
-
 #include "utils/timer.h"
 
-#include <ctime>
-#include <cerrno>
-#include <cstring>
+#include <string>
 
-#include "utils/timer_task.h"
-#include "utils/date_time.h"
 #include "utils/macro.h"
+#include "utils/logger.h"
+#include "utils/lock.h"
+#include "utils/timer_task.h"
+#include "utils/conditional_variable.h"
+#include "utils/threads/thread.h"
+#include "utils/threads/thread_delegate.h"
 
 CREATE_LOGGERPTR_GLOBAL(logger_, "Utils")
 
-using date_time::DateTime;
-
-namespace timer {
-
-// Function HandlePosixTimer is not in anonymous namespace
-// because we need to set this func as friend to Timer
-// and for setting friend function must be located in same namespace with class
-void HandlePosixTimer(sigval signal_value) {
-  LOG4CXX_AUTO_TRACE(logger_);
-
-  DCHECK_OR_RETURN_VOID(signal_value.sival_ptr)
-
-  timer::Timer* timer = static_cast<timer::Timer*>(signal_value.sival_ptr);
-  timer->OnTimeout();
-}
-}  // namespace timer
-
-namespace {
-const int kErrorCode = -1;
-
-itimerspec MillisecondsToItimerspec(const timer::Milliseconds miliseconds) {
-  struct itimerspec result;
-
-  result.it_value.tv_sec = miliseconds / DateTime::MILLISECONDS_IN_SECOND;
-  result.it_value.tv_nsec = (miliseconds % DateTime::MILLISECONDS_IN_SECOND) *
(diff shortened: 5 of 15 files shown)


src/components/utils/src/timer.cc
```diff
@@ -29,190 +29,129 @@
...
 void timer::Timer::Stop() {
   LOG4CXX_AUTO_TRACE(logger_);
-  sync_primitives::AutoLock auto_lock(lock_);
-  repeatable_ = false;
-  if (is_running_) {
-    const bool stop_result = StopUnsafe();
-    DCHECK(stop_result);
+  DCHECK(thread_);
+  delegate_.set_stop_flag(true);
+  if (!pthread_equal(pthread_self(), thread_->thread_handle())) {
```
malirod: @anosach-luxoft, please avoid direct pthread usage. This code is not portable.

src/components/utils/src/timer.cc
```diff
@@ -29,190 +29,129 @@
...
+  set_stop_flag(false);
+  while (!stop_flag()) {
+    const Milliseconds curr_timeout = timeout();
+    LOG4CXX_DEBUG(logger_, "Milliseconds left to wait: " << curr_timeout);
+    if (sync_primitives::ConditionalVariable::kTimeout == termination_condition_.WaitFor(auto_lock, curr_timeout)) {
+      LOG4CXX_DEBUG(logger_, "Timer has finished counting. Timeout (ms): " << curr_timeout);
+      DCHECK(timer_);
+      timer_->OnTimeout();
+    } else {
+      LOG4CXX_DEBUG(logger_, "Timeout reset force (ms): " << curr_timeout);
```
malirod: @anosach-luxoft, on my opinion it's better to save timestamp before wait and after wait log the actual waited time. Currently it's not clear how long the timer actually has spent.

src/components/application_manager/src/request_controller.cc
```diff
@@ -496,12 +495,8 @@ void RequestController::UpdateTimer() {
                    << date_time::DateTime::getmSecs(current_time - end_time)
                    << " Request timeout (sec): "
                    << front->timeout_msec()/date_time::DateTime::MILLISECONDS_IN_SECOND);
-      timer_.Start(0u, false);
+      timer_.Start(0u, true);
     }
-  } else {
-    LOG4CXX_DEBUG(logger_, "Sleep for default sleep time "
-                  << default_sleep_time_ << " milliseconds.");
-    timer_.Start(default_sleep_time_, false);
```
malirod: @anosach-luxoft, please describe in the commit message why the starts are redundant.

src/components/utils/src/timer.cc
```diff
@@ -29,190 +29,129 @@
...
-    const bool stop_result = StopUnsafe();
-    DCHECK_OR_RETURN_VOID(stop_result);
-  }
-  StartUnsafe();
+  Stop();
+  delegate_.set_timeout(timeout);
+  delegate_.set_stop_flag(false);
+  single_shot_ = single_shot;
+  DCHECK(thread_);
+  thread_->start();
```
malirod: @anosach-luxoft, will this method wait until the delegate is actually started?

src/components/utils/src/timer.cc
anosach-luxoft: @malirod it's redundant. We check condvar termination status: `kTimeout`. I think it's enough, don't you?

src/components/utils/src/timer.cc
anosach-luxoft: @malirod yes, it will

Kozoriz:
@anosach-luxoft , i don`t sure about correct behavior, can you clarify behavior with @pestOO? 
What timer must do when it running now, and we call Start again?

anosach-luxoft:
@AKozoriz, @pestOO is a participant of this PR. In case we call `Start()` on running timer - timer will be restarted (stopped and started again with new timeout)

src/components/utils/src/timer.cc
malirod: @anosach-luxoft, seems you didn't understand me. To make thing simpler, just remove curr_timeout

src/components/utils/src/timer.cc
```diff
@@ -29,190 +29,129 @@
...
+  DCHECK(thread_);
+  DeleteThread(thread_);
   sync_primitives::AutoLock auto_lock(task_lock_);
   DCHECK(task_);
   delete task_;
+  LOG4CXX_DEBUG(logger_, "Timer " << name_ << " has been destroyed");
 }
 
-void timer::Timer::Start(const Milliseconds timeout, const bool repeatable) {
+void timer::Timer::Start(const Milliseconds timeout, const bool single_shot) {
```
AGaliuzov: nice naming `single_shot`:)

src/components/utils/src/timer.cc
AGritsevich: @anosach-luxoft, you are right, `thread_->start()` really wait until thread started, but in this condition it's not every time signalling when thread started. So, create extra sync object it's really redundant and better find root cause of this problem.

src/components/utils/include/utils/timer.h
```diff
@@ -29,105 +29,116 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
-
 #ifndef SRC_COMPONENTS_UTILS_INCLUDE_UTILS_TIMER_H_
 #define SRC_COMPONENTS_UTILS_INCLUDE_UTILS_TIMER_H_
 
-#include <stdint.h>
```
AByzhynar: @anosach-luxoft Why did you delete this header?

anosach-luxoft: @AByzhynar please remind me what for do we need it?

AGritsevich:
@pestOO, need your opinion here!
1. Have to be here feature to restart one-shot timer with new time-out, like it was released in old timer?
2. Possibly now change Interfaces `Timer `class?

src/components/utils/src/timer.cc
AGritsevich: @anosach-luxoft, And I'm not sure that good solution use own copy `start flag` if we have already flag in `ThreadDelegate`.

src/components/utils/include/utils/timer.h
pestOO: @anosach-luxoft, may be for `uint32_t`?

src/components/utils/include/utils/timer.h
```diff
@@ -29,105 +29,116 @@
...
+    explicit TimerDelegate(const Timer* timer);
 
-  /**
-   * @brief method for setting correct timeout.
-   * @param timeout - if it`s value = 0, timeout will be setted to 1
-   * There would be no way to stop thread if timeout in lopper will be 0
-   * and if we puts to timer_create zero timeout then we get sys error(22)
-   */
-  void SetTimeoutUnsafe(const Milliseconds timeout);
+    void set_timeout(const Milliseconds timeout);
```
pestOO: MAybe also provide doxygen?
For example, Is this time includes `OntImeout` logics routine?
What will be in case of 0 or maximum uint32 value?

src/components/utils/include/utils/timer.h
```diff
@@ -44,20 +44,61 @@ namespace timer {
...
+   */
   class TimerDelegate : public threads::ThreadDelegate {
    public:
+    /**
+     * @brief Constructor
+     * @param timer Timer instance pointer for callback calling
+     */
     explicit TimerDelegate(const Timer* timer);
 
     void set_timeout(const Milliseconds timeout);
```
pestOO: @anosach-luxoft, 
Maybe also provide doxygen?
For example, Is this time includes `OntImeout` logics routine?
What will be in case of 0 or maximum uint32 value? 
Or seuggetion to follow Timer tinout value

src/components/utils/include/utils/timer.h
```diff
@@ -29,105 +29,116 @@
...
-   * @brief alone function which sends to posix_timer as callee
-   * @param signal_value - structure with parameters of posix_timer callee
+   * @brief Callback called on timeout
    */
-  friend void HandlePosixTimer(sigval signal_value);
+  void OnTimeout() const;
+
+  const std::string name_;
+  TimerTask* task_;
+  mutable sync_primitives::Lock task_lock_;
```
pestOO: Amy be use `sync_primitives::Atomic<T>`?

src/components/utils/src/timer.cc
```diff
@@ -29,190 +29,129 @@
...
 }
 
 timer::Timer::~Timer() {
   LOG4CXX_AUTO_TRACE(logger_);
-  LOG4CXX_DEBUG(logger_, "Timer is to be destroyed " << name_);
   Stop();
+  DCHECK(thread_);
+  DeleteThread(thread_);
   sync_primitives::AutoLock auto_lock(task_lock_);
   DCHECK(task_);
```
pestOO: DCHECK is redundant. a) It not lead to any logic problems) b) deletion null is legal

src/components/utils/src/timer.cc
```diff
@@ -29,190 +29,129 @@
...
-  if (is_running_) {
-    const bool stop_result = StopUnsafe();
-    DCHECK_OR_RETURN_VOID(stop_result);
-  }
-  StartUnsafe();
+  Stop();
+  delegate_.set_timeout(timeout);
+  delegate_.set_stop_flag(false);
+  single_shot_ = single_shot;
+  DCHECK(thread_);
```
pestOO: May be `DCHECK_OR_RETURN`?

src/components/utils/src/timer.cc
pestOO: @anosach-luxoft, @AGritsevich, @malirod, in case of first start it will really waits for thread creation, and in case of second iteration it looks like it only notify ConditionalVariable

pestOO: Hm... But i see that in this case 2 - we generate redundant options for already started thread. Need to investigate this.
https://github.com/smartdevicelink/sdl_core/blob/6916a4ea20083315cbf732f1de3c57916f83213b/src/components/utils/src/threads/posix_thread.cc#L182
+
https://github.com/smartdevicelink/sdl_core/blob/6916a4ea20083315cbf732f1de3c57916f83213b/src/components/utils/src/threads/posix_thread.cc#L214

src/components/utils/src/timer.cc
```diff
@@ -29,190 +29,129 @@
...
-  if (is_running_) {
-    const bool stop_result = StopUnsafe();
-    DCHECK_OR_RETURN_VOID(stop_result);
-  }
-  if (repeatable_) {
-    StartUnsafe();
-  }
+  delegate_.set_stop_flag(single_shot_);
+  sync_primitives::AutoLock auto_lock(task_lock_);
+  DCHECK(task_);
```
pestOO: Maybe `OR_RETURN`?

src/components/utils/src/timer.cc
```diff
@@ -29,190 +29,129 @@
...
   }
-  return false;
+  set_timeout(0);
 }
 
-timer::Milliseconds timer::Timer::GetTimeout() const {
-  sync_primitives::AutoLock auto_lock(lock_);
-  return timeout_ms_;
+void timer::Timer::TimerDelegate::exitThreadMain() {
+  sync_primitives::AutoLock auto_lock(state_lock_);
```
pestOO: @anosach-luxoft, lokks liek it will be deadlock - because `threadMain` locks this for whole cycle and the main idea of `exitThreadMain` to stop this `main`

src/components/utils/src/timer.cc
```diff
@@ -29,190 +29,129 @@
...
-  if (StopPosixTimer(timer_)) {
-    is_running_ = false;
-    return true;
+void timer::Timer::TimerDelegate::threadMain() {
+  sync_primitives::AutoLock auto_lock(state_lock_);
+  set_stop_flag(false);
+  while (!stop_flag()) {
+    const Milliseconds curr_timeout = timeout();
+    LOG4CXX_DEBUG(logger_, "Milliseconds left to wait: " << curr_timeout);
+    if (sync_primitives::ConditionalVariable::kTimeout == termination_condition_.WaitFor(auto_lock, curr_timeout)) {
```
pestOO: @anosach-luxoft, May be encapsulate this line to `WaitForInteraption` or something like that method?
may be just `WaitFor` with return result `interapted` and `expired`

src/components/utils/src/timer.cc
pestOO: @anosach-luxoft, agree with @malirod that will be more appropriate to log time elapsed because of interaction instead of waiting timeout.

src/components/utils/src/timer.cc
```diff
@@ -29,190 +29,129 @@
...
+      LOG4CXX_DEBUG(logger_, "Timer has finished counting. Timeout (ms): " << curr_timeout);
+      DCHECK(timer_);
+      timer_->OnTimeout();
+    } else {
+      LOG4CXX_DEBUG(logger_, "Timeout reset force (ms): " << curr_timeout);
+    }
   }
-  return false;
+  set_timeout(0);
```
pestOO: @anosach-luxoft,
Why we need to set a 0 value, which will be set to 1u?

Alos it's mena that after timer exit `get_timout` will return 1u instead of set by user value.

src/components/utils/src/timer.cc
```diff
@@ -29,190 +29,129 @@
...
 }
 
-timer::Milliseconds timer::Timer::GetTimeout() const {
-  sync_primitives::AutoLock auto_lock(lock_);
-  return timeout_ms_;
+void timer::Timer::TimerDelegate::exitThreadMain() {
+  sync_primitives::AutoLock auto_lock(state_lock_);
+  termination_condition_.NotifyOne();
```
pestOO: @anosach-luxoft, do we need to lock CondVeriable for notifying?

pestOO:
Reviewed

src/components/utils/include/utils/timer.h
```diff
@@ -29,105 +29,72 @@
...
+#include "utils/macro.h"
 #include "utils/lock.h"
 #include "utils/timer_task.h"
+#include "utils/threads/thread.h"
+#include "utils/threads/thread_delegate.h"
 
 namespace timer {
+
 typedef uint32_t Milliseconds;
-/**
```
LuxoftAKutsan: @anosach-luxoft   Why did you removed comments?

src/components/utils/include/utils/timer.h
```diff
@@ -29,105 +29,72 @@
...
+
+    void threadMain() OVERRIDE;
+    void exitThreadMain() OVERRIDE;
+
+   private:
+    const Timer* timer_;
+    Milliseconds timeout_;
+
+    bool stop_flag_;
+    sync_primitives::Lock state_lock_;
```
LuxoftAKutsan: @anosach-luxoft  Please write comments with descriptions what this locks locks

src/components/utils/include/utils/timer.h
```diff
@@ -29,105 +29,72 @@
...
-   */
-  void StartUnsafe();
-  bool StopUnsafe();
-
-  /**
-   * @brief alone function which sends to posix_timer as callee
-   * @param signal_value - structure with parameters of posix_timer callee
-   */
-  friend void HandlePosixTimer(sigval signal_value);
+  TimerTask* task_;
```
LuxoftAKutsan: @anosach-luxoft  If timer is responsible to manage task_ lifecycle? if Timer, it is good idea to use autopointer there.

src/components/utils/include/utils/timer.h
```diff
@@ -29,105 +29,72 @@
...
+  mutable sync_primitives::Lock task_lock_;
+
+  mutable TimerDelegate delegate_;
+  threads::Thread* thread_;
+
+  bool single_shot_;
```
LuxoftAKutsan: @anosach-luxoft  Atomic bool would be nice

src/components/utils/src/timer.cc
LuxoftAKutsan: @anosach-luxoft  I am not sure, but in my opinion it would be nice to extract methods to Interface :
Start
StartRepeatable
It is much more clear that some boolean flags.

src/components/utils/src/timer.cc
anosach-luxoft: @pestOO we call `WaitFor()` in `threadMain()` and it will release lock during waiting. Am I right?

src/components/utils/src/timer.cc
LuxoftAKutsan: @anosach-luxoft  If timer try to stop itself, it just won't do anything in your realisation. Is it ok?

src/components/policy/src/policy/usage_statistics/src/counter.cc
```diff
@@ -103,7 +103,7 @@ AppStopwatchImpl::AppStopwatchImpl(
 
 void AppStopwatchImpl::Start(AppStopwatchId stopwatch_type) {
   stopwatch_type_ = stopwatch_type;
-  timer_.Start(time_out_ * date_time::DateTime::MILLISECONDS_IN_SECOND, true);
+  timer_.Start(time_out_ * date_time::DateTime::MILLISECONDS_IN_SECOND, false);
```
LuxoftAKutsan: @malirod @pestOO 
What do you think about my idea to extend interface with 2 methods:
Start
StartRepeatable 
?

src/components/utils/src/timer.cc
anosach-luxoft: @pestOO what do you mean `to lock CondVeriable`?? We have to lock `state_lock_` and we do it

src/components/utils/src/timer.cc
anosach-luxoft: @pestOO, for now `WaitFor()` method returns interruption status (kTimeout, kNoTimeout, etc). Isn't it enough?
You mean we have to add method `WaitForInterruption()`?

src/components/utils/include/utils/timer.h
anosach-luxoft: @LuxoftAKutsan please see 5b988c0

src/components/utils/include/utils/timer.h
anosach-luxoft: @LuxoftAKutsan seems like we don't need to synchronize this variable

src/components/utils/include/utils/timer.h
anosach-luxoft: @LuxoftAKutsan yes, timer is responsible. But our timer class manage this pointer like `auto_ptr` - receives it in constructor and destroys in destructor, so I think `auto_ptr` is redundant here. But I can add if you insist

src/components/utils/src/timer.cc
anosach-luxoft: @LuxoftAKutsan it will not do anything! It will set stop flag on previous line.

src/components/policy/src/policy/usage_statistics/src/counter.cc
anosach-luxoft: @LuxoftAKutsan as for me it's redundant, but if you guys insist I'll do it

src/components/utils/src/timer.cc
pestOO: @anosach-luxoft, I hope Notifing method is thread safety and in this case locking is redundant.

src/components/utils/src/timer.cc
pestOO: @anosach-luxoft , agree it looks good.

src/components/utils/src/timer.cc
pestOO: @anosach-luxoft, Sorry, I messed it up. It works according to Posix.

src/components/policy/src/policy/usage_statistics/src/counter.cc
pestOO: @LuxoftAKutsan, @anosach-luxoft  I agree that it's more OOP way. But I'm not really insist.

pestOO:
@AGritsevich,
> 1. Have to be here feature to restart one-shot timer with new time-out, like it was released in old timer?

Starting timer with new value is looks reasonable.
At least the following case is unappropriated:
```
timer->start(2, SINGLE = TRUE);
// after 2 seconds TimOut occurs
timer->start(4!, SINGLE = TRUE);
// after 21 seconds TimOut occurs
```
Have we test such case?
 
> 2. Possibly now change Interfaces Timerclass?

What do you exactly mean?
> 3. Make Stoptimer every Startcommand?

https://github.com/smartdevicelink/sdl_core/pull/454#discussion_r58364375

src/components/policy/src/policy/usage_statistics/src/counter.cc
malirod: @LuxoftAKutsan, at the moment I don't see any benefit in from methods instead of one.

LuxoftAKutsan: @anosach-luxoft  okey, let it be

src/components/utils/include/utils/timer.h
Kozoriz: @pestOO , who is Amy ? ;)

src/components/application_manager/src/request_controller.cc
Kozoriz: @anosach-luxoft , can you reply that message?

src/components/utils/include/utils/timer.h
pestOO: *may be

src/components/utils/include/utils/timer.h
AByzhynar: @anosach-luxoft : @pestOO is absolutely right.

anosach-luxoft:
All, please review added commits since 	5e35c39. PR has been rebased because of conflicts

anosach-luxoft:
Few words about `DCHECK`s. I insist on using DCHECKs for `task` and `thread` everywhere because according to `Timer` class design these two pointers **should** be alive during **all** `Timer`class life. So I don't found them as redundant

AByzhynar:
Reviewed

AGritsevich:
Reviewed

pestOO:
@anosach-luxoft, please squash and split commits before merge

src/components/utils/src/threads/posix_thread.cc
```diff
@@ -258,7 +262,7 @@ void Thread::stop() {
 
 void Thread::join() {
   LOG4CXX_AUTO_TRACE(logger_);
-  DCHECK(!pthread_equal(pthread_self(), handle_));
+  DCHECK(!IsCurrentThread());
```
pestOO: @anosach-luxoft, May be use `DCHECK_OR_RETUNR`?
In release we will get an leaked thread in stead of deadlock.

anosach-luxoft:
@pestOO, regarding your concern about `Thread` class. This point will be investigated within appropriate [issue](https://adc.luxoft.com/jira/browse/APPLINK-23297) with my own concerns.

src/components/utils/src/timer.cc
pestOO: It looks like we need to delete DCHECK at all all and use returns version only

anosach-luxoft:
Guys, please remind me if I've forgot some of your comments

pestOO:
@anosach-luxoft, thanks for ticket!

pestOO:
@anosach-luxoft, please review _all_ comment - at least few about DCHECK are not answered.

anosach-luxoft:
@pestOO 
Few words about DCHECKs. I insist on using DCHECKs for task and thread everywhere because according to Timer class design these two pointers should be alive during all Timerclass life. So I don't found them as redundant

anosach-luxoft:
@pestOO please see 5007765. After your confirmation I'll squash commits and merge

pestOO:
Reviewed

Issue #461, opened 2016-04-06
Initial policy table load, crashes core on startup


JackLivio:
(https://github.com/smartdevicelink/sdl_core/commit/ea2eaa6a2ac6b4ce59aa3f8e89086aaff1db08b0) - code from that pull request that's causing core to crash on startup after its first build or after the storage folder is deleted. 

This crash only happens on the initial start of SDL Core. Attempting to start SDL Core after this initial crash seems to work fine.

PR #485, opened 2016-04-12
Remove mark_initialization() on SaveUsageAndErrorCounts()

JackLivio:
Original Problem #461 
In cache_manager.cc, the recent addition of [generate snapshot](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/policy/src/policy/src/cache_manager.cc#L1050) post loading the policy table from file would cause Core to crash on its first startup. 

The crash would happen due to the snapshot returning false on [a call to see if the snapshot is valid](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/policy/src/policy/src/cache_manager.cc#L1051).

That call to is_valid() was returning false ultimately due to the two lines which I removed in my pull request.

I couldn't find anywhere in the code where marking count and app_levels as initialized made a difference other than this initial snapshot generation. There is also an inconsistency in what parts of the policy table are actually marked as initialized so that drives me to further believe that "mark_initialized()" serves no functional purpose. 

Needs Review

### DIFF
--- a/src/components/policy/src/policy/src/sql_pt_representation.cc
+++ b/src/components/policy/src/policy/src/sql_pt_representation.cc
@@ -1267,7 +1267,6 @@ bool SQLPTRepresentation::SaveDeviceData(
 
 bool SQLPTRepresentation::SaveUsageAndErrorCounts(
   const policy_table::UsageAndErrorCounts& counts) {
-  const_cast<policy_table::UsageAndErrorCounts&>(counts).mark_initialized();
   dbms::SQLQuery query(db());
   if (!query.Exec(sql_pt::kDeleteAppLevel)) {
     LOG4CXX_WARN(logger_, "Incorrect delete from app level.");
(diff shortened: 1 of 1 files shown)


JackLivio:
@AGaliuzov

AGaliuzov:
@JackLivio just to be clear. I beleive that this is not fxi for #461.
I suspect the problem in #461 is that during start we have incorrect policy table that is the reson for SDL to stop. And currently crash happened somwhere in `LifeCycle::StopComponents`

AGaliuzov:
@JackLivio have you any other questions so far?

JackLivio:
Not right now. I have another solution almost finished. Will update this Pr and ask for review.

JackLivio:
Closing this pr and opening a new pr that fixes #461 with a different name. 

Name of this pr is irrelevant to the actual fix.

PR #498, opened 2016-04-18
Improve Atomic by changing mutex with SpinLock

AGaliuzov:
Since we use `Atomic` class with primitive types
it makes sense to use spinlock insted of mutex for better
performance


anosach-luxoft:
@AGaliuzov what's the advantage of using `SpinLock` here? Why not just a regular mutex?

AGaliuzov:
@anosach-luxoft because it works faster

malirod:
@agaliuzov, usually normal mutext impl uses some kind of spinlock under the hood before calling sys api. Regarding the speed. Do you have any measurments?

anosach-luxoft:
>>> @anosach-luxoft because it works faster

do you have any proof? Did you try to measure performance?

PR #503, opened 2016-04-25
Refactor timer to avoid deadlock

anosach-luxoft:
Next improvements have been implemented:
* Remove all mutexes except `state_lock` in `Timer` (this one is enough for timer proper working)
* Add `finalized` flag to delegate to avoid endless waiting on timer thread's `join` (the case when we're waiting on `join`, but timer restarts itself in callback)
* Move out delegate `start`/`stop` and thread `start`/`stop` logic into separate private methods
* Do not set `stop_flag` to false in threadMain()
* Extend UT coverage of timer

Initial testing on HU has been performed

@AGaliuzov @AGritsevich @Kozoriz @OHerasym @LuxoftAKutsan please review

### DIFF
--- a/src/components/utils/include/utils/timer.h
+++ b/src/components/utils/include/utils/timer.h
@@ -55,7 +55,7 @@ class Timer {
   /**
    * @brief Constructor
    * Does not start timer
-   * @param name Timer name for identity
+   * @param name Timer name for identification
    * @param task Task for tracking
    */
   Timer(const std::string& name, TimerTask* task);
@@ -69,7 +69,7 @@ class Timer {
   /**
    * @brief Starts timer with specified timeout
    * @param timeout Timer timeout
-   * @param single_shot Shows needs to restart timer after timeout
+   * @param single_shot Single shot flag for timer
    */
   void Start(const Milliseconds timeout, const bool single_shot);
 
@@ -101,7 +101,7 @@ class Timer {
      * @brief Constructor
      * @param timer Timer instance pointer for callback calling
      */
-    explicit TimerDelegate(const Timer* timer);
+    TimerDelegate(const Timer* timer, sync_primitives::Lock& state_lock_ref);
 
     /**
      * @brief Sets timer timeout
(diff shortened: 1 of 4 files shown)


LuxoftAKutsan:
Reviewed!

AGritsevich:
Reviewed.

anosach-luxoft:
@AGaliuzov @Kozoriz @OHerasym please review

src/components/utils/src/timer.cc
```diff
@@ -72,98 +73,136 @@ timer::Timer::~Timer() {
...
+void timer::Timer::StopThread() {
+  if (delegate_.finalized_flag()) {
+    return;
+  }
+
+  DCHECK_OR_RETURN_VOID(thread_);
+  if (!thread_->IsCurrentThread()) {
+    delegate_.set_finalized_flag(true);
+    {
+      state_lock_.Release();
```
AGaliuzov: epic workaround:)

src/components/utils/src/timer.cc
```diff
@@ -72,98 +73,136 @@ timer::Timer::~Timer() {
...
+
+void timer::Timer::StopThread() {
+  if (delegate_.finalized_flag()) {
+    return;
+  }
+
+  DCHECK_OR_RETURN_VOID(thread_);
+  if (!thread_->IsCurrentThread()) {
+    delegate_.set_finalized_flag(true);
+    {
```
AGaliuzov: what for this braces?

anosach-luxoft: @AGaliuzov for `AutoUnlock` :smile:. I'll add it

AGaliuzov:
reviwed

src/components/utils/include/utils/timer.h
```diff
@@ -69,7 +69,8 @@ class Timer {
   /**
    * @brief Starts timer with specified timeout
    * @param timeout Timer timeout
-   * @param single_shot Shows needs to restart timer after timeout
+   * @param single_shot Shows if timer should be restarted automatically
```
malirod: @anosach-luxoft, the description of the `single_shot` param looks strange to me. Single shot means that there will be no repeat at all. Please adjust comment.

src/components/utils/include/utils/timer.h
```diff
@@ -127,56 +128,72 @@ class Timer {
      */
     bool stop_flag() const;
 
+    /**
+     * @brief Sets timer delegate finalized flag
+     * @param finalized_flag Bool flag to be set
```
malirod: @anosach-luxoft, the param description is not descriptive at all.

src/components/utils/test/timer_test.cc
```diff
@@ -43,214 +43,276 @@ namespace components {
...
-sync_primitives::Lock test_lock;
-sync_primitives::ConditionalVariable lock_monitor;
+sync_primitives::Lock shot_lock;
+sync_primitives::ConditionalVariable shot_condition;
+
+const std::string kTimerName = "TestTimer";
+const bool kSingleShot = true;
 const uint32_t kDefaultTimeout = 30u;
-const std::string kTimerName = "test_timer";
+const uint32_t kDefaultTimeoutRestart = 45u;
```
OHerasym: @anosach-luxoft , why 45? not 40 or 50

src/components/utils/test/timer_test.cc
```diff
@@ -43,214 +43,276 @@ namespace components {
...
 
-  uint GetCallsCount() const {
+  size_t calls_count() const {
     return calls_count_;
   }
 
- private:
-  mutable uint calls_count_;
+ protected:
+  mutable timer::Timer* timer_;
```
malirod: @anosach-luxoft, why do you need so much of mutable members?

src/components/utils/test/timer_test.cc
```diff
@@ -43,214 +43,276 @@ namespace components {
...
-  }
+// Start - Stop
 
- private:
-  uint calls_count_;
-  timer::Timer internal_timer_;
-};
-}  // namespace
+TEST(TimerTest, Start_Stop_NoLoop_NoCall) {
+  MockTimerTask* mock_task = new MockTimerTask();
```
malirod: @anosach-luxoft, where is delete? Why not stack var or use auto_ptr?

src/components/utils/test/timer_test.cc
```diff
@@ -43,214 +43,276 @@ namespace components {
...
+  EXPECT_FALSE(timer.is_running());
+  EXPECT_EQ(0u, timer.timeout());
+}
 
-TEST_F(TimerTest, Start_NoLoop_OneCall) {
-  // Preconditions
-  test_lock.Acquire();
+TEST(TimerTest, Start_Stop_NoLoop_OneCall) {
+  sync_primitives::AutoLock auto_lock(shot_lock);
   TestTask* task = new TestTask();
```
malirod: @anosach-luxoft, please double check that every new is handled somehow.

Kozoriz:
Reviewed

src/components/utils/test/timer_test.cc
anosach-luxoft: @malirod for special case with timer `Start`/`Stop` from callback. Callback should be `const`. As for me it's acceptable for testing purposes.

src/components/utils/test/timer_test.cc
anosach-luxoft: @malirod because we have to pass pointer to `Task` to `Timer` constructor. `Timer` will delete it in it's destructor.

src/components/utils/test/timer_test.cc
anosach-luxoft: @malirod the same as above

src/components/utils/include/utils/timer.h
anosach-luxoft: @malirod please see 8389d84

src/components/utils/src/timer.cc
anosach-luxoft: @AGaliuzov see f0f59bd

src/components/utils/test/timer_test.cc
malirod: @anosach-luxoft, I still don't get the point of const methods and mutable members

src/components/utils/test/timer_test.cc
anosach-luxoft: @OHerasym these timeout values are both got from ceiling :smile:. We just need two different values. Good question is about different hardware and system performance. I mean: is this timeout enough for some slow machines? Well, I cannot guarantee that, but for now we didn't meet this case so I propose to keep this two values for a while.

OHerasym:
Reviewed

anosach-luxoft:
@OHerasym @malirod see ad792a5

malirod:
Reviewed

PR #505, opened 2016-04-25
Don't use temp string when creating c_str()

### LINKED ISSUE #504: Use after free error causes return info to be blank
in [set global properties request](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/application_manager/src/commands/mobile/set_global_properties_request.cc#L351-L352) a temporary string is used to create the c string which is sent to sendResponse. The temporary string goes out of scope immediately causing the c string to become invalid and sendResponse to send an empty string in the info parameter.

justinjdickow:
fixes #504 

credit @JackLivio also

### DIFF
--- a/src/components/application_manager/src/commands/mobile/set_global_properties_request.cc
+++ b/src/components/application_manager/src/commands/mobile/set_global_properties_request.cc
@@ -344,12 +344,12 @@ void SetGlobalPropertiesRequest::on_event(const event_engine::Event& event) {
 
     mobile_apis::Result::eType result_code;
     const char* return_info = NULL;
+    const std::string return_string = "Unsupported phoneme type sent in a prompt";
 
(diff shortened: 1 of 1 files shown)


src/components/application_manager/src/commands/mobile/set_global_properties_request.cc
```diff
@@ -344,12 +344,12 @@ void SetGlobalPropertiesRequest::on_event(const event_engine::Event& event) {
 
     mobile_apis::Result::eType result_code;
     const char* return_info = NULL;
+    const std::string return_string = "Unsupported phoneme type sent in a prompt";
```
LuxoftAKutsan: @justinjdickow  why you declared ```const std::string return_string``` in this  scope?
You use this string inside if scope.

PR #507, opened 2016-04-25
Add isSDLAllowed to AppPermissions initializer list

justinjdickow:
Fix "Using uninitialized value permissions.isSDLAllowed when calling AppPermissions"


src/components/policy/src/policy/include/policy/policy_types.h
```diff
@@ -234,7 +234,8 @@ struct AppPermissions {
           appRevoked(false),
           appPermissionsConsentNeeded(false),
           appUnauthorized(false),
-          requestTypeChanged(false) {
+          requestTypeChanged(false),
```
Kozoriz: @justinjdickow , invalid order of initialize list. Why you didn`t build your pr?

PR #511, opened 2016-04-26
Fix auto trace test

VVeremjova:
Please review @LuxoftAKutsan

### DIFF
--- a/src/components/utils/test/auto_trace_test.cc
+++ b/src/components/utils/test/auto_trace_test.cc
@@ -51,11 +51,15 @@ using namespace ::logger;
 
 CREATE_LOGGERPTR_GLOBAL(logger_, "AutoTraceTestLog")
 
+namespace{
+const std::string kFileName =
+    file_system::CurrentWorkingDirectory() + "/AutoTraceTestLogFile.log";
+}  // namespace
+
 void Preconditions() {
-  const char* file_name = "AutoTraceTestLogFile.log";
   // Delete file with previous logs
-  if (file_system::FileExists(file_name)) {
-    ASSERT_TRUE(file_system::DeleteFile(file_name))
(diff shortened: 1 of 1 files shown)


src/components/utils/test/auto_trace_test.cc
```diff
@@ -75,7 +75,8 @@ bool CheckAutoTraceDebugInFile(const std::string& testlog) {
   bool isLogFound = false;
   std::string line;
 
-  std::ifstream file_log("AutoTraceTestLogFile.log");
+  std::ifstream file_log(file_system::CurrentWorkingDirectory() +
+                         "/AutoTraceTestLogFile.log");
```
anosach-luxoft: @VVeremjova please use const defined above for file name

src/components/utils/test/auto_trace_test.cc
```diff
@@ -75,7 +75,8 @@ bool CheckAutoTraceDebugInFile(const std::string& testlog) {
   bool isLogFound = false;
```
anosach-luxoft: @VVeremjova I believe you don't need this variable

anosach-luxoft:
@VVeremjova what was the reason of test failure?

LuxoftAKutsan:
Reviewed.  @VVeremjova  good question from @anosach-luxoft

src/components/utils/test/auto_trace_test.cc
VVeremjova: @anosach-luxoft This value actually shows passing or failing test

VVeremjova:
@anosach-luxoft The reason of test failure is not found in log file specific string

VVeremjova:
@anosach-luxoft See 5c865a6

PR #513, opened 2016-04-26
Fix Core dumped after Ignition OFF, Ignition ON and activate EMBEDDED_NAVI

okoreniak:
Calls of ResumeCtrl functions must go before removing application from
the applications set.

Related to: [APPLINK-23161](https://adc.luxoft.com/jira/browse/APPLINK-23161)

Please review @AGritsevich @Anatoliy-Leshin @anosach-luxoft @VProdanov  @VVeremjova.

### DIFF
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -2578,6 +2578,11 @@ void ApplicationManagerImpl::UnregisterApplication(
 
       return;
     }
+    if (is_resuming) {
+      resume_ctrl_.SaveApplication(app_to_remove);
+    } else {
+      resume_ctrl_.RemoveApplicationFromSaved(app_to_remove);
+    }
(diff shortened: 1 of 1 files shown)


anosach-luxoft:
@okoreniak it's not a hotfix but just a regular fix. See [branch naming](https://github.com/smartdevicelink/sdl_core/blob/develop/CONTRIBUTING.md#pull-requests).

anosach-luxoft:
@okoreniak as I can see from previous implementation, application manager tried to save application after removing and as result application could not be saved. So, how did resumption work??

okoreniak:
@anosach-luxoft  1) I'll close this pull request and prepare new one with correct branch name.
2) resumption crashes in some cases when it calls methods of 
 HmiState class  - is_navi_app() or is_media_app() or is_voice_communication_app() which require the presence of application pointer in applications set.

Issue #434, opened 2016-03-23
Policy Table Snapshot Includes Consumer Friendly Messages


JackLivio:
[According to the policy spec](https://github.com/smartdevicelink/sdl_core/wiki/Policy-Manager-Specification#contents-2)  the policy table snapshot should only include the version number of the consumer friendly messages, and not the messages themselves.

Currently core includes these messages in a snapshot.

PR #514, opened 2016-04-26
Rewrite GenerateSnapshot()

JackLivio:
This pr is to fix issue #461. SDL core was crashing due to the snapshot returned from CacheManager::GenerateSnapshot() was being ruled as invalid.

When generating a snapshot, instead of copying the entire contents of the policy table, I copy all contents except for the messages in consumer friendly messages. Policy table snapshot are not supposed to include the actual messages in consumer friendly messages, just the version number of the messages.

I also added an extra condition in ConsumerFriendlyMessages::Validate() that looks to see if messages is initialized. A pt snapshot can have a consumer friendly messages portion, just not the messages included in consumer fiendly messages.

### DIFF
--- a/src/components/policy/src/policy/src/cache_manager.cc
+++ b/src/components/policy/src/policy/src/cache_manager.cc
@@ -790,8 +790,22 @@ utils::SharedPtr<policy_table::Table>
 CacheManager::GenerateSnapshot() {
   CACHE_MANAGER_CHECK(snapshot_);
   sync_primitives::AutoLock lock(cache_lock_);
+
   snapshot_ = new policy_table::Table();
-  snapshot_->policy_table = pt_->policy_table;
+
+  //Copy all members of policy table except messages in consumer friendly messages
+  snapshot_->policy_table.app_policies_section = pt_->policy_table.app_policies_section;
+  snapshot_->policy_table.functional_groupings = pt_->policy_table.functional_groupings;
+  snapshot_->policy_table.consumer_friendly_messages->version = pt_->policy_table.consumer_friendly_messages->version;
+  snapshot_->policy_table.consumer_friendly_messages->mark_initialized();
+  snapshot_->policy_table.module_config = pt_->policy_table.module_config;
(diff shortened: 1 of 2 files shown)


JackLivio:
Fixes #461 as well as #434

JackLivio:
@AGaliuzov update to closed pr #485

LuxoftAKutsan:
@JackLivio I am not sure that such changes are good in master branch.
Also is there some unit tests that can prevent same bugs in future? And if this changes are applicable for develop branch please create also another one pull request to develop.

@dev-gh  Please also review this changes as domain expert

src/components/policy/src/policy/src/cache_manager.cc
```diff
@@ -790,8 +790,22 @@ utils::SharedPtr<policy_table::Table>
 CacheManager::GenerateSnapshot() {
   CACHE_MANAGER_CHECK(snapshot_);
   sync_primitives::AutoLock lock(cache_lock_);
+
   snapshot_ = new policy_table::Table();
-  snapshot_->policy_table = pt_->policy_table;
```
LuxoftAKutsan: @JackLivio Why copy constructor of policy table can't do it?

JackLivio: Because that would copy the entire contents of pt_, and pt_ contains all of the messages in consumer friendly messages. The policy table "snapshot" is no supposed to contain these messages.

JackLivio:
@LuxoftAKutsan I believe this does need to be in Master branch because of the two issues that it fixes. #461 and #434. I can explain how this fixes these issues in more detail if needed.

I will also provide unit tests.

JackLivio:
Fixing #517 before I can add unit tests

justinjdickow:
@AGaliuzov Can you review? Unit Tests will be added back in after these fixes are in place. Right now the unit tests around Policy Tables are not working so they'll be fixed in a separate PR and then unit tests for CacheManager:GenerateSnapshot() will be added.

PR #515, opened 2016-04-27
Fix Core dumped after Ignition OFF, Ignition ON and activate EMBEDDED_NAVI

okoreniak:
Calls of ResumeCtrl functions must go before removing application from
the applications set.

Related to: [APPLINK-23161](https://adc.luxoft.com/jira/browse/APPLINK-23161)

Please review @vlantonov  @AGritsevich @Anatoliy-Leshin @anosach-luxoft @VProdanov @VVeremjova.

### DIFF
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -2578,6 +2578,11 @@ void ApplicationManagerImpl::UnregisterApplication(
 
       return;
     }
+    if (is_resuming) {
+      resume_ctrl_.SaveApplication(app_to_remove);
+    } else {
+      resume_ctrl_.RemoveApplicationFromSaved(app_to_remove);
+    }
     applications_.erase(app_to_remove);
     AppV4DevicePredicate finder(handle);
     ApplicationSharedPtr app = FindApp(accessor, finder);
@@ -2589,13 +2594,6 @@ void ApplicationManagerImpl::UnregisterApplication(
       SendUpdateAppList();
     }
   }
-
-  if (is_resuming) {
-    resume_ctrl_.SaveApplication(app_to_remove);
-  } else {
-    resume_ctrl_.RemoveApplicationFromSaved(app_to_remove);
-  }
-
   if (audio_pass_thru_active_) {
     // May be better to put this code in MessageHelper?
     EndAudioPassThrough();


okoreniak:
Commit didn't change file  src/components/include/utils/message_queue.h 
so the reason of failed check step :
 "codecov/changes — 1 file has unexpected coverage changes not found in diff."
is unclear.

vlantonov:
The same error appeared in my pull request and also is unclear why.

okoreniak:
Please review @vlantonov @AGritsevich @Anatoliy-Leshin @anosach-luxoft @VProdanov @VVeremjova. Very simple change - 5 lines moved only.

VProdanov:
Reviewed

AGritsevich:
Reviewed.

vlantonov:
Reviewed.

okoreniak:
Please review @Anatoliy-Leshin @anosach-luxoft  @VVeremjova. Very simple change - 5 lines moved only.

src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -2578,6 +2578,11 @@ void ApplicationManagerImpl::UnregisterApplication(
 
       return;
     }
+    if (is_resuming) {
+      resume_ctrl_.SaveApplication(app_to_remove);
```
AGaliuzov: `App` is a shared pointer so I beleive it alive. How does this afix works?

okoreniak: @AGaliuzov App is alive, but in some cases DCHECK() happens and gives core dump in processing function calls which are trying to get this App but set doesn't contain it already. For example - function is_media_app(). So resumption should be called when App exists in the set.

VVeremjova:
Reviewed

src/components/application_manager/src/application_manager_impl.cc
okoreniak: Please review @Anatoliy-Leshin @anosach-luxoft. Very simple change - 5 lines moved only.

Anatoliy-Leshin:
Reviewed

src/components/application_manager/src/application_manager_impl.cc
okoreniak: Please review @anosach-luxoft.

PR #532, opened 2016-05-10
Revert to right state condition some timers

AGritsevich:
After re-implement thread-timer to Genivi,
has been inverted timer condition flag.
Earlier was 'true' is loop timer, 'false' one shot timer.
After re-implement 'false' is loop timer, 'true' one shot timer.
Now in code repository exist mixing both variants. Need check
all timers with some commit that happen before re-factoring timer,
and after it change timer flag where it need.

Related-issues: [APPLINK-24255](https://adc.luxoft.com/jira/browse/APPLINK-24255)

Please review @anosach-luxoft, @AByzhynar , @OHerasym, @LuxoftAKutsan .


AGritsevich:
To see right condition I looked at commit f79d32a78799245c56193941ff77cdf097702e13. It earlier than all timer changes.

LuxoftAKutsan:
Reviewed

anosach-luxoft:
Reviewed

OHerasym:
Reviewed

AByzhynar:
Reviewed

PR #546, opened 2016-05-17
Fix StartStream retry sequence

Anatoliy-Leshin:
Fix StartStream retry sequence
    
    Start retry sequence in case user press "Cancel" in pop-up which appears when application
    try to start video service
    
    Issue: APPLINK-21628

### DIFF
--- a/src/components/application_manager/src/commands/hmi/navi_start_stream_request.cc
+++ b/src/components/application_manager/src/commands/hmi/navi_start_stream_request.cc
@@ -110,7 +110,7 @@ void NaviStartStreamRequest::on_event(const event_engine::Event& event) {
       }
       if (hmi_apis::Common_Result::REJECTED == code) {
         LOG4CXX_INFO(logger_, "StartStream response REJECTED ");
-        SendRequest();
+        RetryStartSession();
         break;
       }
     }


Anatoliy-Leshin:
@vlantonov, @anosach-luxoft, @AGritsevich, @okoreniak 
Prease, review

vlantonov:
Reviewed

src/components/application_manager/src/commands/hmi/navi_start_stream_request.cc
```diff
@@ -110,7 +110,7 @@ void NaviStartStreamRequest::on_event(const event_engine::Event& event) {
       }
       if (hmi_apis::Common_Result::REJECTED == code) {
         LOG4CXX_INFO(logger_, "StartStream response REJECTED ");
-        SendRequest();
+        RetryStartSession();
```
AGritsevich: @Anatoliy-Leshin, why earlier SDL does send Request in case `REJECTED`, but now you think it unnecessary?

Anatoliy-Leshin: Because this is wrong logic and SDL should resend StartStream request in case user pressed "cancel" on pop-up

anosach-luxoft: @Anatoliy-Leshin you're going to send one more `NaviStartStream` request when you call `RetryStartSession()` but not going to terminate existing one. What's the reason to call `RetryStartSession()` if `SendRequest()` performs exactly what you need (send NaviStartStream to HMI)?
And seems like this fix is not related to task specified in description. Please correct me if I'm not right.

Anatoliy-Leshin:
@anosach-luxoft, if we'll just `SendRequest()` we get in the loop: ... SDL receive REJECTED -> SDL send Request -> User press cancel -> SDL receive REJECTED -> SDL send Request -> ...
That's why we need `RetryStartSession()` here. 
But i agree about terminating and creation of new request object. Maybe we need to use `SendRequest()` vs `MessageHelper::SendNaviStartStream` in `RetryStartSession()`. But it works, this is strange. Anyway i will check.
Yes, this is small issue that i found during my task and decided to fix.

src/components/application_manager/src/commands/hmi/navi_start_stream_request.cc
AGritsevich: @Anatoliy-Leshin , ok. Thanks.

AGritsevich:
Reviewed.

Anatoliy-Leshin:
suspended due to higher priority task

PR #555, opened 2016-05-23
Fix tests after ptu impl moving

Kozoriz:
- Removed some warnings
- Disabled tests fixed and enabled


Kozoriz:
Uncorect branch name.

PR #558, opened 2016-05-28
Fix long waiting in Async Gmock

LuxoftAKutsan:
Remove dependency of wait time from count of mocked objects
    In old realisation wait time depends from count of mocked objects
    because waiting in cycle. In new realisation Async Wait will wait
    exactly so much time that was passed within parameter

### DIFF
--- a/src/3rd_party-static/gmock-1.7.0/src/gmock-spec-builders.cc
+++ b/src/3rd_party-static/gmock-1.7.0/src/gmock-spec-builders.cc
@@ -761,7 +761,7 @@ bool Mock::AsyncVerifyAndClearExpectations(int timeout_msec)
   return AsyncVerifyAndClearExpectationsLocked(timeout_msec);
 }
 
-bool Mock::AsyncVerifyAndClearExpectationsLocked(const int timeout_msec_in)
+bool Mock::AsyncVerifyAndClearExpectationsLocked(int timeout_msec)
     GTEST_EXCLUSIVE_LOCK_REQUIRED_(internal::g_gmock_mutex) {
   internal::g_gmock_mutex.AssertHeld();
   MockObjectRegistry::StateMap& state_map = g_mock_object_registry.states();
@@ -773,7 +773,6 @@ bool Mock::AsyncVerifyAndClearExpectationsLocked(const int timeout_msec_in)
   // TODO(ezamakhov@gmail.com): refactor the next loops
   bool expectations_met = true;
   timeval first_register_time {0, 0};
-  int timeout_msec = timeout_msec_in;
   for (MockObjectRegistry::StateMap::iterator mock_it = state_map.begin();
       state_map.end() != mock_it; ++mock_it) {
     MockObjectState& state = mock_it->second;
@@ -825,6 +824,8 @@ bool Mock::AsyncVerifyAndClearExpectationsLocked(const int timeout_msec_in)
       } while (true);
     } // mockers iteration
 
+  } // state_map iteration
+
     if (expectations_met) {
       const long elapsed_usecs =
           // first_register_time is empty on no expectations in mocks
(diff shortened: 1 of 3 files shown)


LuxoftAKutsan:
@AGaliuzov @VVeremjova @pestOO Please review

anosach-luxoft:
@LuxoftAKutsan could you please provide description with main idea and reason of such changes?
It looks very strange to change `gmock` logic.

VVeremjova:
@LuxoftAKutsan Reviewed

src/3rd_party-static/gmock-1.7.0/src/gmock-spec-builders.cc
```diff
@@ -761,7 +761,7 @@ bool Mock::AsyncVerifyAndClearExpectations(int timeout_msec)
   return AsyncVerifyAndClearExpectationsLocked(timeout_msec);
 }
 
-bool Mock::AsyncVerifyAndClearExpectationsLocked(const int timeout_msec_in)
+bool Mock::AsyncVerifyAndClearExpectationsLocked(int timeout_msec)
```
pestOO: I recommend to use `const` in all cases where it possible.

pestOO: And where it's not possible - make it possible ;)

src/3rd_party-static/gmock-1.7.0/src/gmock-spec-builders.cc
```diff
@@ -825,22 +824,28 @@ bool Mock::AsyncVerifyAndClearExpectationsLocked(const int timeout_msec_in)
       } while (true);
     } // mockers iteration
 
+  } // state_map iteration
```
pestOO: I think the previous comment was not redundant for such a big loop.

pestOO:
@LuxoftAKutsan, 
I disagree with new implementation.
We 'fix' long test logic, which guaranty correct testing (in case of normal deviation of CPU usage during test-case execution)
In the new implementation in test with a [100+ expectations](https://github.com/smartdevicelink/sdl_core/blob/ea2eaa6a2ac6b4ce59aa3f8e89086aaff1db08b0/src/components/protocol_handler/test/protocol_header_validator_test.cc#L152) we can face with a unexpected fail because of waiting the whole N seconds instead of waiting for the each one.

pestOO:
After discussion with @LuxoftAKutsan we have an agreement that it's seems to be good solution for test with a lot of mocked objects.
@LuxoftAKutsan, please provide extended description for commit message with information about what fixed why and leads to. Thanks!

LuxoftAKutsan:
@pestOO  please approve review

LuxoftAKutsan:
@pestOO please approve review ( description added)
@anosach-luxoft  description added, this is old our modification to gmock.

src/3rd_party-static/gmock-1.7.0/src/gmock-spec-builders.cc
LuxoftAKutsan: @pestOO I think that it is not so good to create completely similar variable just to avoid removing const from function parameter.

LuxoftAKutsan:
@pestOO please approve review

src/3rd_party-static/gmock-1.7.0/src/gmock-spec-builders.cc
pestOO: @LuxoftAKutsan, 
In case of final solution, when we wait all this amount of `msecs`, it's correct way to use only this variable and reduce it directly.

pestOO:
@LuxoftAKutsan, Approved

src/3rd_party-static/gmock-1.7.0/src/gmock-spec-builders.cc
LuxoftAKutsan: @pestOO  This code will be re-factor within task https://adc.luxoft.com/jira/browse/APPLINK-24989

LuxoftAKutsan:
Task for async UT improvement  https://adc.luxoft.com/jira/browse/APPLINK-24989

AGaliuzov:
Reviewed

PR #573, opened 2016-06-06
Applink-24911 Transfer required changes from [PASA] to [GENIVI]

LevchenkoS:
Been done:
- `HMICapabilities` class been
  divided on interface and realization;
- Added specific abstract classes
  for commands testing(`CommandsTest`
  and `CommandRequestTest`);
- Been added unit tests for
  base command classes(`CommandImpl`,
  `CommandRequestImpl`,
  `CommandResponseImpl`);
- Been added unit tests for
  next mobile response commands:
  - `ListFilesResponse`
  - `ReadDIDResponse`
  - `AlertManeuverResponse`
  - `AlertResponse`
  - `SubscribeButtonResponse`

Resolves: [APPLINK-24911](https://adc.luxoft.com/jira/browse/APPLINK-24911)

### DIFF
--- a/src/components/application_manager/include/application_manager/application_manager_impl.h
+++ b/src/components/application_manager/include/application_manager/application_manager_impl.h
@@ -39,6 +39,7 @@
 #include <set>
 #include <deque>
 #include <algorithm>
+#include <memory>
 
 #include "application_manager/hmi_command_factory.h"
 #include "application_manager/application_manager.h"
@@ -1416,7 +1417,7 @@ class ApplicationManagerImpl
   // Thread that pumps messages audio pass thru to mobile.
   impl::AudioPassThruQueue audio_pass_thru_messages_;
 
-  HMICapabilities hmi_capabilities_;
+  std::auto_ptr<HMICapabilities> hmi_capabilities_;
   // The reason of HU shutdown
   mobile_api::AppInterfaceUnregisteredReason::eType unregister_reason_;
 
--- a/src/components/application_manager/include/application_manager/hmi_capabilities.h
+++ b/src/components/application_manager/include/application_manager/hmi_capabilities.h
@@ -97,7 +97,8 @@ class HMICapabilities {
   virtual void set_is_ivi_cooperating(const bool value) = 0;
 
   /*
-   * @brief Interface used to store information about software version of the target
+   * @brief Interface used to store information about software version of the
+   *target
    *
    * @param ccpu_version Received system/hmi software version
    */
@@ -108,7 +109,7 @@ class HMICapabilities {
    *
    * @return TRUE if it supported, otherwise FALSE
    */
-  virtual  const std::string& ccpu_version() const = 0;
+  virtual const std::string& ccpu_version() const = 0;
 
   /*
    * @brief Retrieves if mixing audio is supported by HMI
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -42,6 +42,7 @@
 #include "application_manager/commands/command_impl.h"
 #include "application_manager/commands/command_notification_impl.h"
 #include "application_manager/message_helper.h"
+#include "application_manager/hmi_capabilities_impl.h"
 #include "application_manager/mobile_message_handler.h"
 #include "application_manager/policies/policy_handler.h"
 #include "protocol_handler/protocol_handler.h"
@@ -114,7 +115,7 @@ ApplicationManagerImpl::ApplicationManagerImpl(
     , messages_from_hmi_("AM FromHMI", this)
     , messages_to_hmi_("AM ToHMI", this)
     , audio_pass_thru_messages_("AudioPassThru", this)
-    , hmi_capabilities_(*this)
+    , hmi_capabilities_(new HMICapabilitiesImpl(*this))
     , unregister_reason_(
           mobile_api::AppInterfaceUnregisteredReason::INVALID_ENUM)
     , resume_ctrl_(*this)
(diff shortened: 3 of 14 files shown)


LevchenkoS:
@AByzhynar, @okozlovlux, @Kozoriz, @LuxoftAKutsan, @VVeremjova, please review.

src/components/application_manager/src/hmi_capabilities_impl.cc
```diff
@@ -619,7 +716,7 @@ bool HMICapabilities::load_capabilities_from_file() {
 
       if (check_existing_json_member(ui, "language")) {
         const std::string lang = ui.get("language", "EN-US").asString();
-        set_active_ui_language(MessageHelper::CommonLanguageFromString(lang));
+        ui_language_ = (MessageHelper::CommonLanguageFromString(lang));
```
okozlovlux: @LevchenkoS  parentheses is redundant

src/components/application_manager/test/commands/command_request_impl_test.cc
```diff
@@ -0,0 +1,483 @@
...
+}
+
+TEST_F(CommandRequestImplTest, BasicMethodsOverloads_SUCCESS) {
+  CommandPtr command = CreateCommand<UCommandRequestImpl>();
+
+  // Current implementation always return `true`
+  EXPECT_TRUE(command->Init());
+  EXPECT_TRUE(command->CleanUp());
+  EXPECT_NO_THROW(command->Run());
+  application_manager::event_engine::Event ev(kInvalidFunctionId_);
```
okozlovlux: @LevchenkoS can you namt it "event", not just "ev"

LevchenkoS: @okozlovlux Fixed in 6cc5591.

src/components/application_manager/src/hmi_capabilities_impl.cc
LevchenkoS: @okozlovlux Fixed in 6cc5591.

okozlovlux:
@LevchenkoS reviewed

src/components/application_manager/include/application_manager/application_manager_impl.h
```diff
@@ -1416,7 +1417,7 @@ class ApplicationManagerImpl
   // Thread that pumps messages audio pass thru to mobile.
   impl::AudioPassThruQueue audio_pass_thru_messages_;
 
-  HMICapabilities hmi_capabilities_;
+  std::auto_ptr<HMICapabilities> hmi_capabilities_;
```
vlantonov: auto_ptr will be redundant. I would suggest you to use utils::SharedPtr

LevchenkoS: @vlantonov, perhaps. But `ApplicationManagerImpl` class have absolute control over `HMICapabilities` and share it only by reference. So, in my opinion, there no need to use `SharedPtr` in this case.

src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -114,7 +115,7 @@ ApplicationManagerImpl::ApplicationManagerImpl(
     , messages_from_hmi_("AM FromHMI", this)
     , messages_to_hmi_("AM ToHMI", this)
     , audio_pass_thru_messages_("AudioPassThru", this)
-    , hmi_capabilities_(*this)
+    , hmi_capabilities_(new HMICapabilitiesImpl(*this))
```
vlantonov: If this is a SharedPtr it is good to be initialized using MakeShared

LuxoftAKutsan:
@LevchenkoS  very bad pull request caption. Please make it more informative,

LuxoftAKutsan:
@LevchenkoS  it is confusing that you merge rather splited logic within one pull request. Why you don't create different pull request for different scopes of changes?

src/components/application_manager/include/application_manager/hmi_capabilities.h
```diff
@@ -445,152 +438,10 @@ class HMICapabilities {
...
-  std::string ccpu_version_;
-  bool is_navigation_supported_;
-  bool is_phone_call_supported_;
-
-  ApplicationManager& app_mngr_;
-  HMILanguageHandler hmi_language_handler_;
-
-  DISALLOW_COPY_AND_ASSIGN(HMICapabilities);
+  virtual void convert_json_languages_to_obj(
+      Json::Value& json_languages, smart_objects::SmartObject& languages) = 0;
```
vlantonov: This may be a const function - it does not seem to modify class fields. Also json_languages does not seem to be modified inside the method, asString() is a const method, so it may be const too.

src/components/application_manager/include/application_manager/application_manager_impl.h
vlantonov: @LevchenkoS I have to say that in the C++11 standard auto_ptr is declared deprecated. So future use of this feature may cause problems. I suggested SharedPtr as it is a wrapper which can effectively handle problems like this. And if there was a functionality like C++11 unique_ptr I would have suggested you to use it.

LuxoftAKutsan: @vlantonov I think that for this purpose auti_ptr is better then Shared because of efficiency. When we will use C++11 auto_ptr can same easily be changed to  unique_ptr.

vlantonov: @LuxoftAKutsan @LevchenkoS Agreed

LevchenkoS:
Pull request now will be closed. All these comments will be taken into consideration in a bowl pull requests, which be logically divided.

PR #578, opened 2016-06-07
Create tools to test commands

LevchenkoS:
Been done:
- Added specific abstract classes for
  commands testing(CommandsTest
  and CommandRequestTest);
- Been added unit tests for base
  command classes (CommandImpl,
  CommandRequestImpl, CommandResponseImpl);
- Been added unit tests for next
  mobile response commands:
  - ListFilesResponse
  - ReadDIDResponse
  - AlertManeuverResponse
  - AlertResponse
  - SubscribeButtonResponse

Related to: [APPLINK-24911](https://adc.luxoft.com/jira/browse/APPLINK-24911)

### DIFF
--- a/src/components/application_manager/test/commands/command_impl_test.cc
+++ b/src/components/application_manager/test/commands/command_impl_test.cc
@@ -78,30 +78,28 @@ class CommandImplTest : public CommandsTest<CommandsTestMocks::kIsNice> {
   };
 
   // Create `SmartObject` which handle array of `SmartObjects`
-  static void CreateArray(MessageSharedPtr& msg,
-                          const size_t msg_count = kDefaultMsgCount_) {
-    if (!msg) {
-      msg = CreateMessage(smart_objects::SmartType_Array);
-    }
-    ::smart_objects::SmartArray* array = msg->asArray();
-    SmartObject obj;
+  static MessageSharedPtr CreateArrayMessage(
+      const size_t msg_count = kDefaultMsgCount_) {
+    MessageSharedPtr msg = CreateMessage(smart_objects::SmartType_Array);
+    smart_objects::SmartArray* array = msg->asArray();
     for (size_t i = 0u; i < msg_count; ++i) {
+      SmartObject obj;
       obj[strings::app_id] = i;
       array->push_back(obj);
     }
+    return msg;
   }
   // Create `SmartObject` which handle map of `SmartObjects`
-  static void CreateMap(MessageSharedPtr& msg,
-                        const size_t msg_count = kDefaultMsgCount_) {
-    if (!msg) {
-      msg = CreateMessage(smart_objects::SmartType_Map);
-    }
+  static MessageSharedPtr CreateMapMessage(
+      const size_t msg_count = kDefaultMsgCount_) {
+    MessageSharedPtr msg = CreateMessage(smart_objects::SmartType_Map);
     char key[] = {'A', '\0'};
-    SmartObject obj;
     for (size_t i = 0u; i < msg_count; ++i, ++key[0]) {
+      SmartObject obj;
       obj[strings::app_id] = i;
       (*msg)[key] = obj;
     }
+    return msg;
   }
 
   static const uint32_t kDefaultMsgCount_ = 5u;
@@ -169,8 +167,7 @@ TEST_F(CommandImplTest, ReplaceMobileByHMIAppId_SUCCESS) {
 }
 
 TEST_F(CommandImplTest, ReplaceMobileByHMIAppId_Array_SUCCESS) {
-  MessageSharedPtr msg;
-  CreateArray(msg, kDefaultMsgCount_);
+  MessageSharedPtr msg = CreateArrayMessage(kDefaultMsgCount_);
   UCommandImplPtr command = CreateCommand<UCommandImpl>(msg);
 
   MockAppPtr app = CreateMockApp();
@@ -191,8 +188,7 @@ TEST_F(CommandImplTest, ReplaceMobileByHMIAppId_Array_SUCCESS) {
 }
 
 TEST_F(CommandImplTest, ReplaceMobileByHMIAppId_Map_SUCCESS) {
-  MessageSharedPtr msg;
-  CreateMap(msg, kDefaultMsgCount_);
+  MessageSharedPtr msg = CreateMapMessage(kDefaultMsgCount_);
   UCommandImplPtr command = CreateCommand<UCommandImpl>(msg);
 
   MockAppPtr app = CreateMockApp();
@@ -238,8 +234,7 @@ TEST_F(CommandImplTest, ReplaceHMIByMobileAppId_SUCCESS) {
 }
 
 TEST_F(CommandImplTest, ReplaceHMIByMobileAppId_Array_SUCCESS) {
-  MessageSharedPtr msg;
-  CreateArray(msg, kDefaultMsgCount_);
+  MessageSharedPtr msg = CreateArrayMessage(kDefaultMsgCount_);
 
   UCommandImplPtr command = CreateCommand<UCommandImpl>(msg);
   MockAppPtr app = CreateMockApp();
@@ -260,8 +255,7 @@ TEST_F(CommandImplTest, ReplaceHMIByMobileAppId_Array_SUCCESS) {
 }
 
 TEST_F(CommandImplTest, ReplaceHMIByMobileAppId_Map_SUCCESS) {
-  MessageSharedPtr msg;
-  CreateMap(msg, kDefaultMsgCount_);
+  MessageSharedPtr msg = CreateMapMessage(kDefaultMsgCount_);
 
   UCommandImplPtr command = CreateCommand<UCommandImpl>(msg);
   MockAppPtr app = CreateMockApp();
--- a/src/components/application_manager/test/commands/command_impl_test.cc
+++ b/src/components/application_manager/test/commands/command_impl_test.cc
@@ -1,0 +1,285 @@
+/*
+ * Copyright (c) 2016, Ford Motor Company
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following
+ * disclaimer in the documentation and/or other materials provided with the
+ * distribution.
+ *
+ * Neither the name of the Ford Motor Company nor the names of its contributors
+ * may be used to endorse or promote products derived from this software
+ * without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <stdint.h>
+#include <string>
+#include <set>
+
+#include "gtest/gtest.h"
+#include "utils/shared_ptr.h"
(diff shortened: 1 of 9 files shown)


LevchenkoS:
@AByzhynar, @okozlovlux, @Kozoriz, @LuxoftAKutsan, @VVeremjova, @vlantonov, please review.

okozlovlux:
@LevchenkoS  reviewed

src/components/application_manager/test/commands/command_impl_test.cc
```diff
@@ -0,0 +1,285 @@
...
+    DISALLOW_COPY_AND_ASSIGN(UnwrappedCommandImpl);
+  };
+
+  // Create `SmartObject` which handle array of `SmartObjects`
+  static void CreateArray(MessageSharedPtr& msg,
+                          const size_t msg_count = kDefaultMsgCount_) {
+    if (!msg) {
+      msg = CreateMessage(smart_objects::SmartType_Array);
+    }
+    ::smart_objects::SmartArray* array = msg->asArray();
```
vlantonov: Is :: required?

LevchenkoS: @vlantonov Obviously - not. But it's more to avoid misunderstandings about "Where it comes from?".

vlantonov: In the rest of the file the smart_objects namespace is accessed without this addition. So to me it seemed that there was special reason to write this and I wondered why.

vlantonov:
Reviewed

src/components/application_manager/test/commands/command_request_impl_test.cc
```diff
@@ -0,0 +1,485 @@
...
+  EXPECT_CALL(app_mngr_, CheckPolicyPermissions(_, _, _, _, _))
+      .WillOnce(DoAll(GetArg3(&params), Return(kMobResultSuccess_)));
+
+  EXPECT_TRUE(command->CheckPermissions());
+  EXPECT_TRUE(params.end() !=
+              std::find(params.begin(), params.end(), kPolicyAppId_));
+}
+
+TEST_F(CommandRequestImplTest, AddDisallowedParameters_SUCCESS) {
+  const std::string kDisParam =
```
AByzhynar: @LevchenkoS As for me `DisParam` - is unclear. I would recommend to rename it. 
Dis... - is disabled, disallowed, dismissed or what?

src/components/application_manager/test/commands/command_request_impl_test.cc
```diff
@@ -0,0 +1,485 @@
...
+typedef CommandRequestImplTest::UnwrappedCommandRequestImpl UCommandRequestImpl;
+typedef SharedPtr<UCommandRequestImpl> CommandPtr;
+
+TEST_F(CommandRequestImplTest, OnTimeOut_StateCompleted_UNSUCCESS) {
+  CommandPtr command = CreateCommand<UCommandRequestImpl>();
+
+  // Should be called twice:
+  // First -- on `onTimeOut` method call
+  // Second -- on destruction;
+  EXPECT_CALL(ev_disp_, remove_observer(_)).Times(2);
```
AByzhynar: @LevchenkoS `ev_disp_` - is event (evaluation, evacuation) display? Give more clear name to variable

src/components/application_manager/test/commands/command_impl_test.cc
```diff
@@ -0,0 +1,285 @@
...
+  EXPECT_TRUE(command->CheckPermissions());
+  EXPECT_TRUE(command->CleanUp());
+
+  // Default value of `allowed_to_terminate_` is true
+  EXPECT_TRUE(command->AllowedToTerminate());
+  command->SetAllowedToTerminate(false);
+  EXPECT_FALSE(command->AllowedToTerminate());
+
+  const uint32_t kCorrelationId = 3u;
+  const int32_t kFunctionId = 4u;
```
AByzhynar: @LevchenkoS `int32_t = uint`?

src/components/application_manager/test/commands/command_impl_test.cc
```diff
@@ -0,0 +1,285 @@
...
+    }
+    ::smart_objects::SmartArray* array = msg->asArray();
+    SmartObject obj;
+    for (size_t i = 0u; i < msg_count; ++i) {
+      obj[strings::app_id] = i;
+      array->push_back(obj);
+    }
+  }
+  // Create `SmartObject` which handle map of `SmartObjects`
+  static void CreateMap(MessageSharedPtr& msg,
```
AByzhynar: @LevchenkoS What if you get as parameter msg with SmartObject type INT, or String or anything else but `SmartType_Map` ? I mean it's worth to add check for incoming message smart object type before trying to add something to it.

src/components/application_manager/test/commands/command_impl_test.cc
AByzhynar: @LevchenkoS What if you get as parameter msg with SmartObject type INT, bool, or String or anything else but SmartType_Array ? You will get NULL pointer here. And then in loop body you are dereferencing it. Guess what you get?

src/components/application_manager/test/include/application_manager/commands/command_request_test.h
```diff
@@ -0,0 +1,74 @@
...
+using ::application_manager::commands::Command;
+
+template <const CommandsTestMocks kIsNice = CommandsTestMocks::kNotNice>
+class CommandRequestTest : public CommandsTest<kIsNice> {
+ public:
+  typedef typename TypeIf<kIsNice,
+                          NiceMock<MockEventDispatcher>,
+                          MockEventDispatcher>::Result MockEventDisp;
+
+  MockEventDisp ev_disp_;
```
AByzhynar: @LevchenkoS UNCLEAR VARIABLE NAME

src/components/application_manager/test/include/application_manager/commands/commands_test.h
```diff
@@ -0,0 +1,142 @@
...
+    return ::utils::MakeShared<Command>(msg, app_mngr_);
+  }
+
+  enum { kDefaultTimeout_ = 100u };
+
+  MockAppManager app_mngr_;
+  MockAppManagerSettings app_mngr_set_;
+
+ protected:
+  virtual void InitCommand(const uint32_t& timeout) {
```
AByzhynar: @LevchenkoS I really sorry but why do we have const ref to uint?

LevchenkoS: @AByzhynar Method `InitCommand` get invoked in template method `CreateCommand` to prepare all mocks, before the command constructor will ask for needed data. But `default_timeout` from `ApplicationManagerSettings` was return ref to its data. So to prepare it we should guarantee that the object we want to return will still be alive when we need it. But we need it outside the `InitCommand` method. So, if `timeout` will be a value, when command constructor ask for `default_timeout`, mock will return ref on dead object.

src/components/application_manager/test/include/application_manager/commands/command_request_test.h
LevchenkoS: @AByzhynar I understand that name like `event_dispatcher_` will be more clear, but, in my opinion, this name is too long.

src/components/application_manager/test/include/application_manager/commands/commands_test.h
```diff
@@ -0,0 +1,142 @@
...
+
+enum CommandsTestMocks { kNotNice = 0, kIsNice };
+
+template <const CommandsTestMocks kIsNice = CommandsTestMocks::kNotNice>
+class CommandsTest : public ::testing::Test {
+ public:
+  typedef typename TypeIf<kIsNice,
+                          NiceMock<MockApplicationManager>,
+                          MockApplicationManager>::Result MockAppManager;
+  typedef typename TypeIf<kIsNice,
```
LuxoftAKutsan: @LevchenkoS I think that application_manager_setting should be nice_mock in completely all cases.

src/components/application_manager/test/commands/command_impl_test.cc
LevchenkoS: @AByzhynar If message have type different form `SmartType_Map` this method just simply override it to `SmartType_Map`.

src/components/application_manager/test/commands/command_impl_test.cc
```diff
@@ -0,0 +1,285 @@
...
+        : CommandImpl(message, application_manager) {}
+
+   private:
+    DISALLOW_COPY_AND_ASSIGN(UnwrappedCommandImpl);
+  };
+
+  // Create `SmartObject` which handle array of `SmartObjects`
+  static void CreateArray(MessageSharedPtr& msg,
+                          const size_t msg_count = kDefaultMsgCount_) {
+    if (!msg) {
```
LuxoftAKutsan: @LevchenkoS if msg is output parameter it is better to make it pointer and pass as second parameter.
Also in Unit test  it would be better to DCHECK case if you pass not empty message as output parameter.

src/components/application_manager/test/commands/command_request_impl_test.cc
LevchenkoS: @AByzhynar If you look at test case name it will gets more clear.

src/components/application_manager/test/include/application_manager/commands/commands_test.h
LevchenkoS: @LuxoftAKutsan May be, but some times, result can depends on `ApplicationManagerSetting` data. So if something in test case went wrong because of unknown call, it will be easier to find a problem.

src/components/application_manager/test/commands/command_impl_test.cc
LevchenkoS: @AByzhynar Compiler optimization?

src/components/application_manager/test/include/application_manager/commands/commands_test.h
LuxoftAKutsan: @LevchenkoS  Settings class is data class, our logic cant strictly depend on calling any methods from settings class. If you find case where you want to test that some class will call any methods in settings class, you probably do something wrong.

src/components/application_manager/test/include/application_manager/commands/command_request_test.h
AByzhynar: @LevchenkoS Yes. `mock_event_dispatcher_` will be even better and clear. But it's not about smbd's opinion. It's about reading of your code by other people. Think about next generations)

src/components/application_manager/test/commands/command_impl_test.cc
```diff
@@ -0,0 +1,285 @@
...
+using ::application_manager::ApplicationManager;
+using ::application_manager::commands::MessageSharedPtr;
+using ::application_manager::ApplicationSharedPtr;
+using ::test::components::application_manager_test::MockApplication;
+
+typedef SharedPtr<MockApplication> MockAppPtr;
+
+class CommandImplTest : public CommandsTest<CommandsTestMocks::kIsNice> {
+ public:
+  class UnwrappedCommandImpl : CommandImpl {
```
pestOO: What the reason to wrapping `CommandImpl` in case we are testing it?

src/components/application_manager/test/commands/command_impl_test.cc
```diff
@@ -0,0 +1,285 @@
...
+    UnwrappedCommandImpl(const MessageSharedPtr& message,
+                         ApplicationManager& application_manager)
+        : CommandImpl(message, application_manager) {}
+
+   private:
+    DISALLOW_COPY_AND_ASSIGN(UnwrappedCommandImpl);
+  };
+
+  // Create `SmartObject` which handle array of `SmartObjects`
+  static void CreateArray(MessageSharedPtr& msg,
```
pestOO: Look's like `msg` is output parameter.
In this case better to return `MessageSharedPtr` as a result of this method.

pestOO: Also, in this case name could be like `CreateArrayMessage`

src/components/application_manager/test/commands/command_impl_test.cc
LevchenkoS: @pestOO, the reason is to test its protected methods `ReplaceMobileByHMIAppId` and `ReplaceHMIByMobileAppId`

src/components/application_manager/test/commands/command_impl_test.cc
```diff
@@ -0,0 +1,285 @@
...
+                          const size_t msg_count = kDefaultMsgCount_) {
+    if (!msg) {
+      msg = CreateMessage(smart_objects::SmartType_Array);
+    }
+    ::smart_objects::SmartArray* array = msg->asArray();
+    SmartObject obj;
+    for (size_t i = 0u; i < msg_count; ++i) {
+      obj[strings::app_id] = i;
```
pestOO: I think it's better to reduce scope of this SO.
In this implementation seems _like_ we append some _tricky_ initialized object. (Yes, I realized that `app_id' is const expretion)

LevchenkoS: @pestOO Fixed in aa945bb.

src/components/application_manager/test/commands/command_impl_test.cc
LevchenkoS: @pestOO Fixed in aa945bb.

src/components/application_manager/test/include/application_manager/commands/commands_test.h
LevchenkoS: @LuxoftAKutsan Fixed in aa945bb.

src/components/application_manager/test/include/application_manager/commands/command_request_test.h
LevchenkoS: @AByzhynar Fixed in aa945bb.

src/components/application_manager/test/commands/command_impl_test.cc
LevchenkoS: @AByzhynar Fixed in aa945bb.

src/components/application_manager/test/commands/command_request_impl_test.cc
LevchenkoS: @AByzhynar Fixed in aa945bb.

AByzhynar:
@LevchenkoS **Wrong branch name**

PR #580, opened 2016-06-07
Response with error if HMI send SDL.ActivateApp request to not registered applicaiton

LuxoftAKutsan:
If during resumption application was unregistered HMI still can send SDL.ActivateAppRequest to SDL.
SDL should not fail in this condition, but just report about it in log file and send response with error to HMI.

### DIFF
--- a/src/components/application_manager/src/commands/hmi/sdl_activate_app_request.cc
+++ b/src/components/application_manager/src/commands/hmi/sdl_activate_app_request.cc
@@ -58,6 +58,7 @@ void SDLActivateAppRequest::Run() {
     LOG4CXX_ERROR(
         logger_,
         "Can't find application within regular apps: " << application_id);
+    SendResponse(false, correlation_id(), SDL_ActivateApp, APPLICATION_NOT_REGISTERED);
     return;
   }
 


LuxoftAKutsan:
Please review :
@AByzhynar @AGritsevich @okozlov @nk0leg @Anatoliy-Leshin @dtrunov

src/components/application_manager/src/commands/hmi/sdl_activate_app_request.cc
```diff
@@ -58,6 +58,7 @@ void SDLActivateAppRequest::Run() {
     LOG4CXX_ERROR(
         logger_,
         "Can't find application within regular apps: " << application_id);
+    SendResponse(false, correlation_id(), SDL_ActivateApp, APPLICATION_NOT_REGISTERED);
```
AGritsevich: @LuxoftAKutsan , do you know, that according to req-t you mustn't immediately send response. We have to wait time-out and then send response. And in line 106 we have already one implementation of it.  See [APPLINK-24453](https://adc.luxoft.com/jira/browse/APPLINK-24453).

dtrunov:
Reviewed

LuxoftAKutsan:
Closed according to  APPLINK-24453.:

PR #583, opened 2016-06-08
Parse PT String for HTTP Request Body

JackLivio:
The "body" portion of the httprequest needs to be a valid string. Before this fix, the policy table content of the request did not have properly escaped double quotes and was causing the JSON received by the mobile side to be considered invalid.

The new method "ParsePtString()" adds an escaped backslash for every occurance of a double quote or other escaped backslash. This method also removes any newlines located in the policy table string.

### DIFF
--- a/src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
+++ b/src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
@@ -116,8 +116,8 @@ void OnSystemRequestNotification::AddHeader(BinaryMessage& message) const {
   if (0 > sprintf(timeout_str, "%d", timeout)) {
     memset(timeout_str, 0, sizeof(timeout_str));
   }
-  std::string temp_string = std::string(message.begin(), message.end());
-  std::string policy_table_string = ParsePTString(temp_string);
+  std::string policy_table_string = std::string(message.begin(), message.end());
+  ParsePTString(policy_table_string);
   const std::string header =
 
   "{"
@@ -145,11 +145,11 @@ void OnSystemRequestNotification::AddHeader(BinaryMessage& message) const {
       logger_, "Header added: " << std::string(message.begin(), message.end()));
 }
 
-std::string OnSystemRequestNotification::ParsePTString(std::string& pt_string) const{
+void OnSystemRequestNotification::ParsePTString(std::string& pt_string) const{
   std::string result;
-  int length = pt_string.length();  
+  size_t length = pt_string.length();  
   result.reserve(length*2);
-  for(int i=0;i<length;++i){
+  for(size_t i=0;i<length;++i){
     if(pt_string[i]=='\"' || pt_string[i]=='\\'){
       result += '\\';
     } else if(pt_string[i] == '\n') {
@@ -157,7 +157,7 @@ std::string OnSystemRequestNotification::ParsePTString(std::string& pt_string) c
     }
     result += pt_string[i];
   }
-  return result;
+  pt_string = result;
 }
 #endif
 
--- a/src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
+++ b/src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
@@ -116,7 +116,8 @@ void OnSystemRequestNotification::AddHeader(BinaryMessage& message) const {
   if (0 > sprintf(timeout_str, "%d", timeout)) {
     memset(timeout_str, 0, sizeof(timeout_str));
   }
-
+  std::string temp_string = std::string(message.begin(), message.end());
+  std::string policy_table_string = ParsePTString(temp_string);
   const std::string header =
 
   "{"
@@ -133,7 +134,7 @@ void OnSystemRequestNotification::AddHeader(BinaryMessage& message) const {
               "\"charset\": \"utf-8\","
               "\"Content_Length\": " + std::string(size_str) +
           "},"
-          "\"body\": \"" + std::string(message.begin(), message.end()) + "\""
+          "\"body\": \"" + policy_table_string + "\""
       "}"
   "}";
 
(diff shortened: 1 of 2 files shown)


src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
```diff
@@ -143,6 +144,21 @@ void OnSystemRequestNotification::AddHeader(BinaryMessage& message) const {
   LOG4CXX_DEBUG(
       logger_, "Header added: " << std::string(message.begin(), message.end()));
 }
+
+std::string OnSystemRequestNotification::ParsePTString(std::string& pt_string) const{
+  std::string result;
+  int length = pt_string.length();  
+  result.reserve(length*2);
+  for(int i=0;i<length;++i){
+    if(pt_string[i]=='\"' || pt_string[i]=='\\'){
```
pestOO: @JackLivio,
1.
May be better to use `std::replace` for replacing `"` with `\`?
2. 
Checking `\` has no sense in case of replacing with the same char.

src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
```diff
@@ -143,6 +144,21 @@ void OnSystemRequestNotification::AddHeader(BinaryMessage& message) const {
   LOG4CXX_DEBUG(
       logger_, "Header added: " << std::string(message.begin(), message.end()));
 }
+
+std::string OnSystemRequestNotification::ParsePTString(std::string& pt_string) const{
```
pestOO: It's better to pass `const` string in case of copy return.

src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
```diff
@@ -143,6 +144,21 @@ void OnSystemRequestNotification::AddHeader(BinaryMessage& message) const {
...
+  std::string result;
+  int length = pt_string.length();  
+  result.reserve(length*2);
```
pestOO: Do we need to reserve twice more size?
We replace and escape chars only and size seems to be only less.

src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
JackLivio: @pestOO I will change that.

src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
JackLivio: @pestOO This does not replace those characters, it just adds a backslash before those two characters are copied into the parsed string. Will point out line where original character is copied after backslash is added.

src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
```diff
@@ -143,6 +144,21 @@ void OnSystemRequestNotification::AddHeader(BinaryMessage& message) const {
...
+  for(int i=0;i<length;++i){
+    if(pt_string[i]=='\"' || pt_string[i]=='\\'){
+      result += '\\';
+    } else if(pt_string[i] == '\n') {
+      continue;
+    }
+    result += pt_string[i];
```
JackLivio: @pestOO Here is where the original character is copied into the string. So  ' \" ' would become ' \\\" ' and ' \\' would become ' \\\\ '

src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
JackLivio: @pestOO Not replacing characters, this method prepends ' \\ ' before ' \" ' and ' \\ '

src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
```diff
@@ -143,6 +144,21 @@ void OnSystemRequestNotification::AddHeader(BinaryMessage& message) const {
   LOG4CXX_DEBUG(
       logger_, "Header added: " << std::string(message.begin(), message.end()));
 }
+
+std::string OnSystemRequestNotification::ParsePTString(std::string& pt_string) const{
+  std::string result;
+  int length = pt_string.length();
```
pestOO: `length` shall be `size_t` and `i` in loop for comparation with `lenght`

src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
pestOO: @JackLivio, yes, my fault =)

src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
```diff
@@ -143,6 +144,21 @@ void OnSystemRequestNotification::AddHeader(BinaryMessage& message) const {
   LOG4CXX_DEBUG(
       logger_, "Header added: " << std::string(message.begin(), message.end()));
 }
+
+std::string OnSystemRequestNotification::ParsePTString(std::string& pt_string) const{
+  std::string result;
```
AGaliuzov: @JackLivio In fact I believe you can work with the same string instead of copy. The string class has the `insert` method which is I believe fit you purpose

JackLivio: @AGaliuzov This will be more efficient then using std::string insert() every time those characters are detected. But if you would rather this modify the the same string instead of returning a copy I can change ParsePTString to
```
void OnSystemRequestNotification::ParsePTString(std::string& pt_string) const{
  std::string result;
  size_t length = pt_string.length();  
  result.reserve(length*2);
  for(size_t i=0;i<length;++i){
    if(pt_string[i]=='\"' || pt_string[i]=='\\'){
      result += '\\';
    } else if(pt_string[i] == '\n') {
      continue;
    }
    result += pt_string[i];
  }
  pt_string.reserve(result.length());
  pt_string = result;
}
```

JackLivio: (edited)

PR #623, opened 2016-06-18
Support Clang

LuxoftAKutsan:
Fix all compile warnings with clang compiler


src/components/transport_manager/src/bluetooth/bluetooth_socket_connection_posix.cc
```diff
@@ -222,7 +222,7 @@ void BluetoothSocketConnection::Send() {
       Abort();
       return;
     }
-    if (bytes_sent >= 0) {
+    if (bytes_sent == 0) {
```
pestOO: @LuxoftAKutsan, looks your need to use `0u`

LuxoftAKutsan:
Temporary closed

PR #642, opened 2016-06-23
Fix "Content-Length" Spelling and Calculation

JackLivio:
Content-Length was using an underscore instead of a Hyphen. Android was throwing an exception because there was no "Content-Length" key supplied in the request header.

Also fixed the way the Content-Length is calculated. Android removes the extra backslashes before sending the request to sdl server so I have Content-Length calculated first before "ParsePTString()" is called and the escaping backslashes are added. Then contentLength is sent to ParsePtString to be adjusted if characters are removed from the parsed policy table string (ie '\n' the newline character).

Tested on android and ios.

### DIFF
--- a/src/components/application_manager/include/application_manager/commands/mobile/on_system_request_notification.h
+++ b/src/components/application_manager/include/application_manager/commands/mobile/on_system_request_notification.h
@@ -72,7 +72,7 @@ class OnSystemRequestNotification : public CommandNotificationImpl {
    * @param message Message
    */
   void AddHeader(BinaryMessage& message) const;
-  void ParsePTString(std::string& pt_string, size_t& contentLength) const;
+  size_t ParsePTString(std::string& pt_string) const;
 #endif
 
   DISALLOW_COPY_AND_ASSIGN(OnSystemRequestNotification);
--- a/src/components/application_manager/include/application_manager/commands/mobile/on_system_request_notification.h
+++ b/src/components/application_manager/include/application_manager/commands/mobile/on_system_request_notification.h
@@ -72,7 +72,7 @@ class OnSystemRequestNotification : public CommandNotificationImpl {
    * @param message Message
    */
   void AddHeader(BinaryMessage& message) const;
-  void ParsePTString(std::string& pt_string) const;
+  void ParsePTString(std::string& pt_string, size_t& contentLength) const;
 #endif
 
   DISALLOW_COPY_AND_ASSIGN(OnSystemRequestNotification);
--- a/src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
+++ b/src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
@@ -106,18 +106,28 @@ void OnSystemRequestNotification::AddHeader(BinaryMessage& message) const {
   LOG4CXX_AUTO_TRACE(logger_);
   const int timeout = policy::PolicyHandler::instance()->TimeoutExchange();
 
-  char size_str[24];
-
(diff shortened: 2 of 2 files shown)


src/components/application_manager/include/application_manager/commands/mobile/on_system_request_notification.h
```diff
@@ -72,7 +72,7 @@ class OnSystemRequestNotification : public CommandNotificationImpl {
    * @param message Message
    */
   void AddHeader(BinaryMessage& message) const;
-  void ParsePTString(std::string& pt_string) const;
+  void ParsePTString(std::string& pt_string, size_t& contentLength) const;
```
anosach-luxoft: @JackLivio please use variable names convention `content_length`

src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
```diff
@@ -106,18 +106,28 @@ void OnSystemRequestNotification::AddHeader(BinaryMessage& message) const {
...
   std::string policy_table_string = std::string(message.begin(), message.end());
-  ParsePTString(policy_table_string);
+  
+  /* The Content-Length to be sent in the HTTP Request header should be 
+  calculated before additional escape characters are added to the 
+  policy table string. The mobile proxy will remove the escape 
+  characters after receiving this request. */
+  contentLength = policy_table_string.length();
+
+  ParsePTString(policy_table_string, contentLength);
```
anosach-luxoft: @JackLivio cann't you get required value (content length) by subtraction initial and result `policy_table_string` length ? In this case you dont need to pass variable to function

justinjdickow:
@JackLivio Did you see @anosach-luxoft's comment?

src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
JackLivio: @anosach-luxoft I don't think I understand your suggestion? 

We also don't know if `ParsePTString()` removed a newline character so we can't just subtract the difference of the two lengths to see how many characters were added.

Instead, I can change `ParsePTString()` to have a return value that returns the new content length, and I won't pass content length as a parameter at all.

anosach-luxoft: @JackLivio I prefer second way:
>>> Instead, I can change ParsePTString() to have a return value that returns the new content length, and I won't pass content length as a parameter at all.

JackLivio:
@anosach-luxoft Updated

anosach-luxoft:
Reviewed

src/components/application_manager/src/commands/mobile/on_system_request_notification.cc
```diff
@@ -106,18 +106,28 @@ void OnSystemRequestNotification::AddHeader(BinaryMessage& message) const {
...
   const int timeout = policy::PolicyHandler::instance()->TimeoutExchange();
 
-  char size_str[24];
-
-  if (0 > sprintf(size_str, "%zu", static_cast<size_t>(message.size()))) {
-    memset(size_str, 0, sizeof(size_str));
-  }
-
+  size_t contentLength;
+  char size_str[24];
```
LuxoftAKutsan: @JackLivio  Please move size_str definition closer to its usage

LuxoftAKutsan: @JackLivio Why 24?

LuxoftAKutsan:
Reviewed

Issue #636, opened 2016-06-22
Policy update is been triggered Continiously if we enable  flag ENABLE_EXTENDED_POLICY


madhuy:
1) Enabled ENABLE_EXTENDED_POLICY in SDL core.
2) Policy update been triggered continuously 

Please look for "BasicCommunication.PolicyUpdate" in logs. 

[SmartDeviceLinkCore.txt](https://github.com/smartdevicelink/sdl_core/files/328778/SmartDeviceLinkCore.txt)

JackLivio:
@madhuy looking into this

JackLivio:
@madhuy Do you see this behavior every time? 

I am unable to replicate this behavior at the moment. Using recent versions of SPT, Core, and the web hmi I only see the initial policy update.

JackLivio:
I also do not see BasicCommunication.OnSystemRequest in your logs. I think you might have something hooked up wrong in your HMI?

madhuy:
@JackLivio 
with ENABLE_EXTENDED_POLICY it calls the  "BasicCommunication.PolicyUpdate" NOT "BasicCommunication.OnSystemRequest" for policy update

I think I know the reason . We need to call 
 policy::PolicyHandler::instance()->OnUpdateRequestSentToMobile(); to reset update_required_ value.

in [SendPolicyUpdate](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/application_manager/src/message_helper.cc#L1580)()

With this fix it is working I will create PULL request.

madhuy:
@JackLivio  Please review change https://github.com/smartdevicelink/sdl_core/pull/644/files

AnastasiyaBritanova:
Hi Guys!

Please - it seems you'll break the expected flow of events with these fixes :)

Correct flow for ENABLE_EXTENDED_POLICY: **ON**

1. policies update is triggered: SDL -> HMI: PolicyUpdate (with Snapshot of current policies)
2. HMI **must request** the policices urls via GetURLs RPC 
3. Afterwards, **HMI must generate OnSystemRequest** through SDL to mobile app.

You should not fix SDL to generate OnSystemRequest - in this case, it's HMI's responsibility.

AnastasiyaBritanova:
Please see the attached diagram for details.
![policies_proprietary_flow](https://cloud.githubusercontent.com/assets/11644716/16690788/3d946dde-4532-11e6-976c-f4a20d7d5318.png)

[...]

PR #644, opened 2016-06-23
fix for issue 636

madhuy:
Resetting update_required flag after sending policy update.otherwise it will continiously calling policy update in case of ENABLE_EXTENDED_POLICY

### DIFF
--- a/src/components/application_manager/src/message_helper.cc
+++ b/src/components/application_manager/src/message_helper.cc
@@ -1602,6 +1602,8 @@ void MessageHelper::SendPolicyUpdate(
     object[strings::msg_params][hmi_request::retry][i] = retries[i];
   }
   ApplicationManagerImpl::instance()->ManageHMICommand(message);
+  // to Reset the update_required_ flag
+  policy::PolicyHandler::instance()->OnUpdateRequestSentToMobile();
 }
 
 void MessageHelper::SendGetUserFriendlyMessageResponse(
--- a/src/components/application_manager/src/message_helper.cc
+++ b/src/components/application_manager/src/message_helper.cc
@@ -1602,8 +1602,6 @@ void MessageHelper::SendPolicyUpdate(
     object[strings::msg_params][hmi_request::retry][i] = retries[i];
   }
   ApplicationManagerImpl::instance()->ManageHMICommand(message);
-  // to Reset the update_required_ flag
-  policy::PolicyHandler::instance()->OnUpdateRequestSentToMobile();
 }
 
 void MessageHelper::SendGetUserFriendlyMessageResponse(
--- a/src/components/policy/src/policy/src/update_status_manager.cc
+++ b/src/components/policy/src/policy/src/update_status_manager.cc
@@ -89,6 +89,7 @@ void UpdateStatusManager::OnValidUpdateReceived() {
   update_status_thread_delegate_->updateTimeOut(0); // Stop Timer
   set_exchange_pending(false);
   set_exchange_in_progress(false);
+  set_update_required(false);
 }
 
 void UpdateStatusManager::OnWrongUpdateReceived() {


src/components/application_manager/src/message_helper.cc
```diff
@@ -1602,6 +1602,8 @@ void MessageHelper::SendPolicyUpdate(
     object[strings::msg_params][hmi_request::retry][i] = retries[i];
   }
   ApplicationManagerImpl::instance()->ManageHMICommand(message);
+  // to Reset the update_required_ flag
+  policy::PolicyHandler::instance()->OnUpdateRequestSentToMobile();
```
anosach-luxoft: @madhuy I see that `OnUpdateRequestSentToMobile()` method performs a lot of logic but not only `update_required` flag resetting. Are you sure you need to start this logic here?

madhuy:
@anosach-luxoft 

I am fallowing the flow from [code](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/application_manager/src/application_manager_impl.cc#L1269) 

The above flow is used to update policy with "BasicCommunication.OnSystemRequest"
Same flow I am applying in case of "BasicCommunication. PolicyUpdate"

Hope fully it should be fine. Please do let me know if you have any concerns.

madhuy:
@anosach-luxoft Could you please review this PR

madhuy:
@anosach-luxoft do you any concern regarding this PR ?

anosach-luxoft:
@madhuy sorry for delay, tomorrow I'll review and update you

anosach-luxoft:
@madhuy I have only one concern regarding your implementation. I see that function `MessageHelper::SendPolicyUpdate()` sends message to HMI, but you're going to call `PolicyHandler::instance()->OnUpdateRequestSentToMobile()` which name looks like it should be called after sending message to mobile. As for me it looks strange. @AGaliuzov please help us to clarify.

justinjdickow:
@AGaliuzov Looking for your input on this PR

AGaliuzov:
@madhuy For mee seems like you have called the api for only reason - to reset the appropriate flag. Which is looks wrong. I'm agree with @anosach-luxoft . In case this flag should be changed we have to find proper way to do this.
So, for me it looks like we need to debug the whole sequence and find the real cause of the described problem

madhuy:
@AGaliuzov we need to set 'update_required_' to 'false'. right now it is been set only from the onSystemRequest().  I think it is better to set only after policy update being completed. Please suggest the right place .

madhuy:
Now restting update_required_ flag in UpdateStatusManager::OnValidUpdateReceived()`

Thanks @justin for your suggestion.

LuxoftAKutsan:
@madhuy  do we have any unit tests for this logic. If not, is it possible to create? Does this changes fails existed unit tests?

LuxoftAKutsan:
> @AGaliuzov we need to set 'update_required_' to 'false'. right now it is been set only from the onSystemRequest(). I think it is better to set only after policy update being completed. Please suggest the right place .

@madhuy you is right, but shouldn't we remove setting update_required to false from oSystemRequest?

madhuy:
@JackLivio  Please comment on @LuxoftAKutsan review comments.

justinjdickow:
@AGaliuzov Can you explain in more detail the case where this logic breaks a policy table update? Maybe there is a test case that covers the expectations for the update that @madhuy can work with?

AGaliuzov:
@justinjdickow  I'm not sure about test cases but the flow is the next
1. PTU has been triggered but not yet finished
2. meanwhile user has requested another PTU directly from HMI - SDL set the flag `update_required` to handle this case.
3. First triggered PTU has been finished - SDL check if another update required and triggers on more PTU

Hence we don't miss the requested by user PTU. 
Let me know in case of any further questions

AnastasiyaBritanova:
@madhuy , @AGaliuzov , @joeygrover 
1. if you review the [diagram](https://cloud.githubusercontent.com/assets/11644716/16690788/3d946dde-4532-11e6-976c-f4a20d7d5318.png) you'd see that SDL changes the flag `UPDATE_NEEDED` to `UPDATING` right after `OnSystemRequest` is sent to mobile app.
2. Since @madhuy's HMI **does not send** `OnSystemRequest` per [comment](https://github.com/smartdevicelink/sdl_core/issues/636#issuecomment-232197490), there is no trigger for SDL to reset this flag.
3. Resetting this flag after `OnReceivedPolicyUpdate` is a **wrong** solution, since there is a requirement for SDL to start another PTU if any trigger occurred and changed teh flag to `UPDATE_NEEDED`

**Suggested solution:**
Reset the flag `UPDATE_NEEDED` to `UPDATING` right after `BC.PolicyUpdate` with SnapshotPT is sent to HMI.
By pre-analysis, this should fix "continuous PTU" scenario.

joeygrover:
@AnastasiyaBritanova's solution seems to be a more logical fix.

@madhuy can you comment?

madhuy:
@joeygrover I don't have any issue if @AnastasiyaBritanova suggestion fixes the issue. 
@AnastasiyaBritanova Could please update this PR with your suggestions so that i can experiment?

AnastasiyaBritanova:
@madhuy, do you mean this (from my previous [comment](https://github.com/smartdevicelink/sdl_core/pull/644#issuecomment-239119413) ): 
>Reset the flag UPDATE_NEEDED to UPDATING right after BC.PolicyUpdate with SnapshotPT is sent to HMI.

madhuy:
@AnastasiyaBritanova Yes.

madhuy:
@AnastasiyaBritanova Could you please provide me suggestion to fix this.

AnastasiyaBritanova:
@madhuy 
please see the diagram attached - the marked section: SDL should reset status to `UPDATING` right after it sends `BC.PolicyUpdate` to HMI (instead of previously implemented when SDL changed to `UPDATING` only after `OnSystemRequest` was sent to the app).

proprietary_ptu_flow

madhuy:
@AnastasiyaBritanova please Review #865 .

AnastasiyaBritanova:
@AGaliuzov please review #865

mghiumiusliu:
@madhuy do you plan to resolve conflicts in this PR?

madhuy:
@mghiumiusliu  Closing this there s another PR for sam #865. Please review #865

Issue #479, opened 2016-04-11
AddCommand API responses GENERIC_ERROR response after Sending  200 addCommand


madhuy:
We are trying to test with 200 AddCommand.

We noticed,
1. for first140 requests SDL core responses SUCCESS
2. remaining 60 Requests it reponse GENERIC_ERROR. but it added to command list and shows in HMI.
Note: We are using default time out 20 sec.

Please let us know if we are doing anything wrong.

Attaching logs.
[SmartDeviceLinkCore-2.txt](https://github.com/smartdevicelink/sdl_core/files/213913/SmartDeviceLinkCore-2.txt)

madhuy:
in iOS, SDL core responds multiple SUCCESS responses for same AddCommand Request. it is reproducible with Wifi Transport also.

MprenticeUIE:
The issue appears to be tied to timing (SDL core takes to process and respond) not a specific request limitation.

JackLivio:
@madhuy @MprenticeUIE 

So the issue here its that the AddCommands are timing out. Since there are so many add commands being sent by mobile at once, it doesn't appear that core is able to process all 200 commands before hitting the default timeout. 

By default this value is set 10 seconds in the smartDeviceLink.ini file. Once possible fix is to increase this timeout value. 

Another suggestion would be to try and throttle how fast the mobile device is sending the add commands.

JackLivio:
Additionally I am looking into if the request controller should be updating its timers under certain circumstances. This might be a case.

madhuy:
@JackLivio  the default time out is 20 sec. 
"Note: We are using default time out 20 sec"

MprenticeUIE:
Is there any evidence that an app would add more then 140 commands?

madhuy:
We experimented below as per Ford Recommendation in their Developer portal, It is working. 

_Is there a limit to how fast I can send AddCommand requests?
If you need to send a large number of requests to set up your application up when leaving HMI Level of NONE, the following is recommended:
Send a small number of the most important requests immediately. These usually include button subscriptions, custom prompts (SetGlobalProperties), and four or five of the most important commands.
Using a timer or background thread, send the remaining AddCommand requests in batches of 5 or 10 every second or so.
[...]

PR #654, opened 2016-06-30
Hotfix/ATF_AddSubMenu known issues resolve

BrandonHe:
Please review @pestOO @AGaliuzov @anosach-luxoft 
### Update:
- RequestController::UpdateTimer():<br>
  Change the timer update logic from waiting until close and restart to unblock the waiting threads, avoid timeout issue
- RequestController::OnTimer():<br>
  Add Autolock and wait to process the condition of NOT probably_expired
- Set clean up data to NULL after it's deleted

### Test:
- Both ATF_AddSubMenu and ATF_SetMediaClockTimer ATF tests are passed, do the test several times
- SPT AddSubMenu test passed(successfully add ten submenu)
- make test passed result:

      Start  1: test_JSONCPP
 1/21 Test  #1: test_JSONCPP .....................   Passed    0.04 sec
      Start  2: transport_manager_test
 2/21 Test  #2: transport_manager_test ...........   Passed    0.32 sec
      Start  3: resumption_test
 3/21 Test  #3: resumption_test ..................   Passed    0.02 sec
      Start  4: formatters_test
 4/21 Test  #4: formatters_test ..................   Passed    0.22 sec
      Start  5: protocol_handler_test
 5/21 Test  #5: protocol_handler_test ............   Passed   42.93 sec
      Start  6: connection_handler_test
 6/21 Test  #6: connection_handler_test ..........   Passed   83.30 sec
      Start  7: utils_test
 7/21 Test  #7: utils_test .......................   Passed   20.61 sec
      Start  8: generator_test
 8/21 Test  #8: generator_test ...................   Passed    0.02 sec
      Start  9: security_manager_test
 9/21 Test  #9: security_manager_test ............   Passed    1.00 sec
      Start 10: policy_test
10/21 Test #10: policy_test ......................   Passed    7.29 sec
      Start 11: rpc_base_test
11/21 Test #11: rpc_base_test ....................   Passed    0.03 sec
      Start 12: smart_object_test
12/21 Test #12: smart_object_test ................   Passed    0.73 sec
      Start 13: application_manager_test
13/21 Test #13: application_manager_test .........   Passed    1.89 sec
      Start 14: data_resumption_test
14/21 Test #14: data_resumption_test .............   Passed    0.03 sec
      Start 15: state_controller_test
15/21 Test #15: state_controller_test ............   Passed    0.39 sec
      Start 16: commands_test
16/21 Test #16: commands_test ....................   Passed    0.11 sec
      Start 17: mobile_commands_test
[...]

Issue #621, opened 2016-06-17
Scrollablemessage with DEFAULT_ACTION/STEAL_FOCUS buttons does not send onBUttonPress notification if mobile app send scrollbale message twice.


madhuy:
1. Send Scrollmessage with DEFAULT_ACTION/STEAL_FOCUS buttons
2. Click Button  DEFAULT_ACTION/STEAL_FOCUS  
3. send onBUttonPress notification to mobile app- works as expected.
4. send another scrollable message with DEFAULT_ACTION
5. Click Button  DEFAULT_ACTION/STEAL_FOCUS  
6. Does NOT send onBUttonPress notification to mobile app- NOT working.

Please let me know if you need any info.

AnastasiyaBritanova:
SDL logs would be helpful in this case - to identify who's responsible: HMI for not providing or SDL for not transferring these notifications.

justinjdickow:
@madhuy Please see Ana's comments and provide logs to this issue.

madhuy:
attaching SDL core logs
[SmartDeviceLinkCore.txt](https://github.com/smartdevicelink/sdl_core/files/360681/SmartDeviceLinkCore.txt)

jacobkeeler:
I've not been able to reproduce the problem on my end using the steps you gave, the OnButtonPress still went through to the mobile side on my end.

Based on the logs you uploaded, it appears that the HMI is sending the OnButtonPress to core correctly, as I see the log with the message from the HMI.

```
INFO  [05 Feb 2016 20:46:10,924][ApplicationManager] 			Message to convert: protocol 0; json {"jsonrpc":"2.0","method":"Buttons.OnButtonPress","params":{"appID":1845945318,"customButtonID":5508,"mode":1,"name":15}}
```

The issue may be that your app is not being subscribed to the buttons in the ScrollableMessage (or Core is not correctly keeping track of which buttons are subscribed).

```
ERROR [05 Feb 2016 20:46:10,926][Commands] virtual void application_manager::commands::mobile::OnButtonPressNotification::Run(): Application doesn't subscribed to this custom_button_id.
```

I'll keep looking through the logs and code to see if I can figure out why it isn't being subscribed to the buttons.

AnastasiyaBritanova:
@madhuy @jacobkeeler - please pay attention to the **order of messages**:
line 16471: SDL -> HMI : UI.ScrollableMessage with SBs id=5506 and id=5507 
[...]

PR #660, opened 2016-07-01
Fix/resolve atf add sub menu known issues

BrandonHe:
Please help review @AByzhynar @pestOO @justinjdickow @anosach-luxoft 
Replace PR #654, change branch name from hotfix to fix

### Code Update:
- RequestController::UpdateTimer():<br>
  Change the timer update logic from waiting until close and restart to unblock the waiting threads, avoid timeout issue
- RequestController::OnTimer():<br>
  Add Autolock and wait to process the condition of NOT probably_expired
- SmartObject::cleanup_data():<br>
  Set Smart Object clean up data to NULL after it's deleted

### Verify Test:
- PASSED: Both ATF_AddSubMenu and ATF_SetMediaClockTimer ATF tests, do the test several times
- PASSED: SPT AddSubMenu function test(successfully add ten submenu)
- PASSED: Unit test

**Attached test result:**
[unit_test_reuslt.txt](https://github.com/smartdevicelink/sdl_core/files/342775/unit_test_reuslt.txt)
[ATF_AddSubMenu_fix_test_result.txt](https://github.com/smartdevicelink/sdl_core/files/342777/ATF_AddSubMenu_fix_test_result.txt)
[ATF_SetMediaClockTimer_fix_test_result.txt](https://github.com/smartdevicelink/sdl_core/files/342778/ATF_SetMediaClockTimer_fix_test_result.txt)

### DIFF
--- a/src/components/application_manager/src/request_controller.cc
+++ b/src/components/application_manager/src/request_controller.cc
@@ -379,12 +379,12 @@ void RequestController::onTimer() {
     }
     if (!probably_expired->isExpired()) {
       SDL_DEBUG("onTimer for "
-               << (RequestInfo::HMIRequest == probably_expired->requst_type()
-                       ? "HMI"
-                       : "Mobile")
-               << " request id: " << probably_expired->requestId()
-               << " connection_key: " << probably_expired->app_id()
-               << " NOT expired");
+                << (RequestInfo::HMIRequest == probably_expired->requst_type()
+                        ? "HMI"
+                        : "Mobile")
+                << " request id: " << probably_expired->requestId()
+                << " connection_key: " << probably_expired->app_id()
+                << " NOT expired");
       sync_primitives::AutoLock auto_lock(timer_lock);
       const TimevalStruct current_time = date_time::DateTime::getCurrentTime();
       const TimevalStruct end_time = probably_expired->end_time();
--- a/src/components/application_manager/src/request_controller.cc
+++ b/src/components/application_manager/src/request_controller.cc
@@ -59,7 +59,7 @@ RequestController::RequestController(const RequestControlerSettings& settings)
     , timer_("AM RequestCtrlTimer",
              new timer::TimerTaskImpl<RequestController>(
                  this, &RequestController::onTimer))
-	, stop_flag_(false)
+	  , stop_flag_(false)
     , is_low_voltage_(false)
     , settings_(settings) {
   SDL_AUTO_TRACE();
@@ -376,20 +376,20 @@ void RequestController::onTimer() {
 	  sync_primitives::AutoLock auto_lock(timer_lock);
 	  timer_condition_.Wait(auto_lock);
 	  continue;
-	}
-	if (!probably_expired->isExpired()) {
-	  SDL_INFO("onTimer:isExpired");
-	  sync_primitives::AutoLock auto_lock(timer_lock);
-	  const TimevalStruct current_time = date_time::DateTime::getCurrentTime();
-	  const TimevalStruct end_time = probably_expired->end_time();
-	  if (current_time < end_time) {
-	    const uint32_t msecs = 
-			static_cast<uint32_t>(date_time::DateTime::getmSecs(end_time - current_time));
-		SDL_DEBUG("Sleep for" << msecs << " millisecs");
-		timer_condition_.WaitFor(auto_lock, msecs);
+	  }
+	  if (!probably_expired->isExpired()) {
+	    SDL_INFO("onTimer:isExpired");
+	    sync_primitives::AutoLock auto_lock(timer_lock);
+	    const TimevalStruct current_time = date_time::DateTime::getCurrentTime();
+	    const TimevalStruct end_time = probably_expired->end_time();
+	    if (current_time < end_time) {
+	      const uint32_t msecs = 
+			  static_cast<uint32_t>(date_time::DateTime::getmSecs(end_time - current_time));
+		    SDL_DEBUG("Sleep for" << msecs << " millisecs");
+		    timer_condition_.WaitFor(auto_lock, msecs);
 	  }
 	  continue;
-	}
+	  }
     SDL_INFO("Timeout for "
              << (RequestInfo::HMIRequest == probably_expired->requst_type()
                      ? "HMI"
--- a/src/components/application_manager/src/request_controller.cc
+++ b/src/components/application_manager/src/request_controller.cc
@@ -59,14 +59,19 @@ RequestController::RequestController(const RequestControlerSettings& settings)
     , timer_("AM RequestCtrlTimer",
              new timer::TimerTaskImpl<RequestController>(
                  this, &RequestController::onTimer))
+	, stop_flag_(false)
     , is_low_voltage_(false)
     , settings_(settings) {
   SDL_AUTO_TRACE();
   InitializeThreadpool();
+  timer_.Start(0, timer::kSingleShot);
 }
 
 RequestController::~RequestController() {
   SDL_AUTO_TRACE();
+  stop_flag_ = true;
+  timer_condition_.Broadcast();
+  timer_.Stop();
   if (pool_state_ != TPoolState::STOPPED) {
     DestroyThreadpool();
   }
@@ -364,9 +369,27 @@ void RequestController::onTimer() {
   SDL_AUTO_TRACE();
   SDL_DEBUG(
       "ENTER Waiting fore response count: " << waiting_for_response_.Size());
-  RequestInfoPtr probably_expired =
-      waiting_for_response_.FrontWithNotNullTimeout();
-  while (probably_expired && probably_expired->isExpired()) {
+  while(!stop_flag_) {
+    RequestInfoPtr probably_expired =
+        waiting_for_response_.FrontWithNotNullTimeout();
+    if (!probably_expired) {
+	  sync_primitives::AutoLock auto_lock(timer_lock);
+	  timer_condition_.Wait(auto_lock);
+	  continue;
+	}
+	if (!probably_expired->isExpired()) {
+	  SDL_INFO("onTimer:isExpired");
+	  sync_primitives::AutoLock auto_lock(timer_lock);
+	  const TimevalStruct current_time = date_time::DateTime::getCurrentTime();
+	  const TimevalStruct end_time = probably_expired->end_time();
+	  if (current_time < end_time) {
+	    const uint32_t msecs = 
+			static_cast<uint32_t>(date_time::DateTime::getmSecs(end_time - current_time));
+		SDL_DEBUG("Sleep for" << msecs << " millisecs");
+		timer_condition_.WaitFor(auto_lock, msecs);
+	  }
+	  continue;
+	}
     SDL_INFO("Timeout for "
              << (RequestInfo::HMIRequest == probably_expired->requst_type()
                      ? "HMI"
@@ -470,32 +493,7 @@ void RequestController::Worker::exitThreadMain() {
 
 void RequestController::UpdateTimer() {
   SDL_AUTO_TRACE();
-  RequestInfoPtr front = waiting_for_response_.FrontWithNotNullTimeout();
-  // Buffer for sending request
-  const uint32_t delay_time = 100u;
-  if (front) {
-    const TimevalStruct current_time = date_time::DateTime::getCurrentTime();
-    TimevalStruct end_time = front->end_time();
-    date_time::DateTime::AddMilliseconds(end_time, delay_time);
-    if (current_time < end_time) {
-      const uint32_t msecs = static_cast<uint32_t>(
-          date_time::DateTime::getmSecs(end_time - current_time));
-      SDL_DEBUG("Sleep for " << msecs << " millisecs");
-      // Timeout for bigger than 5 minutes is a mistake
-      timer_.Start(msecs, timer::kSingleShot);
-    } else {
-      SDL_WARN("Request app_id: "
-               << front->app_id() << " correlation_id: " << front->requestId()
-               << " is expired. "
-               << "End time (ms): " << date_time::DateTime::getmSecs(end_time)
-               << " Current time (ms): "
-               << date_time::DateTime::getmSecs(current_time)
-               << " Diff (current - end) (ms): "
-               << date_time::DateTime::getmSecs(current_time - end_time)
-               << " Request timeout (sec): "
-               << front->timeout_msec() / date_time::kMillisecondsInSecond);
-    }
-  }
+  timer_condition_.NotifyOne();
 }
 
 }  //  namespace request_controller
--- a/src/components/application_manager/src/request_controller.cc
+++ b/src/components/application_manager/src/request_controller.cc
@@ -369,7 +369,7 @@ void RequestController::onTimer() {
   SDL_AUTO_TRACE();
   SDL_DEBUG(
       "ENTER Waiting fore response count: " << waiting_for_response_.Size());
-  while(!timer_stop_flag_) {
+  while (!timer_stop_flag_) {
     RequestInfoPtr probably_expired =
         waiting_for_response_.FrontWithNotNullTimeout();
     if (!probably_expired) {
@@ -378,7 +378,7 @@ void RequestController::onTimer() {
       continue;
     }
     if (!probably_expired->isExpired()) {
-      SDL_INFO("onTimer for "
+      SDL_DEBUG("onTimer for "
                << (RequestInfo::HMIRequest == probably_expired->requst_type()
                        ? "HMI"
                        : "Mobile")
--- a/src/components/application_manager/src/request_controller.cc
+++ b/src/components/application_manager/src/request_controller.cc
@@ -59,7 +59,7 @@ RequestController::RequestController(const RequestControlerSettings& settings)
     , timer_("AM RequestCtrlTimer",
              new timer::TimerTaskImpl<RequestController>(
                  this, &RequestController::onTimer))
-    , stop_flag_(false)
+    , timer_stop_flag_(false)
     , is_low_voltage_(false)
     , settings_(settings) {
   SDL_AUTO_TRACE();
@@ -69,7 +69,7 @@ RequestController::RequestController(const RequestControlerSettings& settings)
 
 RequestController::~RequestController() {
   SDL_AUTO_TRACE();
-  stop_flag_ = true;
+  timer_stop_flag_ = true;
   timer_condition_.Broadcast();
   timer_.Stop();
   if (pool_state_ != TPoolState::STOPPED) {
@@ -369,7 +369,7 @@ void RequestController::onTimer() {
   SDL_AUTO_TRACE();
   SDL_DEBUG(
       "ENTER Waiting fore response count: " << waiting_for_response_.Size());
-  while(!stop_flag_) {
+  while(!timer_stop_flag_) {
     RequestInfoPtr probably_expired =
         waiting_for_response_.FrontWithNotNullTimeout();
     if (!probably_expired) {
--- a/src/components/application_manager/src/request_controller.cc
+++ b/src/components/application_manager/src/request_controller.cc
@@ -59,7 +59,7 @@ RequestController::RequestController(const RequestControlerSettings& settings)
     , timer_("AM RequestCtrlTimer",
              new timer::TimerTaskImpl<RequestController>(
                  this, &RequestController::onTimer))
-	  , stop_flag_(false)
+    , stop_flag_(false)
     , is_low_voltage_(false)
     , settings_(settings) {
   SDL_AUTO_TRACE();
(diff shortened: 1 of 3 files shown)


src/components/smart_objects/src/smart_object.cc
```diff
@@ -723,16 +723,28 @@ void SmartObject::duplicate(const SmartObject& OtherObject) {
 void SmartObject::cleanup_data() {
   switch (m_type) {
     case SmartType_String:
-      delete m_data.str_value;
+      if (m_data.str_value) {
```
AGaliuzov: @BrandonHe What the purpose of this `if` condirion and `NULL` assignment?

AGaliuzov:
@LuxoftAKutsan please participate

src/components/application_manager/src/request_controller.cc
```diff
@@ -364,9 +369,27 @@ void RequestController::onTimer() {
   SDL_AUTO_TRACE();
   SDL_DEBUG(
       "ENTER Waiting fore response count: " << waiting_for_response_.Size());
-  RequestInfoPtr probably_expired =
-      waiting_for_response_.FrontWithNotNullTimeout();
-  while (probably_expired && probably_expired->isExpired()) {
+  while(!stop_flag_) {
```
LuxoftAKutsan: @BrandonHe  in this realisation timer will work just like a thread. It is confuting to use timer for thread like functionality. 
If you think that threadlike functionality will be better, please rewrite request controller, remove timer and use thread. But I don't think that it is a good Idea.

LuxoftAKutsan:
@BrandonHe please write root cause of defects. Because this changes rather confusing.

src/components/application_manager/src/request_controller.cc
```diff
@@ -364,9 +369,27 @@ void RequestController::onTimer() {
...
+  while(!timer_stop_flag_) {
+    RequestInfoPtr probably_expired =
+        waiting_for_response_.FrontWithNotNullTimeout();
+    if (!probably_expired) {
+      sync_primitives::AutoLock auto_lock(timer_lock);
+      timer_condition_.Wait(auto_lock);
+      continue;
+    }
+    if (!probably_expired->isExpired()) {
+      SDL_INFO("onTimer:isExpired");
```
LuxoftAKutsan: @BrandonHe  Please add additional information about request that expired

src/components/application_manager/src/request_controller.cc
```diff
@@ -364,9 +369,27 @@ void RequestController::onTimer() {
...
+      sync_primitives::AutoLock auto_lock(timer_lock);
+      const TimevalStruct current_time = date_time::DateTime::getCurrentTime();
+      const TimevalStruct end_time = probably_expired->end_time();
+      if (current_time < end_time) {
+        const uint32_t msecs = static_cast<uint32_t>(
+            date_time::DateTime::getmSecs(end_time - current_time));
+        SDL_DEBUG("Sleep for" << msecs << " millisecs");
```
LuxoftAKutsan: @BrandonHe space after for

src/components/smart_objects/src/smart_object.cc
```diff
@@ -723,16 +723,28 @@ void SmartObject::duplicate(const SmartObject& OtherObject) {
 void SmartObject::cleanup_data() {
```
LuxoftAKutsan: @BrandonHe It is very strange to see modification of logic request controller and smart object in one pull request. Please describe what for do you need this changes.

src/components/smart_objects/src/smart_object.cc
BrandonHe: @AGaliuzov You know, in C plus, if value in ```if``` condition is ```NULL```, operator ```delete``` should do nothing, check the value that to delete is a good style in C plus. And Also, there is a very small possibility that delete fails, that may cause memory leak. Third, this is to avoid of dangling pointer.

src/components/application_manager/src/request_controller.cc
BrandonHe: @LuxoftAKutsan This is not rewrite, actually the timer logic is not changed here,  here add ```AutoLock``` to make it works better in multi-thread program. And during the test, I usually found ```TIME_OUT`` error based on the machine performance.

src/components/smart_objects/src/smart_object.cc
BrandonHe: @LuxoftAKutsan Yeah, this modification is not relative with logic request controller, because this modification is too small changes to send another PR, so I sent this PR here, please reference [comment](https://github.com/smartdevicelink/sdl_core/pull/660#discussion_r69402044). If really need send another PR for this, let me know. Thanks!

src/components/application_manager/src/request_controller.cc
BrandonHe: done with new commit [9a60a08](https://github.com/BrandonHe/sdl_core/commit/9a60a08a6037fb6c47b1aabb885fa6adf8e9516a)

src/components/application_manager/src/request_controller.cc
BrandonHe: done with new commit [9a60a08](https://github.com/BrandonHe/sdl_core/commit/9a60a08a6037fb6c47b1aabb885fa6adf8e9516a)

LuxoftAKutsan:
@BrandonHe  please note that check style fails. 
And describe  root cause of defect.

src/components/application_manager/src/request_controller.cc
```diff
@@ -364,9 +369,27 @@ void RequestController::onTimer() {
   SDL_AUTO_TRACE();
   SDL_DEBUG(
       "ENTER Waiting fore response count: " << waiting_for_response_.Size());
-  RequestInfoPtr probably_expired =
-      waiting_for_response_.FrontWithNotNullTimeout();
-  while (probably_expired && probably_expired->isExpired()) {
+  while(!timer_stop_flag_) {
```
LuxoftAKutsan: @BrandonHe I can see that onTimer function will be called once, and then it will be just looped in condition variable while it won't be stopped in destructor. I can't understand why do you use timer for this purpose, you should use thread logic for this purpose. You just implemented timer inside timer callback, it is confusing. 

Also within you realisation "expired logic" will be executed each time new request is added (because of NotifyOne in UpdateTimeout), it is redundant.

src/components/smart_objects/src/smart_object.cc
AGaliuzov: @BrandonHe 
1.  C++ standard allows NULL-pointer deletion.
Please see C++98 standard 5.3.5 Delete paragraph 2.
> In either alternative, if the value of the operand of `delete` is null pointer the operation has no effect

2.
> And Also, there is a very small possibility that delete fails, that may cause memory leak

Could you please let me know in which way `if-condition` helps to prevent memory leak?

>Third, this is to avoid of dangling pointer.

To avoid dangled pointer you are using `NULL` assignment which is good.

So, all I want is to avoid redundant `if-condition`

src/components/application_manager/src/request_controller.cc
BrandonHe: @LuxoftAKutsan The while condition can also be stopped by ```break``` besides in destructor, 
This purpose is to keep sure whenever this onTimer is called,  the current thread both under condition ```!expired``` and under condition ```!probably_expired->isExpired()``` can be blocked to wake up(previous lack of this purpose). Other executed logic is same with original condition ```(probably_expired && probably_expired->isExpired())```. I am not very clearly understand your thread logic

src/components/smart_objects/src/smart_object.cc
BrandonHe: @AGaliuzov 
> 1. C++ standard allows NULL-pointer deletion.
> Please see C++98 standard 5.3.5 Delete paragraph 2.

I didn't mean that delete NULL-pointer is not allowed, I just mean if is already NULL, the program will no nothing, in case delete two times which is a disaster.
> 2. Could you please let me know in which way if-condition helps to prevent memory leak?

I mean this not a reason for ```if``` but for ```NULL```
Thanks!

AGaliuzov: @BrandonHe Al I try to say is that `if-condition` has no ane sense here. Please remove it

BrandonHe: @AGaliuzov 
Done in commit [`461e69f`](https://github.com/smartdevicelink/sdl_core/pull/660/commits/461e69fc7510ed435e71a6aa0e00242a128221ff)

LuxoftAKutsan:
@BrandonHe I still cant figure out root cause of problem that you tried to solve

BrandonHe:
@LuxoftAKutsan 
OK, Here is more detailed infos, hopes let you know the cause.
I tried to solve in this PR is that during the ATF_AddSubMenu testing, we found test fails as well as in [KnownIssues.md](https://github.com/smartdevicelink/sdl_atf_test_scripts/blob/master/KnownIssues.md)

We analysed and found it caused by timeout. As you know, ```RequestController``` is responsible for handling timeouts of responsibles and notifications from HMI. 

The previous ```UpdateTimer``` is just wait a fixed time to close the Timer and restart a new one, BUT program here relate to waiting and waking up of multi-threading, the actual wait time may longer than that fixed time. It will cause timeout.

BrandonHe:
@AGaliuzov @LuxoftAKutsan 
Do you have any other feedback?

LuxoftAKutsan:
@BrandonHe But how your fix remove issue that next real timeout may  y be called a little bit later then fixed one?

BrandonHe:
@LuxoftAKutsan 
By adding auto_lock here:
```
+      sync_primitives::AutoLock auto_lock(timer_lock);
+      timer_condition_.Wait(auto_lock)
...
...
+      timer_condition_.NotifyOne();
```
BTW, we run all ATF test scripts on limited performance port(such as WinCE), most failed results are timeout.

src/components/application_manager/src/request_controller.cc
LuxoftAKutsan: @BrandonHe I am trying to understand new flow. I can see that you mixed timers and loops. There is simplified current logic of RequestCtrl [sequence](https://www.websequencediagrams.com/cgi-bin/cdraw?lz=dGl0bGUgUmVxdWVzdCBjb250cm9sbGVyCgoKQU0tPgASB0N0cmw6IEFkZAAiCAoADgsAFw9QdXQgcgBJB2luIHF1ZXVlIAAiDlRpbWVyOiBVcGRhdGUgdGltZXIKAA8FAAUWAA18RG8gbm90IHUAgRgTAIIPDU9uVGltZW91AIF9HXJvY2VzcyBmaXJzAIISEgCBZj0APVYgKGxhc3Qgb25lKQCDLBVTdXNwZW5kAIMtFQAODgo&s=earth) Could you create sequence diagram with new logic?

LuxoftAKutsan: @BrandonHe  do you have any updates?

BrandonHe: @LuxoftAKutsan Sorry for delay reply, I missed the mail. New sequence diagram please reference [**here**](https://www.websequencediagrams.com/cgi-bin/cdraw?lz=dGl0bGUgUmVxdWVzdCBDb250cm9sbGVyCgpBTS0-ABEHQ3RybDogQWRkACEICgAOCwAXD3B1dCByAEgHaW4gcXVldWUAIQ5UaW1lcjogU3RhcnQgAAgFCgpsb29wIG5vdCB0aW1lcl9zdG9wX2ZsYWcKCm9wdAA_BiBpcyBlbXB0eQoAPAUAPwlBdXRvIGxvY2sgd2FpdAplbmQALwZub3QgaXNFeHBpcmVkACUPd2FpdCBmb3IgbXNlY3MALwUAMwUAgRQUVXBkYXRlAIETBgBsD1dhaXQAgXE4AIF_KQBgIACCBwcAgx4Nb25UaW1lb3V0AIFEDwCDQw1Qcm9jZXNzIGZpcnMAAYELAHkTIChsYXN0IG9uZSkAhDUWdXNwZW5kAIMPFQCEEA8AhSAaRXhpAIVJDwCFIgcAhQQPID0gdHJ1ZQoK&s=rose), I think the most logic is not changed.

BrandonHe: @LuxoftAKutsan Any other concerns do you have?

LuxoftAKutsan:
@BrandonHe could you create Unit Test where you can reproduce defect in request controller that will be passed after your changes?

BrandonHe:
@LuxoftAKutsan Sorry, I am not familiar with create Unit Test:disappointed:. But I think it's unnecessary to rewrite unit test code for this PR as most logic is not changed, supposed you have compared [previous sequence](https://www.websequencediagrams.com/cgi-bin/cdraw?lz=dGl0bGUgUmVxdWVzdCBjb250cm9sbGVyCgoKQU0tPgASB0N0cmw6IEFkZAAiCAoADgsAFw9QdXQgcgBJB2luIHF1ZXVlIAAiDlRpbWVyOiBVcGRhdGUgdGltZXIKAA8FAAUWAA18RG8gbm90IHUAgRgTAIIPDU9uVGltZW91AIF9HXJvY2VzcyBmaXJzAIISEgCBZj0APVYgKGxhc3Qgb25lKQCDLBVTdXNwZW5kAIMtFQAODgo&s=earth) and [updated sequence](https://www.websequencediagrams.com/cgi-bin/cdraw?lz=dGl0bGUgUmVxdWVzdCBDb250cm9sbGVyCgpBTS0-ABEHQ3RybDogQWRkACEICgAOCwAXD3B1dCByAEgHaW4gcXVldWUAIQ5UaW1lcjogU3RhcnQgAAgFCgpsb29wIG5vdCB0aW1lcl9zdG9wX2ZsYWcKCm9wdAA_BiBpcyBlbXB0eQoAPAUAPwlBdXRvIGxvY2sgd2FpdAplbmQALwZub3QgaXNFeHBpcmVkACUPd2FpdCBmb3IgbXNlY3MALwUAMwUAgRQUVXBkYXRlAIETBgBsD1dhaXQAgXE4AIF_KQBgIACCBwcAgx4Nb25UaW1lb3V0AIFEDwCDQw1Qcm9jZXNzIGZpcnMAAYELAHkTIChsYXN0IG9uZSkAhDUWdXNwZW5kAIMPFQCEEA8AhSAaRXhpAIVJDwCFIgcAhQQPID0gdHJ1ZQoK&s=rose). Also, I have run unit test after changing the codes, and all passed, you can see result in **Attached test result:** in my first comment.

LuxoftAKutsan:
@BrandonHe As I understand there were some problem in Request controller, so you made this changes. We need unit test to be sure that this problem won't repeated in future at least.  And Unit test would be rather good approve that this problem existed without this changes and not it is not reproduces. ATF is very  common check, area of defects found by ATF is rely big. if any code is possible to check with UT, we should check it with UT. 
And any way UT is part of pull request [checkout ](https://github.com/smartdevicelink/sdl_core/blob/develop/COMMITTERS.md#pull-request-checklisthttps://github.com/smartdevicelink/sdl_core/blob/develop/COMMITTERS.md#pull-request-checklist )
There are already  have some unit tests for request controller, you can see how did they done and create new one.

BrandonHe:
@LuxoftAKutsan After research, I think the present UT has covered this changed in [request_controller_test.cc](https://github.com/smartdevicelink/sdl_core/blob/develop/src/components/application_manager/test/request_controller/request_controller_test.cc). This is no parameters or return values changed in ```void RequestController::onTimer()``` and in```void RequestController::UpdateTimer()```, hence, I think there is no need to add new extra Mock or Expect behaviors. Could you give more instructions if you still think need this?

LuxoftAKutsan:
@BrandonHe I wan't to find out what exactly problem did you fix. So I would like you to write Unit test that will fail with old realisation of RequestCtrl and will not fail on current.

LuxoftAKutsan:
@BrandonHe I am reading this changes. 
I can see, that in case if probably expired request in not expired, you prefer to wait on condition variable while it won't be expired. UpdateTimer function and timer implementation will actually do this. 
UpdateTimer will take FronWithNotNullTimeout and start timer for this value.  Timer will wait this time on condition variable and call onTimer again. 
So I can't figure out what problem does it fix. You just copied part of timer realisation inside onTimer method. Not it looks very very complicated because of couple of continue and breaks inside cycle, big amount of code in onTimer method, mixing of condition variables usages and Timer to find expired request.

LuxoftAKutsan:
@BrandonHe  Do you have any updates?

BrandonHe:
@LuxoftAKutsan 
Sorry for ignore your comment as received too much mail from smartdevicelink repo every day. Yeah, you are exactly right about what i did in this PR. The original reason is that I tried to fix ATF_AddSubMenu fails in [KnowIssues.md](https://github.com/smartdevicelink/sdl_atf_test_scripts/blob/master/KnownIssues.md), I analysed that may cause by timeout, so I updated the codes and **PASSED** ATF_AddSubMenu test as you see in the PR description attachment [ATF_AddSubMenu_fix_test_result.txt](https://github.com/smartdevicelink/sdl_core/files/342777/ATF_AddSubMenu_fix_test_result.txt)

But after reading code again and again, I think you are right, it seems has same logic with previous code, and may make code more complicated. Anyway. I think there is also no new issue added.

AGaliuzov:
@BrandonHe In this case do we still need this pull request if it
>seems has same logic with previous code, and may make code more complicated.

BrandonHe:
@AGaliuzov @LuxoftAKutsan 
Could you please have a discussion, and let me know if i need to close this PR. Thanks!

LuxoftAKutsan:
@BrandonHe if there is not changes in logic and code became more complicated, you can close  this pull request. 

I suppose that ATF_AddSubMenu test fails due to some other reason, and you changes timings that probably shadows fails of ATF_AddSubMenu test. This is just my subjection, and fails of ATF_AddSubMenu test  need additional investigation.

BrandonHe:
OKay.
Close this pull request, need further investigation

BrandonHe:
@AGaliuzov @LuxoftAKutsan @JackLivio @madhuy 
Reopen this PR since this fix is effective to resolve issue #479 , please see comment in #479.
Could you guys review again?

LuxoftAKutsan:
@BrandonHe we have same situation as before. Can you find out root case? Why with previous implementation of RequestCtrl ATF tests failed, but with new does not. I can see that you copied part of  timer implementation(loop condition variable and stop flag staff) in RequestCtrl OnTimer method and I can't see how it can help because it doesn't change logic, just made it more complex.

jacobkeeler:
@BrandonHe Small note: if this fixes issues that currently exist in master, as seems to be stated by @madhuy in #479 and #621, then this should be set up as a hotfix and be based off of master.

I actually have already performed the rebasing as part of an attempt to test your changes, so I could push the rebased version to your fork under a new branch if that is alright.

BrandonHe:
@LuxoftAKutsan As mentioned in previous comments thread, I still think the root cause is not use ```auto_lock``` method in process of each request. It may time out if there have many requests to respond but SDL can not process them as quick as within timeout value. So I use loop to keep sure each request executes  ```timer_condition_.WaitFor(auto_lock, msecs)```. Alternatively, we can change the configuration in smartdevicelink.ini to set a longer time out value, but it's a workaround. Both Android and iOS has this problem. All in all, the most logic in request controller is not changed here I think.

Attached sequence diagram again for all of you reference:
**previous sequence:**
request controller sequence diagram

-------
**update the sequence:**
request controller sequence diagram updated by brandon

BrandonHe:
@jacobkeeler OK, thanks. I am not very familiar with rule, so I changed branch name from hotfix to fix. anything here need I to do, let me know.

jacobkeeler:
@BrandonHe I don't see any changes in your SDL fork. 

This branch is still based off of develop, and it needs to be based off of master since it fixes existing issues in the current release of SDL. Since this will likely need to be done on a new branch, a new PR will likely need to be created. 

I have already performed the rebase of these changes and created [this new branch](https://github.com/smartdevicelink/sdl_core/tree/hotfix/resolve_ATF_AddSubMenu_KnownIssues) on the main repo.

BrandonHe:
@jacobkeeler OK, I will branch off from master to create a new hotfix branch, rebase these changes and send a new pull request later.

jacobkeeler:
@BrandonHe By the way, I attempted to run the tests that this PR was originally made to fix (ATF_AddSubMenu.lua), and none of the AddSubMenu tests mentioned in [KnownIssues.md](https://github.com/smartdevicelink/sdl_atf_test_scripts/blob/master/KnownIssues.md) fail on the master branch as it stands.

Do these tests only fail under special conditions, or were they already fixed in master through another PR?

From what I can understand, this PR is meant to take care of cases where requests can timeout despite getting a response in time, because they are unable to process them before time runs out. Is this correct?

madhuy:
@jacobkeeler @BrandonHe  Please prioritize this PR.

BrandonHe:
@madhuy Resend PR #909 
@jacobkeeler For preconditions, I noticed you had discussion in issue #621, maybe you can also reference #479, I run across issue very similar with #479. I did not send another PR.
>
>From what I can understand, this PR is meant to take care of cases where requests can timeout despite getting a response in time, because they are unable to process them before time runs out. Is this correct?

Yes, this is correct. I originally try to resolve "GENERIC_ERROR" caused by ATF test "ATF_AddSubMenu.lua", especially plenty AddSubMenu requests in very short time alike stress testing.

jacobkeeler:
Closed in favor of #909

PR #661, opened 2016-07-01
All lables on one line

LuxoftAKutsan:
Test github integration


LuxoftAKutsan:
Do not actual any more

PR #671, opened 2016-07-05
Cover HMI commands with Unit Tests. Part 14

LevchenkoS:
Has been covered:
- `RequestToHMI`;
- `ResponseFromHMI`;
- `VIIsReadDIDRequest`;
- `VIReadDIDResponse`;
- `VIIsReadyRequest`;
- `VIIsReadyResponse`;
- `VIGetVehicleTypeRequest`;
- `VIGetVehicleTypeResponse`;
- `VISubscribeVehicleDataRequest`;
- `VISubscribeVehicleDataResponse`.

Related to: [APPLINK-25908](https://adc.luxoft.com/jira/browse/APPLINK-25908)

@AByzhynar, @okozlovlux, @Kozoriz, @LuxoftAKutsan, @vlantonov, @wolfylambova22, please review.

### DIFF
--- a/src/components/application_manager/test/commands/hmi/response_from_hmi_test.cc
+++ b/src/components/application_manager/test/commands/hmi/response_from_hmi_test.cc
@@ -1,0 +1,121 @@
+/*
+ * Copyright (c) 2016, Ford Motor Company
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following
+ * disclaimer in the documentation and/or other materials provided with the
+ * distribution.
+ *
+ * Neither the name of the Ford Motor Company nor the names of its contributors
+ * may be used to endorse or promote products derived from this software
+ * without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <stdint.h>
+
+#include "gtest/gtest.h"
+#include "utils/shared_ptr.h"
+#include "smart_objects/smart_object.h"
+#include "application_manager/smart_object_keys.h"
+#include "application_manager/commands/commands_test.h"
+#include "application_manager/commands/command.h"
+#include "application_manager/mock_event_dispatcher.h"
+#include "hmi/response_from_hmi.h"
+
+namespace test {
+namespace components {
+namespace commands_test {
+namespace hmi_commands_test {
+
+using ::testing::_;
+using ::testing::Return;
+using ::testing::SaveArg;
+using ::testing::DoAll;
+
+using ::utils::SharedPtr;
+using ::test::components::event_engine_test::MockEventDispatcher;
+
+namespace am = ::application_manager;
+
+using am::commands::ResponseFromHMI;
+
+typedef SharedPtr<ResponseFromHMI> ResponseFromHMIPtr;
+
+class ResponseFromHMITest : public CommandsTest<CommandsTestMocks::kIsNice> {};
+
+TEST_F(ResponseFromHMITest, BasicMethodsOverloads_SUCCESS) {
+  ResponseFromHMIPtr command(CreateCommand<ResponseFromHMI>());
+
+  // Current implementation always return `true`
+  EXPECT_TRUE(command->Init());
+  EXPECT_NO_THROW(command->Run());
+  EXPECT_TRUE(command->CleanUp());
+}
+
+TEST_F(ResponseFromHMITest, SendResponseToMobile_SUCCESS) {
+  ResponseFromHMIPtr command(CreateCommand<ResponseFromHMI>());
+
+  MessageSharedPtr msg(CreateMessage(smart_objects::SmartType_Map));
+
+  EXPECT_CALL(app_mngr_, ManageMobileCommand(msg, _));
+
+  command->SendResponseToMobile(msg, app_mngr_);
+
+  const application_manager::MessageType kReceivedMessageType =
+      static_cast<application_manager::MessageType>(
+          (*msg)[am::strings::params][am::strings::message_type].asInt());
+
+  EXPECT_EQ(application_manager::MessageType::kResponse, kReceivedMessageType);
+}
+
+TEST_F(ResponseFromHMITest, CreateHMIRequest_SUCCESS) {
+  ResponseFromHMIPtr command(CreateCommand<ResponseFromHMI>());
+
+  MessageSharedPtr result_msg;
+  EXPECT_CALL(app_mngr_, ManageHMICommand(_))
+      .WillOnce(DoAll(SaveArg<0>(&result_msg), Return(true)));
+
+  const hmi_apis::FunctionID::eType kPostedFunctionId =
+      hmi_apis::FunctionID::INVALID_ENUM;
+  MessageSharedPtr dummy_msg_params = CreateMessage();
+  command->CreateHMIRequest(kPostedFunctionId, *dummy_msg_params);
+
+  ASSERT_TRUE(result_msg);
+
+  const application_manager::MessageType kReceivedMessageType =
+      static_cast<application_manager::MessageType>(
+          (*result_msg)[am::strings::params][am::strings::message_type]
+              .asInt());
+
+  EXPECT_EQ(am::MessageType::kRequest, kReceivedMessageType);
+
+  const hmi_apis::FunctionID::eType kReceivedFunctionId =
+      static_cast<hmi_apis::FunctionID::eType>(
+          (*result_msg)[am::strings::params][am::strings::function_id].asInt());
+
+  EXPECT_EQ(kPostedFunctionId, kReceivedFunctionId);
+}
+
+}  // hmi_commands_test
+}  // namespace commands_test
+}  // namespace components
+}  // namespace test
--- a/src/components/application_manager/test/commands/hmi/simple_requests_to_hmi_test.cc
+++ b/src/components/application_manager/test/commands/hmi/simple_requests_to_hmi_test.cc
@@ -1,0 +1,107 @@
+/*
+ * Copyright (c) 2016, Ford Motor Company
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following
+ * disclaimer in the documentation and/or other materials provided with the
+ * distribution.
+ *
+ * Neither the name of the Ford Motor Company nor the names of its contributors
+ * may be used to endorse or promote products derived from this software
+ * without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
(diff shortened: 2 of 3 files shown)


wolfylambova22:
Reviewed.

src/components/application_manager/test/commands/hmi/response_from_hmi_test.cc
```diff
@@ -0,0 +1,121 @@
...
+  EXPECT_TRUE(command->Init());
+  EXPECT_NO_THROW(command->Run());
+}
+
+TEST_F(ResponseFromHMITest, SendResponseToMobile_SUCCESS) {
+  ResponseFromHMIPtr command(CreateCommand<ResponseFromHMI>());
+
+  MessageSharedPtr msg(CreateMessage(smart_objects::SmartType_Map));
+
+  EXPECT_CALL(this->app_mngr_, ManageMobileCommand(msg, _));
```
vlantonov: @LevchenkoS Is `this->` needed?

src/components/application_manager/test/commands/hmi/response_from_hmi_test.cc
```diff
@@ -0,0 +1,121 @@
...
+          (*msg)[am::strings::params][am::strings::message_type].asInt());
+
+  EXPECT_EQ(application_manager::MessageType::kResponse, kReceivedMessageType);
+}
+
+TEST_F(ResponseFromHMITest, CreateHMIRequest_SUCCESS) {
+  ResponseFromHMIPtr command(CreateCommand<ResponseFromHMI>());
+
+  MessageSharedPtr result_msg;
+  EXPECT_CALL(this->app_mngr_, ManageHMICommand(_))
```
vlantonov: @LevchenkoS Is `this->` needed?

src/components/application_manager/test/commands/hmi/simple_requests_to_hmi_test.cc
```diff
@@ -0,0 +1,107 @@
...
+class RequestToHMITest : public CommandsTest<CommandsTestMocks::kIsNice> {};
+
+TEST_F(RequestToHMITest, BasicMethodsOverloads_SUCCESS) {
+  SharedPtr<commands::RequestToHMI> command(
+      CreateCommand<commands::RequestToHMI>());
+
+  // Current implementation always return `true`
+  EXPECT_TRUE(command->CleanUp());
+  EXPECT_TRUE(command->Init());
+  EXPECT_NO_THROW(command->Run());
```
vlantonov: @LevchenkoS It seems that the calls should be like:
`EXPECT_TRUE(command->Init());`
`EXPECT_NO_THROW(command->Run());`
`EXPECT_TRUE(command->CleanUp());`

src/components/application_manager/test/commands/hmi/simple_requests_to_hmi_test.cc
```diff
@@ -0,0 +1,107 @@
...
+}
+
+TEST_F(RequestToHMITest, SendRequest_SUCCESS) {
+  SharedPtr<commands::RequestToHMI> command(
+      CreateCommand<commands::RequestToHMI>());
+
+  EXPECT_CALL(this->app_mngr_, SendMessageToHMI(NotNull()));
```
vlantonov: @LevchenkoS Is `this->` needed?

src/components/application_manager/test/commands/hmi/simple_requests_to_hmi_test.cc
```diff
@@ -0,0 +1,107 @@
...
+              commands::VISubscribeVehicleDataRequest> RequestCommandsList;
+
+TYPED_TEST_CASE(RequestToHMICommandsTest, RequestCommandsList);
+
+TYPED_TEST(RequestToHMICommandsTest, Run_SendMessageToHMI_SUCCESS) {
+  typedef typename TestFixture::CommandType CommandType;
+
+  SharedPtr<CommandType> command = this->template CreateCommand<CommandType>();
+
+  EXPECT_CALL(this->app_mngr_, SendMessageToHMI(NotNull()));
```
vlantonov: @LevchenkoS Is `this->` needed?

src/components/application_manager/test/commands/hmi/simple_response_from_hmi_test.cc
```diff
@@ -0,0 +1,168 @@
...
+  typedef typename TestFixture::CommandData CommandData;
+  typedef typename CommandData::CommandType CommandType;
+
+  hmi_apis::FunctionID::eType kExpectedEventId =
+      static_cast<hmi_apis::FunctionID::eType>(CommandData::kExpectedEventId);
+
+  SharedPtr<CommandType> command = this->template CreateCommand<CommandType>();
+
+  hmi_apis::FunctionID::eType event_id;
+  EXPECT_CALL(this->event_dispatcher_, raise_event(_))
```
vlantonov: @LevchenkoS Is `this->` needed?

src/components/application_manager/test/commands/hmi/response_from_hmi_test.cc
```diff
@@ -0,0 +1,121 @@
...
+class ResponseFromHMITest : public CommandsTest<CommandsTestMocks::kIsNice> {};
+
+TEST_F(ResponseFromHMITest, BasicMethodsOverloads_SUCCESS) {
+  ResponseFromHMIPtr command(CreateCommand<ResponseFromHMI>());
+
+  // Current implementation always return `true`
+  EXPECT_TRUE(command->CleanUp());
+  EXPECT_TRUE(command->Init());
+  EXPECT_NO_THROW(command->Run());
+}
```
vlantonov: @LevchenkoS It seems that the calls should be like:
`EXPECT_TRUE(command->Init());`
`EXPECT_NO_THROW(command->Run());`
`EXPECT_TRUE(command->CleanUp());`

src/components/application_manager/test/commands/hmi/simple_requests_to_hmi_test.cc
LevchenkoS: @vlantonov, yes. It's mostly needed in template classes.

src/components/application_manager/test/commands/hmi/simple_response_from_hmi_test.cc
LevchenkoS: @vlantonov, yes. It's mostly needed in template classes.

Kozoriz:
Reviewed

src/components/application_manager/test/commands/hmi/simple_response_from_hmi_test.cc
```diff
@@ -0,0 +1,168 @@
...
+  ResponseFromHMICommandsTest() {
+    ON_CALL(app_mngr_, event_dispatcher())
+        .WillByDefault(ReturnRef(event_dispatcher_));
+  }
+};
+
+template <class Command, hmi_apis::FunctionID::eType kEventId>
+struct CommandData {
+  typedef Command CommandType;
+  enum { kExpectedEventId = kEventId };
```
vlantonov: @LevchenkoS The enum hack is deprecated. It implicitly converts the enum to `int `You may declare variable constant like this `const hmi_apis::FunctionID::eType kExpectedEventId ` and initialize it in the constructor without conversion to `int` Also further uses of `CommandData::kExpectedEventId` will avoid the implicit conversion from `int` to `hmi_apis::FunctionID::eType`

LevchenkoS: @vlantonov First of all, its should be **static**, cause  objects of that type **never** will be created. 
Secondly: Yes, it's hack as for normal classes/structures. **But**, it's not for template classes/structures. Generally it is the only way to store **static** int constants in the template classes/structures at ease and manageable way. Other ways it will be impossible provide value for **all** of possible combinations for this template structures.

src/components/application_manager/test/commands/hmi/response_from_hmi_test.cc
LevchenkoS: @vlantonov, fixed in 8907a0d.

src/components/application_manager/test/commands/hmi/simple_requests_to_hmi_test.cc
LevchenkoS: @vlantonov, fixed in 8907a0d.

src/components/application_manager/test/commands/hmi/simple_requests_to_hmi_test.cc
LevchenkoS: @vlantonov, fixed in 8907a0d.

src/components/application_manager/test/commands/hmi/response_from_hmi_test.cc
LevchenkoS: @vlantonov, fixed in 8907a0d.

src/components/application_manager/test/commands/hmi/response_from_hmi_test.cc
LevchenkoS: @vlantonov, fixed in 8907a0d.

src/components/application_manager/test/commands/hmi/simple_response_from_hmi_test.cc
vlantonov: @LevchenkoS The constant may not be static. It is good practice to make the class constants static, however in this case it is template parameter which could take different value on class initialization. So you can write const `hmi_apis::FunctionID::eType kExpectedEventId = kEventId` and this will work under Visual Studio and Linux. However if this causes problems under Qt do not make that change.

okozlovlux:
@LevchenkoS reviewed

src/components/application_manager/test/commands/hmi/simple_response_from_hmi_test.cc
```diff
@@ -0,0 +1,168 @@
...
+
+ACTION_P(GetEventFunctionID, function_id) {
+  (*function_id) = arg0.id();
+}
+
+TYPED_TEST(ResponseFromHMICommandsTest, Run_SendMessageToHMI_SUCCESS) {
+  typedef typename TestFixture::CommandData CommandData;
+  typedef typename CommandData::CommandType CommandType;
+
+  hmi_apis::FunctionID::eType kExpectedEventId =
```
LuxoftAKutsan: @LevchenkoS  I don't like that you have variable name that same as enum value.

src/components/application_manager/test/commands/hmi/response_from_hmi_test.cc
LuxoftAKutsan: @LevchenkoS what is the second parameter of MangeMobileCommand? Do we need to check it?

src/components/application_manager/test/commands/hmi/simple_response_from_hmi_test.cc
LevchenkoS: @vlantonov, please look at usage of it (lines 102 - 103):
>  hmi_apis::FunctionID::eType kExpectedEventId =
>  static_cast< hmi_apis::FunctionID::eType >(CommandData::kExpectedEventId);

If I will change it according to your proposal it will be like this(In the bast case):
>  hmi_apis::FunctionID::eType kExpectedEventId = CommandData().kExpectedEventId;

I do not see any sense in creation of new object every single time then I will need just a one variable. Especially then I know it's value on compile time.

LuxoftAKutsan:
Reviewed

src/components/application_manager/test/commands/hmi/simple_response_from_hmi_test.cc
vlantonov: @LevchenkoS `const hmi_apis::FunctionID::eType kExpectedEventId = kEventId;` 
Or if you prefer initialize constant at the constructor.

LevchenkoS: @vlantonov, please, explain me: what sens of your proposal? What do we achieve if I make it as you said?

src/components/application_manager/test/commands/hmi/simple_response_from_hmi_test.cc
vlantonov: @LevchenkoS What is the value of `CommandData::kExpectedEventId` ?

src/components/application_manager/test/commands/hmi/simple_response_from_hmi_test.cc
vlantonov: @LevchenkoS Removal of a deprecate construct and `hmi_apis::FunctionID::eType` to `int` to `hmi_apis::FunctionID::eType` double transformation.

src/components/application_manager/test/commands/hmi/simple_response_from_hmi_test.cc
LevchenkoS: @vlantonov, value of CommandData::kExpectedEventId specified in TypeList `ResponseCommandsList`
So for each type of template `CommandData` in the `ResponseCommandsList` will be created their own copy of typed test `Run_SendMessageToHMI_SUCCESS` and of fixture `ResponseFromHMICommandsTest` which get  `CommandData`  as a template parameter.

src/components/application_manager/test/commands/hmi/simple_requests_to_hmi_test.cc
vlantonov: @LevchenkoS I meant `EXPECT_CALL(this->app_mngr_, SendMessageToHMI(NotNull()));`

src/components/application_manager/test/commands/hmi/simple_response_from_hmi_test.cc
vlantonov: @LevchenkoS I meant `EXPECT_CALL(this->event_dispatcher_, raise_event(_))`

LevchenkoS: @vlantonov, yes, it's required, because event_dispatcher_ is inherited from another template class which take one of template parameters of this template class, so it cant know about its fields and methods.

src/components/application_manager/test/commands/hmi/simple_response_from_hmi_test.cc
LuxoftAKutsan: @vlantonov  kExpectedEventId used as static variable if it will just const variable that will be initialized in constructor, it will be impossible to use it as static variable : ```CommandData::kExpectedEventId```
Also please notice that all this type casting executed at compile time. 
Using enum in template class instead of static variable is rather useful hack, and I cant see how cont variable can solve this problem.

vlantonov: @LevchenkoS `hmi_apis::FunctionID::eType kExpectedEventId = CommandData().kExpectedEventId;` Also please note that the double transformation causes not only time related problems.

src/components/application_manager/test/commands/hmi/simple_response_from_hmi_test.cc
vlantonov: @LevchenkoS The compiler knows about the template parameters - otherwise it will not compile it. You may try.

src/components/application_manager/test/commands/hmi/simple_response_from_hmi_test.cc
LuxoftAKutsan: @vlantonov  please describe problem that can cause this transformation. 
I think that CommandData().kExpectedEventId looks confusing.  You create object to get constant.
Probably we need some other opinion @anosach-luxoft @AGaliuzov what do you think?

vlantonov: @LevchenkoS Loss of precision and portability problems as the size of the `int` may vary.

src/components/application_manager/test/commands/hmi/simple_response_from_hmi_test.cc
vlantonov: @LevchenkoS Thank you for the answer.

LevchenkoS: @LuxoftAKutsan, @vlantonov, fixed in f3c06ec.

src/components/application_manager/test/commands/hmi/response_from_hmi_test.cc
LevchenkoS: @LuxoftAKutsan, second parameter of `MangeMobileCommand` is `CommandOrigin`.

LevchenkoS:
@vlantonov, @LuxoftAKutsan, Static cast should be returned
because gcc on build server does
not allow to implisetly convert
enum of one type to enum of another type.

>simple_response_from_hmi_test.cc:95:22: 
error: comparison between const enum CommandData< VIReadDIDResponse, (hmi_apis::FunctionID::eType)109 >::EventId and const EventID {aka const enum hmi_apis::FunctionID::eType} [-Werror=enum-compare]
   return function_id == arg.id();

vlantonov:
@LevchenkoS This means that still we have the double conversion I would like to be avoided. However this opinion is not supported by other reviewers so I accept this change. Reviewed.

PR #683, opened 2016-07-11
Fix invalid memory access.

vlantonov:
Fixed access to deleted fields in Timer and StreamerAdapter.

Fixed uninitialized values in request_controller.

Related Issue: APPLINK-25098

### DIFF
--- a/src/components/application_manager/src/request_controller.cc
+++ b/src/components/application_manager/src/request_controller.cc
@@ -190,8 +190,8 @@ RequestController::TResult RequestController::addHMIRequest(
 
   const uint64_t timeout_in_mseconds =
       static_cast<uint64_t>(request->default_timeout());
-  RequestInfoPtr request_info_ptr(
-      new HMIRequestInfo(request, timeout_in_mseconds));
+  RequestInfoPtr request_info_ptr =
+      utils::MakeShared<HMIRequestInfo>(request, timeout_in_mseconds);
 
   if (0 == timeout_in_mseconds) {
     LOG4CXX_DEBUG(logger_,
@@ -446,8 +446,8 @@ void RequestController::Worker::threadMain() {
                                           // default timeout
 
     const uint32_t timeout_in_mseconds = request_ptr->default_timeout();
-    RequestInfoPtr request_info_ptr(
-        new MobileRequestInfo(request_ptr, timeout_in_mseconds));
+    RequestInfoPtr request_info_ptr =
+        utils::MakeShared<MobileRequestInfo>(request_ptr, timeout_in_mseconds);
 
     request_controller_->waiting_for_response_.Add(request_info_ptr);
     LOG4CXX_DEBUG(logger_, "timeout_in_mseconds " << timeout_in_mseconds);
--- a/src/components/utils/src/timer.cc
+++ b/src/components/utils/src/timer.cc
@@ -47,8 +47,8 @@ timer::Timer::Timer(const std::string& name, TimerTask* task)
     : name_(name)
     , task_(task)
     , state_lock_()
-    , delegate_(this, state_lock_)
-    , thread_(threads::CreateThread(name_.c_str(), &delegate_))
+    , delegate_(new TimerDelegate(this, state_lock_))
+    , thread_(threads::CreateThread(name_.c_str(), delegate_.get()))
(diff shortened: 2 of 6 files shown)


vlantonov:
Please review: @AByzhynar, @AGaliuzov , @anosach-luxoft , @VProdanov, @wolfylambova22

src/components/utils/src/timer.cc
```diff
@@ -64,6 +64,8 @@ timer::Timer::~Timer() {
   StopDelegate();
   single_shot_ = true;
 
+  thread_->set_delegate(NULL);
```
anosach-luxoft: @vlantonov you don't need to call `set_delegate(NULL)` explicitly. It will be done by `ThreadDelegate` destructor

src/components/application_manager/src/request_controller.cc
```diff
@@ -190,8 +190,9 @@ RequestController::TResult RequestController::addHMIRequest(
 
   const uint64_t timeout_in_mseconds =
       static_cast<uint64_t>(request->default_timeout());
-  RequestInfoPtr request_info_ptr(
-      new HMIRequestInfo(request, timeout_in_mseconds));
+  RequestInfoPtr request_info_ptr =
+      utils::MakeShared<HMIRequestInfo>(request, timeout_in_mseconds);
+  request_info_ptr->set_hmi_level(mobile_apis::HMILevel::INVALID_ENUM);
```
anosach-luxoft: @vlantonov it's not clear for me how these changes relate to core crash on USB unplug. Maybe add some explanation to PR description?

anosach-luxoft:
Reviewed

src/components/application_manager/src/request_controller.cc
vlantonov: @anosach-luxoft After I failed to reproduce the bad_alloc() exception I decided to do a check with a `valgrind` in order to reveal hidden memory access problems which may cause the observed crash. I discovered usage of uninitialized variable and after a consultation with @dev-gh I added the discussed line to which corresponds the `Fixed uninitialized values in request_controller.` line in description of commit. I am not sure if this may cause the observed crash but definitely this is a bug and it should be fixed.

src/components/utils/src/timer.cc
vlantonov: @anosach-luxoft I am aware of the `ThreadDelegate` destructor behavior however it is unintuitive and is not documented. There is no clear indication that it will stand the same so I decided to take precautions.

AByzhynar:
Reviewed

src/components/application_manager/src/request_controller.cc
anosach-luxoft: @vlantonov ok

src/components/utils/src/timer.cc
anosach-luxoft: @vlantonov I think these calls are redundant, but doublecheck `ThreadDelegate`'s destructor if you will remove them.

VProdanov:
Reviewed

src/components/application_manager/src/request_controller.cc
AGaliuzov: @vlantonov for me it looks like we have to set `INVALID_ENUM` to the `HMIRequestInfo` in constructor. As result you can avoid additional call to `set_hmi_level` function

src/components/utils/src/timer.cc
vlantonov: @anosach-luxoft Checked - `ThreadDelegate` 's destructor sets `thread_ `delegate to `NULL`. Fixed in b1d62949f85705c676e053167f589ac9f1e982dd

src/components/application_manager/src/request_controller.cc
vlantonov: @anosach-luxoft @AGaliuzov Fixed in b1d62949f85705c676e053167f589ac9f1e982dd

wolfylambova22:
Reviewed.

AGaliuzov:
Reviewed

AByzhynar:
@vlantonov Were you going to squash commits?

vlantonov:
@AByzhynar Commits squashed.

PR #698, opened 2016-07-15
Fix_core_dump_due_to_connection_been_created_twice

VProdanov:
The reason for the core dump, is that ConnectionCreated() is
been called twice, from parent and child threads while the
transport_adapter is restoring the last state.
First is called in tcp_connection_factory.cc then in
threaded_socket_connection.cc from the child threadMain() with
connection->Start().

Related-issues:
APPLINK-25379, APPLINK-22021

### DIFF
--- a/src/components/transport_manager/test/transport_manager_default_test.cc
+++ b/src/components/transport_manager/test/transport_manager_default_test.cc
@@ -41,11 +41,12 @@ namespace components {
 namespace transport_manager_test {
 
 using ::testing::Return;
+
 TEST(TestTransportManagerDefault, Init_LastStateNotUsed) {
   MockTransportManagerSettings transport_manager_settings;
   transport_manager::TransportManagerDefault transport_manager(
       transport_manager_settings);
-  resumption::LastState last_state("app_storage_folder", "app_info_storage");
+  resumption::LastState last_state("app_storage_folder", "app_info_storage2");
 
   EXPECT_CALL(transport_manager_settings, use_last_state())
       .WillRepeatedly(Return(false));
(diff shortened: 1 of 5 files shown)


VProdanov:
Please review @AGaliuzov @AGritsevich @vlantonov

AGritsevich:
Reviewed.

AGaliuzov:
Reviewed

vlantonov:
Reviewed.

VProdanov:
Please review @anosach-luxoft

src/components/transport_manager/test/transport_manager_default_test.cc
```diff
@@ -66,6 +65,7 @@ TEST(TestTransportManagerDefault, DISABLED_Init_LastStateUsed) {
   EXPECT_CALL(transport_manager_settings, transport_manager_tcp_adapter_port())
       .WillRepeatedly(Return(1u));
   transport_manager.Init(last_state);
+  usleep(3000);
```
anosach-luxoft: @VProdanov we're trying to reduce `usleep` usages in project. Can you avoid it here? BTW `usleep` is not crosspatform and would not be compiled on Windows platform

VProdanov: @anosach-luxoft can I use some alternative, because I noticed when the test finish there is one tread that in not called.. and when it become active we are in other test..

anosach-luxoft: @VProdanov in this case you could use conditional variable for example to wait thread. Could you?

anosach-luxoft:
@VProdanov I see you've removed `usleep()` call and it's great, but are you sure you will not get an issue with thread again? Have you fix it completely?

VProdanov:
@anosach-luxoft I added delay with this dynamic allocations and since then it always pass. But probably is environment dependent. I will think about some alternatives.

anosach-luxoft:
Reviewed

AByzhynar:
@VProdanov Please add file extension to `app_info_storage2` file

PR #702, opened 2016-07-18
Fix WARNINGS response when AppHMIType received

### LINKED ISSUE #647: RegisterAppInterface returns WARNING message if app has AppHMIType
1. Added HMItype in app policies
2. send AppHMIType from mobile app 

RegisterAppInterface responses WARNING with message
""info" : "Following AppHMITypes are not present in policy table:5,""

jacobkeeler:
When comparing the lists of AppHMITypes from the policy table and RegisterAppInterface message, the values are mistakenly read from the message object as strings when they are actually in enum form, resulting in a warning even if both lists include the same AppHMIType. 

This PR changes this so that the enum values are converted to strings when comparing the two lists, so that they are compared correctly.

Fixes #647

### DIFF
--- a/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
+++ b/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
@@ -89,11 +89,8 @@ std::string AppHMITypeToString(mobile_apis::AppHMIType::eType type) {
 
   std::map<mobile_apis::AppHMIType::eType, std::string>::const_iterator iter =
       appHMITypeMap.find(type);
-  if (appHMITypeMap.end() == iter) {
-    return "";
-  } else {
-    return iter->second;
-  }
+  
+  return appHMITypeMap.end() == iter ? iter->second : std::string("");
 }
 
 struct AppHMITypeInserter {
--- a/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
+++ b/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
@@ -36,7 +36,6 @@
 #include <unistd.h>
 #include <algorithm>
 #include <map>
-#include <stdexcept>
 #include <string.h>
 
 #include "application_manager/application_manager_impl.h"
@@ -87,11 +86,13 @@ std::string AppHMITypeToString(mobile_apis::AppHMIType::eType type) {
     {mobile_apis::AppHMIType::TESTING, "TESTING"},
     {mobile_apis::AppHMIType::SYSTEM, "SYSTEM"}
   };
-  try {
-    return appHMITypeMap.at(type);
-  }
-  catch (const std::out_of_range &ex) {
+
+  std::map<mobile_apis::AppHMIType::eType, std::string>::const_iterator iter =
+      appHMITypeMap.find(type);
+  if (appHMITypeMap.end() == iter) {
     return "";
+  } else {
+    return iter->second;
   }
 }
 
@@ -116,7 +117,8 @@ struct CheckMissedTypes {
       : policy_app_types_(policy_app_types), log_(log) {}
 
   bool operator()(const smart_objects::SmartArray::value_type &value) {
-    std::string app_type_str = AppHMITypeToString(static_cast<mobile_apis::AppHMIType::eType>(value.asInt()));
+    std::string app_type_str = AppHMITypeToString(
+        static_cast<mobile_apis::AppHMIType::eType>(value.asInt()));
     if (!app_type_str.empty()) {
       policy::StringArray::const_iterator it = policy_app_types_.begin();
       policy::StringArray::const_iterator it_end = policy_app_types_.end();
--- a/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
+++ b/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
@@ -72,6 +72,32 @@ mobile_apis::AppHMIType::eType StringToAppHMIType(const std::string &str) {
   }
 }
 
+std::string AppHMITypeToString(mobile_apis::AppHMIType::eType type) {
+  if (mobile_apis::AppHMIType::DEFAULT == type) {
+    return "DEFAULT";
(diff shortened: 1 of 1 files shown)


LuxoftAKutsan:
@jacobkeeler please make proper commit name

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -72,6 +72,32 @@ mobile_apis::AppHMIType::eType StringToAppHMIType(const std::string &str) {
   }
 }
 
+std::string AppHMITypeToString(mobile_apis::AppHMIType::eType type) {
+  if (mobile_apis::AppHMIType::DEFAULT == type) {
```
LuxoftAKutsan: @jacobkeeler can you use switch?

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -72,6 +72,32 @@ mobile_apis::AppHMIType::eType StringToAppHMIType(const std::string &str) {
...
+  } else if (mobile_apis::AppHMIType::SOCIAL == type) {
+    return "SOCIAL";
+  } else if (mobile_apis::AppHMIType::BACKGROUND_PROCESS == type) {
+    return "BACKGROUND_PROCESS";
+  } else if (mobile_apis::AppHMIType::TESTING == type) {
+    return "TESTING";
+  } else if (mobile_apis::AppHMIType::SYSTEM == type) {
+    return "SYSTEM";
+  } else {
+    return NULL;
```
LuxoftAKutsan: @jacobkeeler This will cause core crash.  Probably we should have there some DCHECK or at lease log error

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -93,7 +119,7 @@ struct CheckMissedTypes {
       : policy_app_types_(policy_app_types), log_(log) {}
 
   bool operator()(const smart_objects::SmartArray::value_type &value) {
-    std::string app_type_str = value.asString();
+    std::string app_type_str = AppHMITypeToString((mobile_apis::AppHMIType::eType)value.asUInt());
```
LuxoftAKutsan: @jacobkeeler  please avoid using C-cast. Also value can be INVALID_ENUM, that is -1. So using it as unsigned int is not correct.

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
jacobkeeler: I could do so, but there is a reverse function (StringToAppHMIType), already in the file which also uses if statements [here](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc#L49-73). 

Do you just want me to change both of them to switch statements?

jacobkeeler:
@LuxoftAKutsan All notes were fixed, please check changes

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -73,28 +73,20 @@ mobile_apis::AppHMIType::eType StringToAppHMIType(const std::string &str) {
...
+    case mobile_apis::AppHMIType::MEDIA:              return "MEDIA";
+    case mobile_apis::AppHMIType::MESSAGING:          return "MESSAGING";
+    case mobile_apis::AppHMIType::NAVIGATION:         return "NAVIGATION";
+    case mobile_apis::AppHMIType::INFORMATION:        return "INFORMATION";
+    case mobile_apis::AppHMIType::SOCIAL:             return "SOCIAL";
+    case mobile_apis::AppHMIType::BACKGROUND_PROCESS: return "BACKGROUND_PROCESS";
+    case mobile_apis::AppHMIType::TESTING:            return "TESTING";
+    case mobile_apis::AppHMIType::SYSTEM:             return "SYSTEM";
+    default:
+      NOTREACHED();
```
LuxoftAKutsan: @jacobkeeler NOTREACHED will be switched off in release mode, and you will just try to return NULL.
Compile will generate std::string(NULL) that will cause core crash,

AGaliuzov: @jacobkeeler I would rather suggest to create `map` with values instead of switch

jacobkeeler: @LuxoftAKutsan Fixed in 54fda30

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -73,20 +74,23 @@ mobile_apis::AppHMIType::eType StringToAppHMIType(const std::string &str) {
...
+    {mobile_apis::AppHMIType::MEDIA, "MEDIA"},
+    {mobile_apis::AppHMIType::MESSAGING, "MESSAGING"},
+    {mobile_apis::AppHMIType::NAVIGATION, "NAVIGATION"},
+    {mobile_apis::AppHMIType::INFORMATION, "INFORMATION"},
+    {mobile_apis::AppHMIType::SOCIAL, "SOCIAL"},
+    {mobile_apis::AppHMIType::BACKGROUND_PROCESS, "BACKGROUND_PROCESS"},
+    {mobile_apis::AppHMIType::TESTING, "TESTING"},
+    {mobile_apis::AppHMIType::SYSTEM, "SYSTEM"}
+  };
+  try {
```
LuxoftAKutsan: @jacobkeeler we use Google Stle guide and is doesn't allow exceptions 
https://google.github.io/styleguide/cppguide.html#Exceptions

jacobkeeler: @LuxoftAKutsan Alright, style issues should be fixed in 1309e08

LuxoftAKutsan:
Reviewed

PR #719, opened 2016-07-25
Limit SDl log message queue size

LuxoftAKutsan:
In case if log message queue becomes bigger that LOG_MESSAGE_QUEUE_SIZE
environment variable, SDL waiting while log message queue won't be
flushed.

This it temporary fix to avoid memory leaks and should be reworked.

Related issues : APPLINK-17599, APPLINK-22887


pestOO:
@LuxoftAKutsan, this changes need a new SDL requirements and need to be approved by customer,
Also,
* need to be clarified SLD behaviors in case of over limitations
* limitation need to be a configurable variable (ini file)

src/components/utils/include/utils/log_message_loop_thread.h
```diff
@@ -52,9 +54,24 @@ typedef struct {
...
+ public:
+  // Places a message to the therad's queue. Thread-safe.
+  void PostMessage(const Message& message) OVERRIDE {
+    threads::MessageLoopThread<LogMessageQueue>::PostMessage(message);
+
+    // LOG_MESSAGE_QUEUE_SIZE defined in Utils CMakeLists
+    if (GetMessageQueueSize() > LOG_MESSAGE_QUEUE_SIZE) {
+      // TODO (AKutsan): Avoid stopping thread if message queue is full.
+      // Prefere Handle multiple messages from Message loop thread
+      WaitDumpQueue();
```
pestOO: @LuxoftAKutsan, inconsistence of the following message need to be forced logged.
I assume, that logging in case of over-limitation logging sounds strange, but I is mandatory to save a state that all following messages are not trusted any more.
As a one case of implementation - we can log this fatal ones per SDL life_cycle.

LuxoftAKutsan: @pestOO You prefere log somethik like
```FATAL : Some log messages may be missed```
and do not lg them?

**Comment by LuxoftAKutsan** [comment]
@pestOO I completely agree with you. And we will create task to clarify requirements, and we will choose better solution. But now we need to resolve 2 critical defects APPLINK-17599 and APPLINK-22887 so we decide to do such ugly hotfix.

File: src/components/utils/include/utils/log_message_loop_thread.h (hunk 1 above)
Comment (pestOO) [inline src/components/utils/include/utils/log_message_loop_thread.h]: @LuxoftAKutsan, 
Sorry, I've missed that we are waiting, not dropping packets.
In this case we can face with a original issue with slow reading from devices and corrupt all income data from devices (iap and aoa definitely)
So, this fix will mask issue with non-real-time reading from-to external devices!

**Comment by pestOO** [comment]
@LuxoftAKutsan, 
Please, add `Note` with clarification ticket in new code and link to the defects tickets also

**Comment by LuxoftAKutsan** [comment]
@pestOO defect link in TODO will be added as soon as defect will be created. Of cource

**Comment by LuxoftAKutsan** [comment]
@anosach-luxoft @pestOO @AByzhynar @okozlovlux @CallousOdesa @nk0leg please review

**Comment by okozlovlux** [comment]
@LuxoftAKutsan , reviewed

File: src/components/utils/include/utils/log_message_loop_thread.h [hunk 2]
```diff
@@ -35,6 +35,8 @@
 
 #include <string>
 #include <queue>
+#include <list>
+#include <deque>
```
AByzhynar: @LuxoftAKutsan Why did you added includes above? I can't see appropriate usage .

AByzhynar:
@LuxoftAKutsan Please figure out why almost all the checks failed.

Kozoriz:
Reviewed

LuxoftAKutsan:
Closed because it is bad solution

PR #722, opened 2016-07-26
Check for OpenSSL and UDev while running cmake

### LINKED ISSUE #601: libssl-dev and libudev-dev are not caught by cmake
cmake should fail if these libraries are not installed.

jacobkeeler:
Fixes #601

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -199,6 +199,7 @@ endif()
 
 if (BUILD_USB_SUPPORT)
   add_definitions(-DUSB_SUPPORT)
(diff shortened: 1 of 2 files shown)


JackLivio:
@jacobkeeler Is there a cmake find script needed for openSSL like you included for libudev?

jacobkeeler:
It is one of the existing modules that are included with cmake, a list of these can be produced using:
```
cmake --help-module-list
```
https://cmake.org/Wiki/CMake:How_To_Find_Libraries

JackLivio:
Reviewed and merged

PR #730, opened 2016-07-28
Coverity fix unchecked return values

BTPankow:
These commits are intended to address Coverity defects warning about unchecked return values.

In many instances, the return value was checked and an errno was logged.  Typically, no important logic was added as a result of the check.  This means that this pull request may not be critical but could be nice to have to increase logging verbosity and address Coverity issues.


BTPankow:
Added another commit for a similar defect I forgot to address.

PR #734, opened 2016-08-02
Change_VerifyPeer_parameter_in_ini_file_to_true

VProdanov:
By implementing this feature we fix secure service opening
without validation. The problem which is occuring is that
when we remove the server cridentials from SDL and start
secure service we succeed. The reason for that is the VerifyPeer
located in .ini file which by default was set false. As result
SSL_CTX_set_verify is called without SSL_VERIFY_PEER, 
which is the reason server`s cridentials to not be checked.

Requirement: APPLINK-22607
Fix-issue: APPLINK-17757

### DIFF
--- a/src/appMain/smartDeviceLink.ini
+++ b/src/appMain/smartDeviceLink.ini
@@ -156,7 +156,7 @@ SSLMode         = CLIENT
 ;CipherList      = AES256-GCM-SHA384
 CipherList      = ALL
 ; Verify Mobile app certificate (could be used in both SSLMode Server and Client)
-VerifyPeer  = false
+VerifyPeer  = true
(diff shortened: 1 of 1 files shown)


VProdanov:
Please review @VVeremjova @pvvasilev @wolfylambova22 @AGaliuzov @vlantonov @anosach-luxoft

vlantonov:
@VProdanov Reviewed

VVeremjova:
Reviewed

pvvasilev:
Reviewed.

anosach-luxoft:
Reviewed

wolfylambova22:
Reviewed.

PR #744, opened 2016-08-04
Send TIMED_OUT result code upon no HMI response

BTPankow:
When the HMI does not respond to a request within a certain amount of
time, a GENERIC_ERROR message was being created internally by SDL to be
sent to mobile.  This changes the GENERIC_ERROR to a more descriptive
TIMED_OUT.

This addresses the issue of GENERIC_ERRORs being created instead of timeouts for #479.  The cause of timeout for that issue appears to be due to HMI implementation and lack of response to SDL core.


anosach-luxoft:
@BTPankow do you have requirements that confirm such SDL behavior?

BTPankow:
@anosach-luxoft Thanks for the tip, it looks like integration guidelines define different behavior on time out.  I'll close this PR, since it does go against the requirements.

PR #745, opened 2016-08-05
Sdl behaviour if VR not avaliable

LuxoftAKutsan:
Created skeleton for supporting disabled interfaces on HMI side. 
Implemented supporting disabled VR inteface

PR #759, opened 2016-08-16
SDL behaviour if VR is not avaliable

LuxoftAKutsan:
Is VR interface is not available SDL should process VR related requests according rules described in 
CRQ APPLINK-20918 [GENIVI] VR interface: SDL behavior in case HMI does not respond to IsReady_request or respond with "available" = false


LuxoftAKutsan:
All comments were in PR #745

LuxoftAKutsan:
@AGaliuzov please merge

PR #771, opened 2016-08-22
Fix/update logger properties to log only to file

orhan-mehmedov:
Update logger properties to log to file only

Leave only logging to file in log4cxx.properties
due to slow printing to console/terminal.
Workaround for APPLINK-17599.

Related to [APPLINK-17599.](https://adc.luxoft.com/jira/browse/APPLINK-17599)


orhan-mehmedov:
Created by mistake.

PR #535, opened 2016-05-11
Re-adds policy_manager_impl_stress_test.cc with fix

JackLivio:
Adds "device" and "pre_DataConsent" to the app_policies section when the test case is building a sample policy table.

PR #526, opened 2016-05-02
Policy Manager Impl Tests Fix

JackLivio:
Uncommented out policy tests so policy unit tests are included in the build. Also fixed one of the test cases to include the necessary parts for app_policies in the 'LoadPT_SetPT_PTIsLoaded' test. "make test" now passes.

Fixes #517

PR #780, opened 2016-08-25
Test "device" and "pre_DataConsent" must be in PT

orhan-mehmedov:
Implemented explicit UT for testing CRQ26921.
Policies: "device" and "pre_DataConsent"
sections must be present in PT.

Related to APPLINK-26927.

### DIFF
--- a/src/components/policy/test/policy_manager_impl_test.cc
+++ b/src/components/policy/test/policy_manager_impl_test.cc
@@ -574,6 +574,104 @@ TEST_F(PolicyManagerImplTest, LoadPT_SetInvalidUpdatePT_PTIsNotLoaded) {
   EXPECT_FALSE(manager->LoadPT("file_pt_update.json", msg));
 }
 
+// [GENIVI] Policies: "device" and "pre_DataConsent" sections
+// must be present in PT
+TEST_F(PolicyManagerImplTest,
+       CRQ26921_LoadPT_SetInvalidUpdatePT_PTIsNotLoaded) {
+  // Arrange
+  const std::string load_table(
+      "{"
+      "\"policy_table\": {"
+      "\"module_config\": {"
+      "\"preloaded_pt\": true,"
+      "\"exchange_after_x_ignition_cycles\": 10,"
+      "\"exchange_after_x_kilometers\": 100,"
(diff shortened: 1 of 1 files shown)


orhan-mehmedov:
Please review,
@AByzhynar @LuxoftAKutsan @anosach-luxoft @dev-gh @Kozoriz @pvvasilev
Optional: @pestOO 
// @wolfylambova22 is on vacation

dev-gh:
Reviewed

src/components/policy/test/policy_manager_impl_test.cc
```diff
@@ -574,6 +574,104 @@ TEST_F(PolicyManagerImplTest, LoadPT_SetInvalidUpdatePT_PTIsNotLoaded) {
   EXPECT_FALSE(manager->LoadPT("file_pt_update.json", msg));
 }
 
+// [GENIVI] Policies: "device" and "pre_DataConsent" sections
+// must be present in PT
+TEST_F(PolicyManagerImplTest,
+       CRQ26921_LoadPT_SetInvalidUpdatePT_PTIsNotLoaded) {
```
AByzhynar: @orhan-mehmedov Bad unit-test case name. From the name of UT it should be clear what it tests. Nobody can understand what does this `CRQ26921` mean

AByzhynar:
@orhan-mehmedov I think it's worth to add positive tests and negative. So as result you will get valid PT if necessary fields present and also you will get invalid PT if necessary fields are absent. It seems to me that tests can be shorter. I am not sure that it's needed to call LoadPT.

orhan-mehmedov:
Closing justification:
Based on suggestions in:
https://github.com/smartdevicelink/sdl_core/pull/526
https://github.com/smartdevicelink/sdl_core/pull/535
PR526 is already done in LoadPT_SetPT_PTIsLoaded / policy_manager_impl_test.cc.
PR535 can not be done because "policy_manager_impl_stress_test.cc" is removed from build due to build issues and other errors. 
I've tried to add particular test here, but actually testing combinations of "S13j_Applink_Policy_Table_Data_Dictionary_046.xlsx" require more refactoring and work so it is out of scope of this task. Created APPLINK-27414 for further investigation and work in this direction.

PR #786, opened 2016-08-29
Fix instruction about SDL_SQLITE_DIR

orhan-mehmedov:
There is an error in current instructions.
SDL_SQLITE_DIR is used in make_sqlite_lib.cmd to prepare sqlite3.lib in the binary folder by using sqlite3.dll.
So, it is not related to sqlite source folder but binary one.


LuxoftAKutsan:
@orhan-mehmedov It is not good to send pull requests to open source repository with branches names patch-3. Please recreate pull requests with appropriate branch name. Probzbly you can join changes from #786 #787 #788 in one pull request

PR #826, opened 2016-09-19
Reduce compilation time

LevchenkoS:
Added [Cotire](https://github.com/sakra/cotire) CMake module

Cotire (compile time reducer) is a CMake module
that speeds up the build process of CMake based
build systems by fully automating techniques as
precompiled header usage and single compilation
unit builds for C and C++.

This module secured by MIT license, which is does
not contradict with license of SDL. Plus, SDL already
contains some third party components secured by
this license (For example jsoncpp).

-
Single compilation unit builds has been applied for 
am_commands_test target, to reduce compilation time.
Added additional namespaces to avoid name 
collisions in single unit build mode.

-
Incorrect test cases has been fixed.

-
Related to: [APPLINK-27752](https://adc.luxoft.com/jira/browse/APPLINK-27752)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,6 +65,7 @@ option(ENABLE_SECURITY "Security Ford protocol protection" ON)
 option(ENABLE_HMI_PTU_DECRYPTION "Policy table update parsed by hmi" ON)
 option(ENABLE_EXTENDED_POLICY "Turns extended flow which requires embedded system interaction" ON)
 option(ENABLE_THREAD_COUNT "Turn on correct threads shutdown checks" OFF)
+option(FAST_BUILD "Turn on fast build" ON)
 
 # Platform - specific build flags
 if(${OS} STREQUAL "posix")
--- a/src/components/application_manager/test/CMakeLists.txt
+++ b/src/components/application_manager/test/CMakeLists.txt
@@ -166,11 +166,10 @@ list(APPEND LIBRARIES
 
 create_test(am_hmi_capabilities_test "${AM_HMI_CAPABILITIES_SOURCES}" "${LIBRARIES}")
 create_test(am_application_test "${APPLICATION_SOURCES}" "${LIBRARIES}")
-create_test(am_commands_test "${COMMANDS_SOURCES}" "${LIBRARIES}")
 create_test(am_resumption_test "${RESUMPTION_SOURCES}" "${LIBRARIES}")
 create_test(am_state_controller_test "${STATE_CONTROLLER_SOURCES}" "${LIBRARIES}")
 create_test(am_hmi_language_handler_test "${AM_HMI_LANGUAGE_HANDLER}" "${LIBRARIES}")
-
+create_cotired_test(am_commands_test "${COMMANDS_SOURCES}" "${LIBRARIES}")
 
 list(APPEND LIBRARIES
   application_manager
--- a/src/components/application_manager/test/commands/command_impl_test.cc
+++ b/src/components/application_manager/test/commands/command_impl_test.cc
@@ -47,6 +47,7 @@
 namespace test {
 namespace components {
 namespace commands_test {
+namespace command_impl {
 
 using ::testing::Return;
 using ::testing::AtLeast;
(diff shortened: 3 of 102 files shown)


LevchenkoS:
@AGaliuzov, @Kozoriz, @VProdanov, @wolfylambova22, please review.

pestOO:
@LevchenkoS 
Could we load `cotire` as a submodule?
I don't see any reasons to load it directly to sdl repository.

pestOO:
@LevchenkoS 
What's about setting Cotire for other components test?
Will it sufficiently increase speed?

LevchenkoS:
@pestOO, Yes, it can reduce compile time, especially if it need to compile and link many object files now.

pestOO:
@LevchenkoS 
So, will we update creation other tests?
*Note:* Possibly in addition task after approbation this solution with AM

LevchenkoS:
@pestOO, Yes, sure. And also it can be applied not just for tests, but for libraries like `am_commands_library`. But now we got an issue with Windows and Qt builds. So until it will be resolved, this feature will be available only for Linux build.

pestOO:
@LevchenkoS 
What's about using submodule?

LevchenkoS:
@pestOO, currently SDL doesn't use any sub-modules, so I don't see any reasons why this should be used for such a small tool. So, as result, it will force everyone to perform additional actions to use SDL, because of it, which is pretty excessively in this kind of situation.

pestOO:
@LevchenkoS 
We are loading OpenSSL repository (at windows at least)
In near future I hope we will start loading GTest as a submodule.
So, I think it's good choice to start usign submodules instead of loading raw files and increasing SDL repo size without any reasons.

@AGaliuzov, @malirod, please participate in discussion.

malirod:
@pestOO,
Regarding OpenSSL. Now we download and build OpenSSL because it's not available in the same way as on Linux. On linux we use bin version and this speed up the whole build. I don't think that we will use submodules for OpenSSL since there is no good and single solution for all platforms.

Regarding submodules for gtest. Once we will be able to use gtest from the official repo directly (existing customizations will be removed) there will be a possibility to use gtest via submodules. But I'm not sure that this is wise. It's much better to use "fused gtest" (see https://github.com/google/googletest/blob/master/googlemock/scripts/fuse_gmock_files.py) . This will improve gtest build time(there will be only a single cc file with both gtest and gmock included and two headers). Since we don't upgrade gtest very often then the fused version looks more promising to me.

In general, the idea to add this cmake plugin as sub-module looks not consistent with the current approach used in the SDL. At the moment we have special dir `Modules` with some cmake plugins. We don't have sub-modules at the moment. So I see no point to start using sub-modules at this particular case.

anosach-luxoft:
@LevchenkoS I think `tools/cmake/helpers` is more appropriate place for `cotire` source file

LevchenkoS:
@anosach-luxoft, as for me `cotire` is more then a helper. It's complicated tool which allow to easily decrease compile time and use some features, which CMake doesn't support from the box.
Whereas helpers are just a wrappers over standard CMake functionality, to make CMake files look more clear.

anosach-luxoft:
@LevchenkoS but `modules` is a directory specially for cmake `Find` modules. If you don't like the `helpers` dir, please create separate with appropriate name for such tools

LevchenkoS:
@anosach-luxoft, in my opinion, all CMake include files which add new functionality, or extend exiting one, should be pleased in `modules` folder. Whereas all CMake include files which just wraps already existing functionality, to simplify its usage, should be inside of `helpers` folder.

AGaliuzov:
@LevchenkoS Absolutely agree with @anosach-luxoft it is helper - it helps us to reduce compile time.
So the `helper` directoty is appropriate place for this util

AGaliuzov:
@pestOO I agree with @malirod and @LevchenkoS 
Currently using the submodule will make the system more complex.
Some stuff works out of box for other stuff we have to init submodules - it will confuse.

Lets just keep it as a helper

LevchenkoS:
@pestOO, @anosach-luxoft, please, finish review.

Issue #805, opened 2016-09-07
[SDLcore 4.1 release] iOS apps getting onHMIstatus NONE after 1 second of onHMIStatus FULL


madhuy:
1. start iOS app
2. Mobile app receives onHMIStatus FULL
3. after 1 sec Mobile app  receives onHMIstatus NONE.

my question do we need to change any thing is HS to compatible with SDLCore 4.1.0.
same works fine with android.
attaching logs
[SmartDeviceLinkCore.txt](https://github.com/smartdevicelink/sdl_core/files/459972/SmartDeviceLinkCore.txt)

joeygrover:
@NicoleYarroch @joeljfischer can you take a look at this?

So Android apps do not receive the HMIStatus NONE after FULL?

madhuy:
@joeygrover Android app don't receive NONE after FULL

joeljfischer:
What version of sdl_ios, what version of iOS?

madhuy:
sdl_ios 4.2.2
iOS 9.3.2

more info, it works fine  if we launch app manually.
 it is not working, if app been launched by using onSystemRequest(LAUNCH_APP).

madhuy:
@joeljfischer any update on this?

BTPankow:
@madhuy I'm taking a look at this.  It seems as though the iOS app sends a OnHMIStatus notification with `hmiLevel = FULL` and  core responds with an `hmiLevel = NONE`.  This can be seen in the attached log excerpt ([OnHMIStatus.txt](https://github.com/smartdevicelink/sdl_core/files/468098/OnHMIStatus.txt)).  Could this be what is happening?

Would you be able to supply iOS logs that show the issue?

madhuy:
Attaching SDLcore and iOS logs
[ios_onHMIStatus_notworking.txt](https://github.com/smartdevicelink/sdl_core/files/468252/ios_onHMIStatus_notworking.txt)
[sdlcore_onHMIStatus_notwoking.txt](https://github.com/smartdevicelink/sdl_core/files/468253/sdlcore_onHMIStatus_notwoking.txt)

BTPankow:
@madhuy Thanks for the logs, it definitely looks like something strange is occurring.

What HMI are you using?  Does this happen with the [sdl_hmi](https://github.com/smartdevicelink/sdl_hmi)?

madhuy:
[...]

PR #829, opened 2016-09-20
Fix HMI resumption after activation

BTPankow:
This pull request is a fix for issue #805.

If an HMI application is activated to HMI level FULL, it could still be
resumed which causes the application to then receive HMI level NONE.

This is because the HMI level is set during the initial resumption
function using `SetupDefaultHMILevel` which currently only sets the HMI
level to NONE.

The HMI level should only be set if resumption is still needed.  This
occurs in the `StartAppHmiStateResumption` function. If the app is
activated, it is removed from the resumption queue in
`RemoveFromResumption`. This ensures that the HMI is not set during
resumption when the app is already active.

Therefore, there is no need to have `SetupDefaultHMILevel` in the
initial resume function since `SetupDefaultHMILevel` should only be
called when the app actually needs resumption.

### DIFF
--- a/src/components/application_manager/src/resumption/resume_ctrl.cc
+++ b/src/components/application_manager/src/resumption/resume_ctrl.cc
@@ -300,7 +300,6 @@ bool ResumeCtrl::StartResumption(ApplicationSharedPtr application,
                           << " hmi_app_id = " << application->hmi_app_id()
                           << " policy_id = " << application->policy_app_id()
                           << " received hash = " << hash);
-  SetupDefaultHMILevel(application);
   smart_objects::SmartObject saved_app;
   const std::string& device_mac = application->mac_address();
   bool result = resumption_storage_->GetSavedApplication(
@@ -326,7 +325,6 @@ bool ResumeCtrl::StartResumptionOnlyHMILevel(ApplicationSharedPtr application) {
                     << application->app_id() << "with hmi_app_id "
                     << application->hmi_app_id() << ", policy_app_id "
                     << application->policy_app_id());
-  SetupDefaultHMILevel(application);
   const std::string& device_mac = application->mac_address();
   smart_objects::SmartObject saved_app;
   bool result = resumption_storage_->GetSavedApplication(


AGaliuzov:
@BTPankow I believe that this pull request fixes behavior rather then root cause.
`SetupDefaultHMILevel` sets default HMI level which could differ from `NONE`. Actually it reads `HMILevel` from policy table.

As far as I understand the problem happened in case of `SystemRequest(LAUNCH_APP)` I think some additional investigation should be done.

Also please pay attention that CI fails with this pull request

BTPankow:
@AGaliuzov Thanks for the feedback.  I agree that there may need to be a different way to fix this issue.  The requirements for [ActivateApp](https://smartdevicelink.com/en/docs/hmi/master/basiccommunication/activateapp/) from sdl say that 
> In the case of a successful HMILevel resumption, SDL will first assign a 'default_hmi' level from policies with a 3 second timeout, before resuming the application to either FULL or LIMITED.

I'm not sure if that also should occur when a registerAppInterface request is sent.  I believe the tests interpreted it that way and that is why they are failing.  However, registerAppInterface already sends an onHMIStatus notification with default level when it is processed [here](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc#L595).  It would then send another notification for resumption [here](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc#L603) or [here](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc#L605).

Also, this issue occurs with more than just `SystemRequest(LAUNCH_APP)`.  I observed this problem occurring due to ActivateApp being sent from the HMI immediately after the app is registered.  It is not just a problem specifically with LAUNCH_APP, the problem just manifests itself there due to the HMI implementation.  Anytime the HMI is notified of app registration and immediately activates the app, this issue will occur.

Do you think that this is expected behavior for sdl core?  Should it still send the default hmi level if the app has already been activated?  If this is not the case, then we could add a check before sending the default hmi level for resumption.

The best solution could be to just use [ActivateApp](https://smartdevicelink.com/en/docs/hmi/master/sdl/activateapp/) with App Launching instead on the HMI side.

anosach-luxoft:
Reviewed

BTPankow:
I think @AGaliuzov is correct in that this fix may not address the issue directly, and it could violate requirements.  Closed in favor of pull request #840.

PR #830, opened 2016-09-20
Fix compile errors when ENABLE_LOG is 0

BTPankow:
After running cmake with flag -DENABLE_LOG=0, sdl_core fails to build with make.

This pull request fixes the issues that occur so that sdl_core can now be compiled with ENABLE_LOG=0.

### DIFF
--- a/src/components/application_manager/src/resumption/resume_ctrl.cc
+++ b/src/components/application_manager/src/resumption/resume_ctrl.cc
@@ -658,7 +658,6 @@ bool ResumeCtrl::CheckAppRestrictions(
   LOG4CXX_AUTO_TRACE(logger_);
   DCHECK_OR_RETURN(saved_app.keyExists(strings::hmi_level), false);
 
-  const bool is_media_app = application->is_media_application();
   const HMILevel::eType hmi_level =
       static_cast<HMILevel::eType>(saved_app[strings::hmi_level].asInt());
   const bool result = Compare<HMILevel::eType, EQ, ONE>(
@@ -666,8 +665,9 @@ bool ResumeCtrl::CheckAppRestrictions(
                           ? true
                           : false;
   LOG4CXX_DEBUG(logger_,
-                "is_media_app " << is_media_app << "; hmi_level " << hmi_level
-                                << " result " << result);
+                "is_media_app " << application->is_media_application()
+                                << "; hmi_level " << hmi_level << " result "
+                                << result);
   return result;
 }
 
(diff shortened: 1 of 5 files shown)


src/components/transport_manager/src/transport_adapter/transport_adapter_impl.cc
```diff
@@ -751,7 +751,9 @@ void TransportAdapterImpl::ConnectFailed(const DeviceUID& device_handle,
 void TransportAdapterImpl::RemoveFinalizedConnection(
     const DeviceUID& device_handle, const ApplicationHandle& app_handle) {
   const DeviceUID device_uid = device_handle;
+#if ENABLE_LOG
   const ApplicationHandle app_uid = app_handle;
+#endif
```
jacobkeeler: You can actually get around macroing this out by changing Line 760 from:
```cpp
connections_.find(std::make_pair(device_uid, app_handle));
```
to 
```cpp
connections_.find(std::make_pair(device_uid, app_uid));
```
Based on other functions in this file with similar variables, this is how the variable should be used

BTPankow: It looks like `connections_` type `ConnectionMap` is defined as `std::map<std::pair<DeviceUID, ApplicationHandle>, ConnectionInfo>`.  It also looks like all other uses of `connections_.find` with `make_pair` use `app_handle`.
It looks like I could change the following log methods to use `app_handle` in order to make the macro and definition of `app_uid` unnecessary.  What do you think?

anosach-luxoft:
Reviewed

src/components/application_manager/src/resumption/resume_ctrl.cc
```diff
@@ -658,7 +658,9 @@ bool ResumeCtrl::CheckAppRestrictions(
   LOG4CXX_AUTO_TRACE(logger_);
   DCHECK_OR_RETURN(saved_app.keyExists(strings::hmi_level), false);
 
+#ifdef ENABLE_LOG
   const bool is_media_app = application->is_media_application();
+#endif
```
jacobkeeler: You should be able to simply remove this local variable and change the log line so that it doesn't use the variable:
```cpp
  LOG4CXX_DEBUG(logger_,
                "is_media_app " << application->is_media_application()
                                << "; hmi_level " << hmi_level << " result "
                                << result);
```

BTPankow: Fixed in most recent commits.

src/components/transport_manager/src/transport_adapter/transport_adapter_impl.cc
```diff
@@ -751,21 +751,20 @@ void TransportAdapterImpl::ConnectFailed(const DeviceUID& device_handle,
 void TransportAdapterImpl::RemoveFinalizedConnection(
     const DeviceUID& device_handle, const ApplicationHandle& app_handle) {
   const DeviceUID device_uid = device_handle;
-  const ApplicationHandle app_uid = app_handle;
```
pestOO: @BTPankow 
This handle also using inside of `connections_.find` and this delete lead to build fail.

PR #837, opened 2016-09-22
Change Double type parameters in spec to Float type

### LINKED ISSUE #823: SendLocation uses Doubles instead of Floats
The lat and long params are of type Double when they should be floats in the MOBILE_API for the struct Coordinate and in the SendLocation RPC. The lat and long params of SendLocation are also not mandatory which is incorrect.

[See here for Coordinate](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/interfaces/MOBILE_API.xml#L2293)

[See here for SendLocation](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/interfaces/MOBILE_API.xml#L4792)

jacobkeeler:
PR Changes
  - Mobile API
    - longitudeDegrees and latitudeDegrees parameters in Coordinate and SendLocation changed to Float type from Double
  - HMI API
    - longitudeDegrees and latitudeDegrees parameters in Coordinate and GPSData changed to Float type from Double

Fixes #823

### DIFF
--- a/src/components/interfaces/HMI_API.xml
+++ b/src/components/interfaces/HMI_API.xml
@@ -243,10 +243,10 @@
   </enum>
 
   <struct name="Coordinate">
-    <param name="latitudeDegrees" minvalue="-90" maxvalue="90" type="Double" mandatory="true">
+    <param name="latitudeDegrees" minvalue="-90" maxvalue="90" type="Float" mandatory="true">
       <description>Latitude of the location.</description>
     </param>
-    <param name="longitudeDegrees" minvalue="-180" maxvalue="180" type="Double" mandatory="true">
+    <param name="longitudeDegrees" minvalue="-180" maxvalue="180" type="Float" mandatory="true">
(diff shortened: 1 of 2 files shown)


JackLivio:
@jacobkeeler Did the data type need to be changed for this parameter elsewhere in the project? Are they handled as floats/doubles at all?

jacobkeeler:
@JackLivio From what I can tell, the [interface generator](https://github.com/smartdevicelink/sdl_core/blob/master/tools/InterfaceGenerator/generator/parsers/RPCBase.py#L570-L584) treats Double and Float values exactly the same. It shouldn't affect any functionality as a result, this is just for consistency since all of the other floating point values in the API are "Float" rather than "Double".

jacobkeeler:
@mghiumiusliu Please review

JackLivio:
@LuxoftAKutsan @AGaliuzov Any problems you might see with the changes?

PR #840, opened 2016-09-23
Fix HMI NONE after activation

BTPankow:
This is a different fix for issue #805. The other fix is pull request #829, which may not conform to requirements.  This fix still allows default HMI to be immediately assigned on resumption.  This change allows resumption to occur immediately after the HMI is notified of an app registration.

If an HMI activates an application immediately after receving a
register app notification, then the mobile application could receive a
HMI FULL notification from the activation and then an HMI NONE
notification. The HMI NONE notification comes from the resumption
function following the HMI and mobile responses.

By sending the HMI notification last, we can ensure that resumption will
follow before the HMI sends an `ActivateApp`. With this commit, the
mobile application will first be notified of successful registration and
then the HMI will be notified.

### DIFF
--- a/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
+++ b/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
@@ -590,9 +590,9 @@ void RegisterAppInterfaceRequest::SendRegisterAppInterfaceResponseToMobile() {
                                             application->mac_address());
   }
 
+  SendResponse(true, result_code, add_info.c_str(), &response_params);
   SendOnAppRegisteredNotificationToHMI(
       *(application.get()), resumption, need_restore_vr);
-  SendResponse(true, result_code, add_info.c_str(), &response_params);
 
   // Check if application exists, because application might be unregestered
   // during sending reponse to mobile.


src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -590,9 +590,9 @@ void RegisterAppInterfaceRequest::SendRegisterAppInterfaceResponseToMobile() {
                                             application->mac_address());
   }
 
+  SendResponse(true, result_code, add_info.c_str(), &response_params);
   SendOnAppRegisteredNotificationToHMI(
       *(application.get()), resumption, need_restore_vr);
-  SendResponse(true, result_code, add_info.c_str(), &response_params);
```
anosach-luxoft: @BTPankow are you sure you'll not brake requirements with this changes?

madhuy:
@BTPankow this fix also working in our environment.

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
jacobkeeler: @anosach-luxoft After looking through the Confluence project, this fix does appear to break the sequence of messages which are sent after receiving a RegisterAppInterface request:

> The order of requests/responses/notifications during application registering must be the following:
1.app->SDL: RegisterAppInterface (policy_appID, parameters)
2.SDL->HMI: OnAppRegistered (hmi_appID, params)
3.SDL->app: RegisterAppInterface_response (<applicable resultCode>, success:true)
4. SDL->app: OnHMIStatus(hmiLevel, audioStreamingState, systemContext)
5. SDL->app: OnPermissionsChange(params)

Is there any specific reason why the HMI needs to be notified of an app registration prior to the mobile device?

madhuy: @jacobkeeler is there any side effect if we send registerinterface response first and notify HMI?

madhuy:
@BTPankow @jacobkeeler @anosach-luxoft Please provide me ETA on this issue

jacobkeeler:
@anosach-luxoft It doesn't seem to me like we can fix this issue without breaking this requirement to some degree (or delaying the processing of the ActivateApp request). 

The issue stems from the fact that an ActivateApp request can be sent between when the OnAppRegistered notification is sent to the HMI (where upon receiving the response, OnHMIStatus(FULL) is sent) and when the initial OnHMIStatus(NONE) is sent. 

Either the order of the messages needs to be changed (as in this PR), or Core needs to be prevented from processing the ActivateApp message before the initial OnHMIStatus(NONE) is sent, and I'm not certain if there is a clean way to do that.

jacobkeeler:
@mghiumiusliu Do you know whether changing the order of these messages could cause actually cause any other issues, because it seems to me like this is the only real clean solution to this issue?

madhuy:
@BTPankow @jacobkeeler @anosach-luxoft 
is it acceptable solution if HMI delays( 1 to 2 sec) sending ActivateApp ?

jacobkeeler:
@madhuy Performing a delay should theoretically fix the issue on your side, yes. 

I am not certain whether requiring that the HMI perform this delay will be the way that we finally fix the issue as of yet.

madhuy:
@jacobkeeler understand the concern.

madhuy:
@BTPankow @jacobkeeler @anosach-luxoft 
any ETA on this PR?

jacobkeeler:
@mghiumiusliu @LuxoftAKutsan @AGaliuzov Please review or assign someone to review ASAP.

jacobkeeler:
@madhuy In the meantime, I will smoke test this PR, as well as test with the ATF

AGaliuzov:
@jacobkeeler Accrording to the `travis-ci` there is the core crash in policy test within this PR.
So please pay attention.
@BTPankow have you check this crash?

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
anosach-luxoft: @jacobkeeler I don't know any reason to notify HMI first, so if you did not find strict requirement for that it could be changed

jacobkeeler:
@AGaliuzov I'm not certain what is causing that test failure, but it occurs on my local machine whenever I test ANY branch off of master since release 4.1 (including master itself), and seems to fail randomly when making hotfix PRs. You will see that the tests fail at the exact place in the Travis build for #830 ([here](https://travis-ci.org/smartdevicelink/sdl_core/builds/158736224)), but it did not break the Travis build on master [after being merged](https://travis-ci.org/smartdevicelink/sdl_core/builds/162799713).

As a result, it doesn't seem to me like the changes in this PR could have caused this test failure, it is the result to something existing already in master.

jacobkeeler:
@AGaliuzov After re-running the build for this PR, all of the tests passed, further showing that the policy_test failure is intermittent.

jacobkeeler:
@AGaliuzov Any other feedback?

PR #841, opened 2016-09-23
Ubuntu 16.04 (and other OS) Support

BTPankow:
This pr adds support for newer distros which will allow sdl_core to be compiled using recent gcc versions with no warnings. sdl_core can still be compiled on the standard development environment with these changes.

See the commit messages for more information on the changes made.

### DIFF
--- a/src/components/security_manager/src/crypto_manager_impl.cc
+++ b/src/components/security_manager/src/crypto_manager_impl.cc
@@ -130,8 +130,13 @@ bool CryptoManagerImpl::Init() {
 #endif
   switch (get_settings().security_manager_protocol_name()) {
     case SSLv3:
+#ifdef OPENSSL_NO_SSL3
+      SDL_WARN("OpenSSL does not support SSL3 protocol");
+      return false;
+#else
       method = is_server ? SSLv3_server_method() : SSLv3_client_method();
       break;
+#endif
     case TLSv1:
       method = is_server ? TLSv1_server_method() : TLSv1_client_method();
       break;
--- a/src/components/application_manager/include/application_manager/application_manager_impl.h
+++ b/src/components/application_manager/include/application_manager/application_manager_impl.h
@@ -1417,7 +1417,7 @@ class ApplicationManagerImpl
   // Thread that pumps messages audio pass thru to mobile.
   impl::AudioPassThruQueue audio_pass_thru_messages_;
 
-  std::auto_ptr<HMICapabilities> hmi_capabilities_;
+  std::unique_ptr<HMICapabilities> hmi_capabilities_;
   // The reason of HU shutdown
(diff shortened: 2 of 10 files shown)


pestOO:
@BTPankow 
We are not using C++11 feature as ` std::unique_ptr` due to support [C++98](https://smartdevicelink.com/en/guides/core/software-architecture-document/development-view/#481-implementation-technologies).
See also list of officially support platforms: https://smartdevicelink.com/en/guides/core/software-architecture-document/development-view/#483-development-environment-and-standards

@LuxoftAKutsan, @AGaliuzov, please support with a build for Ubuntu 16.04

pestOO:
@LuxoftAKutsan, @abyz

JackLivio:
@pestOO When trying to compile with the flag -std=c++98 the compiler throws this error

```
/usr/include/c++/4.8/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library
support for the ISO C++ 2011 standard. This support is currently experimental, and must be enabled 
with the -std=c++11 or -std=gnu++11 compiler options.

```

along with many other errors found in profile.cc

jacobkeeler:
@pestOO the `cstdint` header was added is C++11, and it is included/used in several places throughout sdl_core (such as [protocol_handler_settings.h](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/include/protocol_handler/protocol_handler_settings.h)), meaning that the currently released version needs to be built using the C++11 standard and uses features from this standard.

LuxoftAKutsan:
@JackLivio @jacobkeeler That right, currently we have some issues, so we can't compile with c++98. But we are removing this issues continuously and avoid adding new ones.

src/components/security_manager/src/crypto_manager_impl.cc
```diff
@@ -129,9 +129,15 @@ bool CryptoManagerImpl::Init() {
   const SSL_METHOD* method;
 #endif
   switch (get_settings().security_manager_protocol_name()) {
+#ifdef OPENSSL_NO_SSL3
```
LuxoftAKutsan: @BTPankow please avoid duplication  ```case SSLv3:``` and put ifdef inside case

jacobkeeler: @LuxoftAKutsan The commit has been amended with this change.

I will be taking over all of BTPankow's PR's in Core from now on, so you can ping me with any more feedback you have

jacobkeeler:
So at this point, is the reversal of the change from `auto_ptr` to `unique_ptr` all that needs to be changed? Are there any other issues with the deprecation fixes that were made in this PR?

LuxoftAKutsan:
@jacobkeeler I don't see any restricted issues in this pull request except uniq_ptr.

jacobkeeler:
Closing in favor of #924

PR #851, opened 2016-09-27
Fix TestObject classes in the same namespace

pvvasilev:
There are two TestObject classes in the same namespace.
This can cause problems at linking.
This PR will rename them.

Related issue: [APPLINK-28224](https://adc.luxoft.com/jira/browse/APPLINK-28224)


pvvasilev:
Please review @dtrunov @Kozoriz @AGritsevich @dev-gh @LuxoftAKutsan @anosach-luxoft

pvvasilev:
@Anatoliy-Leshin @dtrunov Progress suspended because of issues with tests. This is the ticket to fix them https://adc.luxoft.com/jira/browse/APPLINK-28235

PR #855, opened 2016-09-28
Fix SystemRequestTest fails on repeatedly run

LevchenkoS:
Fixed `SystemRequestTest` failure on repeatedly run

`SystemRequestTest` fails because `SystemRequest
every new call, change save file name.
In test provided only one name for save file.
As an result, in repeat mode, `SystemRequest`
change file name, and it mismatch name provided in test.

-
Closes bug: [APPLINK-28272](https://adc.luxoft.com/jira/browse/APPLINK-28272)

@AGaliuzov, @Kozoriz, @LuxoftAKutsan, @VProdanov, @wolfylambova22, please, review.

### DIFF
--- a/src/components/application_manager/include/application_manager/commands/mobile/system_request.h
+++ b/src/components/application_manager/include/application_manager/commands/mobile/system_request.h
@@ -80,6 +80,18 @@ class SystemRequest : public CommandRequestImpl {
    */
   virtual void on_event(const event_engine::Event& event) OVERRIDE;
 
+#ifdef BUILD_TESTS
+  /**
+   * @brief To avoid override of existing file,
+   * command append index as a suffix to
+   * the file name and increment it, each time
+   * when command saves a file.
+   **/
+  static uint32_t file_index() {
+    return index;
+  }
+#endif  // BUILD_TESTS
+
  private:
   /**
    * @brief Validates data coming within QueryApps response
--- a/src/components/application_manager/test/commands/mobile/system_request_test.cc
+++ b/src/components/application_manager/test/commands/mobile/system_request_test.cc
@@ -31,6 +31,8 @@
  */
 
 #include <string>
+#include <sstream>
+
 #include "gtest/gtest.h"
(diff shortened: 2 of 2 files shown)


src/components/application_manager/test/commands/mobile/system_request_test.cc
```diff
@@ -251,10 +250,27 @@ TEST_F(
...
+    snprintf(buf,
+             max_size - 1,
+             "%d%s",
+             SystemRequest::file_index(),
+             file_name.c_str());
+    file_name = buf;
+  }
+
+  const std::string app_full_file_path =
+      kStorageFolder + file_system::GetPathDelimiter() + "app_storage" +
```
Kozoriz: @LevchenkoS , you can use path_concat

LevchenkoS: @Kozoriz, I think `app_full_file_path` more clear for understanding

src/components/application_manager/include/application_manager/commands/mobile/system_request.h
```diff
@@ -80,6 +80,12 @@ class SystemRequest : public CommandRequestImpl {
    */
   virtual void on_event(const event_engine::Event& event) OVERRIDE;
 
+#ifdef BUILD_TESTS
+  static uint32_t file_index() const {
+    return index;
```
LuxoftAKutsan: @LevchenkoS what is index? Doxygen missed

src/components/application_manager/test/commands/mobile/system_request_test.cc
LevchenkoS: @Kozoriz, reworked in fce2b89

src/components/application_manager/include/application_manager/commands/mobile/system_request.h
LevchenkoS: @LuxoftAKutsan, fixed in 6aec3c3

PR #861, opened 2016-09-30
Fix defects found by am_commands_test under WinQt

vlantonov:
Fixed memory leaks.
Fixed Windows path creation.
Fixed Unit test output.
Fixed DataAccessor expectation release.
Refactored AddCommandRequestTest.

Related issue: [APPLINK-27667](https://adc.luxoft.com/jira/browse/APPLINK-27667)

### DIFF
--- a/src/components/application_manager/src/commands/mobile/set_global_properties_request.cc
+++ b/src/components/application_manager/src/commands/mobile/set_global_properties_request.cc
@@ -265,7 +265,7 @@ void SetGlobalPropertiesRequest::on_event(const event_engine::Event& event) {
                 (is_ui_invalid_unsupported && is_tts_succeeded);
 
   mobile_apis::Result::eType result_code = mobile_apis::Result::INVALID_ENUM;
-  const char* return_info = NULL;
+  std::string return_info;
 
   const bool is_ui_or_tts_warning =
       Compare<hmi_apis::Common_Result::eType, EQ, ONE>(
@@ -274,8 +274,7 @@ void SetGlobalPropertiesRequest::on_event(const event_engine::Event& event) {
   if (result && (hmi_apis::Common_Result::UNSUPPORTED_RESOURCE == tts_result_ ||
                  is_ui_or_tts_warning)) {
     result_code = mobile_apis::Result::WARNINGS;
-    return_info =
-        std::string("Unsupported phoneme type sent in a prompt").c_str();
+    return_info = "Unsupported phoneme type sent in a prompt";
   } else {
     result_code =
         MessageHelper::HMIToMobileResult(std::max(ui_result_, tts_result_));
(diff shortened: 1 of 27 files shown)


vlantonov:
Please review: @AByzhynar, @dcherniev, @OHerasym, @VProdanov, @wolfylambova22, @pvvasilev

src/components/application_manager/src/commands/mobile/set_global_properties_request.cc
```diff
@@ -285,8 +284,10 @@ void SetGlobalPropertiesRequest::on_event(const event_engine::Event& event) {
   ApplicationSharedPtr application =
       application_manager_.application(connection_key());
 
-  SendResponse(
-      result, result_code, return_info, &(message[strings::msg_params]));
+  SendResponse(result,
+               result_code,
+               return_info.c_str(),
```
OHerasym: @vlantonov, what will return c_str() in case empty string?? As I remember it returns "" instead of NULL. In this case command will send info="". It's a defect.

src/components/application_manager/src/commands/mobile/speak_request.cc
```diff
@@ -137,13 +137,14 @@ void SpeakRequest::ProcessTTSSpeakResponse(
     return_info = "Unsupported phoneme type sent in a prompt";
   }
 
-  SendResponse(
-      result, result_code, return_info, &(message[strings::msg_params]));
+  SendResponse(result,
+               result_code,
+               return_info.c_str(),
```
OHerasym: @vlantonov , same issue

src/components/application_manager/src/commands/mobile/set_global_properties_request.cc
vlantonov: @OHerasym "" actually is a '\0' a.k.a. `NULL`

vlantonov: @OHerasym Fixed, see 6042aedc0fa09ac42a040a7d0df2024988463d57

src/components/application_manager/src/commands/mobile/speak_request.cc
vlantonov: @OHerasym Fixed, see 6042aedc0fa09ac42a040a7d0df2024988463d57

src/components/application_manager/test/commands/hmi/hmi_notifications/hmi_notifications_test.cc
```diff
@@ -239,6 +239,13 @@ class HMICommandsNotificationsTest
   void SetUp() OVERRIDE {
     message_helper_mock_ =
         application_manager::MockMessageHelper::message_helper_mock();
+    Mock::VerifyAndClearExpectations(message_helper_mock_);
+  }
+
+  void TearDown() OVERRIDE {
+    // Fix DataAccessor release and WinQt crash
+    Mock::VerifyAndClearExpectations(&mock_app_manager_);
+    Mock::VerifyAndClearExpectations(message_helper_mock_);
```
pvvasilev: @vlantonov Why are these two calls in the `TearDown()` here and in destructors in other files? Isn't it possible that they are put in the destructor here as well?

vlantonov: @pvvasilev In this specific case the `SetUp()` was already existing, so I decided to use it while adding `TearDown()`. In other cases it was more convenient just to add already defined `SetUp()/TearDown()` than to set specific constructors/destructors.

vlantonov: @pvvasilev Fixed, see e9de67a266fec5a338e51b387d947cff27f8e383

PR #870, opened 2016-10-05
Remove all incorrect usages of std::string

LevchenkoS:
Currently, in code present incorrect usage
 of std::string like this:
`std::string("Some string").c_str();`
It causes undefined behavior, and should
be deprecated.

-
Related to: [APPLINK-28458](https://adc.luxoft.com/jira/browse/APPLINK-28458)

@AGaliuzov, @Kozoriz, @LuxoftAKutsan, @VProdanov, @wolfylambova22, please, review.

### DIFF
--- a/src/components/application_manager/src/commands/mobile/reset_global_properties_request.cc
+++ b/src/components/application_manager/src/commands/mobile/reset_global_properties_request.cc
@@ -275,13 +275,12 @@ void ResetGlobalPropertiesRequest::on_event(const event_engine::Event& event) {
          (hmi_apis::Common_Result::SUCCESS == tts_result_));
 
     mobile_apis::Result::eType result_code;
-    const char* return_info = NULL;
+    std::string return_info;
 
     if (result) {
       if (hmi_apis::Common_Result::UNSUPPORTED_RESOURCE == tts_result_) {
         result_code = mobile_apis::Result::WARNINGS;
-        return_info =
-            std::string("Unsupported phoneme type sent in a prompt").c_str();
+        return_info = "Unsupported phoneme type sent in a prompt";
       } else {
         result_code = static_cast<mobile_apis::Result::eType>(
             std::max(ui_result_, tts_result_));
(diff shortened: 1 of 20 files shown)


src/components/media_manager/src/audio/from_mic_to_file_recorder_thread.cc
```diff
@@ -120,14 +120,16 @@ media_manager::FromMicToFileRecorderThread::Impl::Impl(
 
   std::stringstream stringStream;
   stringStream << duration / 1000;
-  std::string durationString = stringStream.str();
+  const std::string durationString = stringStream.str();
```
Kozoriz: @LevchenkoS is there is correct naming?

LevchenkoS: @Kozoriz, fixed in c5dcb82.

src/components/utils/src/file_system_posix.cc
```diff
@@ -112,8 +112,9 @@ bool file_system::CreateDirectoryRecursively(const std::string& utf8_path) {
 
   while (ret_val == true && pos <= utf8_path.length()) {
     pos = utf8_path.find(GetPathDelimiter(), pos + 1);
+    const std::string path = utf8_path.substr(0, pos);
     if (!DirectoryExists(utf8_path.substr(0, pos))) {
-      if (0 != mkdir(utf8_path.substr(0, pos).c_str(), S_IRWXU)) {
+      if (0 != mkdir(path.c_str(), S_IRWXU)) {
```
wolfylambova22: @LevchenkoS, so c.str() should remain here?

src/components/application_manager/src/commands/mobile/reset_global_properties_request.cc
```diff
@@ -293,7 +292,7 @@ void ResetGlobalPropertiesRequest::on_event(const event_engine::Event& event) {
 
     SendResponse(result,
                  static_cast<mobile_apis::Result::eType>(result_code),
-                 return_info,
+                 return_info.c_str(),
```
VProdanov: @LevchenkoS  return_info.empty() ? NULL : return_info.c_str()..

PR #804, opened 2016-09-07
Fix failed utils tests on WIN & QT platform

OHerasym:
### PR Changes:
- Fix failed utils tests on WIN & QT platform

Related: [APPLINK-27145](https://adc.luxoft.com/jira/browse/APPLINK-27145), [27156](https://adc.luxoft.com/jira/browse/APPLINK-27156)

PR #871, opened 2016-10-05
Fix failed utils tests on WIN & QT platform

OHerasym:
### PR Changes:
- Fix failed utils tests on WIN & QT platform

Related: APPLINK-27145, 27156

[Old pull request](https://github.com/smartdevicelink/sdl_core/pull/804)


OHerasym:
Please review: @vlantonov, @AGritsevich, @anosach-luxoft , @VProdanov, @orhan-mehmedov

src/components/utils/src/convert_utils.cc
```diff
@@ -70,3 +70,9 @@ uint64_t utils::ConvertLongLongUIntToUInt64(
 
   return static_cast<uint64_t>(value);
 }
+
+int64_t utils::Int64DivisionRoundUp(const int64_t numerator,
+                                    const uint64_t denominator) {
+  return numerator / denominator +
+         (((numerator < 0) ^ (denominator > 0)) && (numerator % denominator));
```
vlantonov: @OHerasym `denominator` is always positive, so the check could be excluded.

src/components/utils/test/message_loop_thread_test.cc
```diff
@@ -1,52 +0,0 @@
...
-
-  ASSERT_EQ(0u, message_loop_thread.GetMessageQueueSize());
-
-  message_loop_thread.PostMessage(true);
-  ASSERT_EQ(1u, message_loop_thread.GetMessageQueueSize());
-}
-
-}  // namespace utils_test
-}  // namespace components
-}  // namespace test
```
vlantonov: @OHerasym Why this test is removed?

src/components/utils/src/convert_utils.cc
```diff
@@ -70,3 +70,9 @@ uint64_t utils::ConvertLongLongUIntToUInt64(
 
   return static_cast<uint64_t>(value);
 }
+
+int64_t utils::Int64DivisionRoundUp(const int64_t numerator,
+                                    const uint64_t denominator) {
```
orhan-mehmedov: @OHerasym mixing arithmetic of signed and unsigned in general is not good idea in C/C++. Small negative numerators will be converted to big unsigned positive numbers. Try numerator -100 and denominator 20. What do you expect? -5? Probably it will be 922337203685477575.
I suggest to store the sign and then do the calculations by absolute values in unsigned format then restore the sign in the final result. OR use uint32_t for unsigned (keep signed bigger).

src/components/utils/test/date_time_test.cc
```diff
@@ -108,7 +110,7 @@ TEST(DateTimeTest, CalculateTimeSpan) {
   // arrange
   const TimevalStruct time = date_time::DateTime::getCurrentTime();
 
-  const uint32_t sleep_time_mSec = 10;
+  const int64_t sleep_time_mSec = 100;
```
orhan-mehmedov: @OHerasym Why it was increased from 10 to 100?

LuxoftAKutsan:
@OHerasym  please resend this pull request to pre_5_0_0_modifications

PR #887, opened 2016-10-10
Add distcc and ccache as cmake options

LuxoftAKutsan:


LuxoftAKutsan:
@Abyzhynar, @agritsevich, @okozlovlux, @dev-gh, @dtrunov, @orhan-mehmedov Please review

LuxoftAKutsan:
@AGaliuzov also take a look

CMakeLists.txt
```diff
@@ -66,6 +66,8 @@ option(ENABLE_HMI_PTU_DECRYPTION "Policy table update parsed by hmi" ON)
 option(ENABLE_EXTENDED_POLICY "Turns extended flow which requires embedded system interaction" ON)
 option(ENABLE_THREAD_COUNT "Turn on correct threads shutdown checks" OFF)
 option(FAST_BUILD "Turn on fast build" ON)
+option(USE_CCACHE "Turn on ccache usage" ON)
+option(USE_DISTCC "Turn on distributed build_usage" ON)
```
AByzhynar: @LuxoftAKutsan I think this option should be turned off by default.

dev-gh:
@LuxoftAKutsan 
Maybe it worth to add distcc hosts file to repo?

CMakeLists.txt
pestOO: @LuxoftAKutsan @AByzhynar 
For Luxoft developers it have sense to enable that, but due to opensource community it shall be disables.
At least for master/releases it shall be OFF.

pestOO: Also, will it affect CI builds?

AByzhynar: @pestOO CI servers - probably no. But code coverage is calculated incorrectly if distributed build enabled.

orhan-mehmedov: @pestOO It will produce just warning message in case ccache/distcc is not found and will take the old way. I do not see any big impact.

LuxoftAKutsan: Note that enabling this flag switch automatic distributed build id developer use multijobs make.
In case if developer use one job make - no distribution will be enabled.

LuxoftAKutsan:
@dev-gh list of our local ip addresses to opensource repo rely?

PR #895, opened 2016-10-12
Fix problem with incorrect success result

dtrunov:
HMI send erronius result code and after that SDL answers with success=true. I fix this problem.

Closes bug [APPLINK-28762](https://adc.luxoft.com/jira/browse/APPLINK-28762)


dtrunov:
@vlantonov @OHerasym @LuxoftAKutsan @dev-gh @VProdanov Please review ASAP

PR #897, opened 2016-10-12
Fix/rui rejected vr unsupported resource delete command

LuxoftAKutsan:
Related issue : https://adc.luxoft.com/jira/browse/APPLINK-28797


LuxoftAKutsan:
@Abyzhynar, @okozlovlux, @dev-gh, @dtrunov, @orhan-mehmedov
Please review as soon as possible

PR #898, opened 2016-10-12
Fix/sdl responds add command with incorrect info

dtrunov:
Fix problem with merge info
[APPLINK-28758](https://adc.luxoft.com/jira/browse/APPLINK-28758)


dtrunov:
Please review @AGritsevich @OHerasym

Issue #890, opened 2016-10-11
Build fails if -DEXTENDED_MEDIA_MODE=ON


yang1070:
[ 95%] Building CXX object src/components/media_manager/CMakeFiles/MediaManager.dir/src/media_manager_impl.cc.o
In file included from /home/zhimin/github/sdl_core/src/components/media_manager/src/media_manager_impl.cc:45:0:
/home/zhimin/github/sdl_core/src/components/media_manager/include/media_manager/audio/a2dp_source_player_adapter.h: In member function ‘void media_manager::A2DPSourcePlayerAdapter::set_connection_handler(connection_handler::ConnectionHandlerImpl*)’:
/home/zhimin/github/sdl_core/src/components/media_manager/include/media_manager/audio/a2dp_source_player_adapter.h:63:5: error: ‘connection_handler_’ was not declared in this scope
     connection_handler_ = connection_handler;
     ^
src/components/media_manager/CMakeFiles/MediaManager.dir/build.make:158: recipe for target 'src/components/media_manager/CMakeFiles/MediaManager.dir/src/media_manager_impl.cc.o' failed
make[2]: *** [src/components/media_manager/CMakeFiles/MediaManager.dir/src/media_manager_impl.cc.o] Error 1
CMakeFiles/Makefile2:2429: recipe for target 'src/components/media_manager/CMakeFiles/MediaManager.dir/all' failed
make[1]: *** [src/components/media_manager/CMakeFiles/MediaManager.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2

this happens in release/4.2.0, and mater with tag 4.1.0, 4.1.1. Didn't try other tags.

jacobkeeler:
@yang1070 Working on this, I believe I have fixed all of the issues with the build and will likely be submitting a PR for it later this week

yang1070:
@jacobkeeler  How do you set up the GStreamer? I mean the include path and link libs. I changed the code locally to remove the connection_handler error. Then I met a "utils/logger.h" file cannot find error. I think somehow the FindGstreamer-1.0.cmake file in the project should be called to setup the Macro, but I didn't figure out how.

jacobkeeler:
@yang1070 The connection_handler issue should have been fixed with the merge of #880.

I went through all of the build issues and fixed them in #900, so the changes in that PR should fix your issues. Note that you will need to manually specify the gstreamer directory as well:

```
[...]

PR #900, opened 2016-10-12
Fix EXTENDED_MEDIA_MODE build and runtime failures

jacobkeeler:
The build for SDL Core currently fails if EXTENDED_MEDIA_MODE=ON, this is partially due to a mistake made in a large CMake refactoring commit which removes the GStreamer include directories ([here](https://github.com/smartdevicelink/sdl_core/commit/9dba2aecd2a4c7b4054067deb413311c44c80cac#diff-d4aa18259c54c5324255aa1f61777fb3L66)). 

In addition, the GStreamer config directory is not included in the build, which is used by some of the headers in the gst directory which are included in the Media Manager. Unfortunately, FindGStreamer-1.0.cmake cannot find the config directory automatically, so the GSTREAMER_DIR environment variable still needs to be set manually for these changes to work (for Ubuntu 16.04 64-bit, this path should be /usr/lib/x86_64-linux-gnu/gstreamer-1.0):

```
export GSTREAMER_DIR=$PATH_TO_GSTREAMER
```

Finally, once Core is built with this feature, a SEGFAULT occurs in media_manager_impl.cc due to the code attempting to access a member of protocol_handler_ before it is set.

PR Changes
  * Edits Media Manager and appMain CMakeLists to fix build errors with Extended Media Mode
  * Fixes SEGFAULT error which occurs when the Media Manager is being initialized with Extended Media Mode

### DIFF
--- a/src/appMain/CMakeLists.txt
+++ b/src/appMain/CMakeLists.txt
@@ -45,6 +45,7 @@ cmake_policy(POP)
 if (EXTENDED_MEDIA_MODE)
 set(default_media_inc
 ${GSTREAMER_gst_INCLUDE_DIR}
+${GSTREAMER_gstconfig_INCLUDE_DIR}
 )
 else(EXTENDED_MEDIA_MODE)
 set(default_media_inc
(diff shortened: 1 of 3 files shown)


jacobkeeler:
Should fix #890

JackLivio:
@jacobkeeler Reviewed

PR #907, opened 2016-10-14
Release 4.2.0

LuxoftAKutsan:


JackLivio:
@dtrunov I noticed that there is no "onTimeout()" function implemented for the navigation is ready request. 

@LuxoftAKutsan @mghiumiusliu Also why is the behavior to send messages to HMI components when a timeout for an is ready request occurs? If an HMI component doesn't respond to the is ready request wouldn't it be assumed that the component is unavailable?

dtrunov:
@JackLivio: "I noticed that there is no "onTimeout()" function implemented for the navigation is ready request". 
It is correct because SDL shouldn't send message If HMI doesn't respond to navigation is ready request during timeout.

LuxoftAKutsan:
@JackLivio it is correct behaviour according to requirements to avoid missing resources in case if IsReady stubs on HMI. 
 
I case if HMI **didn't respond** to INTERFACE.IsReady() request SDL **still should send** INTERFACE specific requests and wait responses, in case time out for them send GENERIC_ERROR on HMI. 

I case if HMI **respond** to INTERFACE.IsReady() request with **available=true** SDL  should send INTERFACE specific requests and wait responses, in case time out for them send GENERIC_ERROR on HMI. 

I case if HMI **respond** to INTERFACE.IsReady() request  with **available=false** SDL **should not send** INTERFACE specific requests and respond UNSUPPORTED_RESOURCE to mobile. 

@AnastasiyaBritanova please approve or amend.

AnastasiyaBritanova:
@LuxoftAKutsan , what you described above is correct.

PR #908, opened 2016-10-18
Set policy db path to storage folder

### LINKED ISSUE #905: SDL core 4.1 - policy.sqlite file crated in current directory instead of 'AppStorageFolder'
in 4.0, it was creating in 'AppStorageFolder' path
in 4.1 it is creating in current directory 

Looked at the code,
can we add below code [here](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/policy/src/sql_pt_representation.cc#L322)

```
#ifndef __QNX__
  std::string path = get_settings().app_storage_folder();
  if (!path.empty()) {
    db_->set_path(path + "/");
  }
#endif // __QNX__
```

jacobkeeler:
Quick change which sets the path for `policy.sqlite` to the app storage folder defined in smartDeviceLink.ini

Fixes #905

### DIFF
--- a/src/components/policy/test/sql_pt_representation_test.cc
+++ b/src/components/policy/test/sql_pt_representation_test.cc
@@ -76,21 +76,20 @@ class SQLPTRepresentationTest : public SQLPTRepresentation,
   static DBMS* dbms;
   static SQLPTRepresentation* reps;
   static const std::string kDatabaseName;
+  static const std::string kAppStorageFolder;
   // Gtest can show message that this object doesn't destroyed
   static std::auto_ptr<policy_handler_test::MockPolicySettings>
       policy_settings_;
 
   static void SetUpTestCase() {
-    const std::string kAppStorageFolder = "storage1";
-    file_system::RemoveDirectory(kAppStorageFolder);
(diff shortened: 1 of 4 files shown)


jacobkeeler:
@LuxoftAKutsan @AGaliuzov Please review

src/components/policy/test/sql_pt_representation_test.cc
```diff
@@ -996,6 +997,7 @@ const std::string kAppStorageFolder = "storage";
 }
 
 TEST(SQLPTRepresentationTest3, Init_InitNewDataBase_ExpectResultSuccess) {
+  file_system::CreateDirectory(kAppStorageFolder);
```
AGaliuzov: @jacobkeeler in case of exception the directory will never be removed.
I propose to use one of the following approaches
1. Either wrpat `Create/RemoveDirectory`  with some RAII class
2. Or use test fixture and perform creationd/deletion in the `setup/teardown` method

jacobkeeler: Fixed in 8e508db

PR #920, opened 2016-10-27
setup_env.sh added

ovvasyliev:
setup_env.sh and support scripts added
Please review
@LuxoftAKutsan 
@AByzhynar 
@antongovorov

### DIFF
--- a/setup_env.sh
+++ b/setup_env.sh
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
 if [[ $EUID -ne 0 ]]; then
    echo "This script must be run as root" 1>&2
    exit 1
@@ -37,6 +39,8 @@ chromium-browser
 libgl1-mesa-dev
 libgtest-dev
 binutils
+cmake-curses-gui
+gitk
 )
 
 if [ $DISTRIB_RELEASE!="14.04" ]; then
@@ -48,10 +52,10 @@ apt-get install -y ${packages[@]}
 
 #if cmake not installed it will be installed. If any version of cmake installed except 2.8.12 it will be removed before installation.
 if ! dpkg -s cmake | grep "Version" | grep 2.8.12; then
-	apt-get remove cmake cmake-data
+	apt-get remove -y cmake cmake-data cmake-curses-gui
 	echo "INSTALLING CMAKE"
 	CMAKE_CORRECT_VERSION="$(apt-cache show cmake | grep Version | grep '2.8.12' | awk '{print $2}')"
-	apt-get install -y --force-yes cmake=$CMAKE_CORRECT_VERSION cmake-data=$CMAKE_CORRECT_VERSION
+	apt-get install -y --force-yes cmake=$CMAKE_CORRECT_VERSION cmake-data=$CMAKE_CORRECT_VERSION cmake-curses-gui=$CMAKE_CORRECT_VERSION
 fi
 
 . ./tools/Utils/update_gcc.sh


setup_env.sh
```diff
@@ -0,0 +1,58 @@
+if [[ $EUID -ne 0 ]]; then
+   echo "This script must be run as root" 1>&2
+   exit 1
+fi
+
+cp /etc/apt/sources.list /etc/apt/sources.list.backup
+
+. /etc/lsb-release
+echo $DISTRIB_RELEASE
```
LuxoftAKutsan: @ovvasyliev I think this is redundant output

setup_env.sh
```diff
@@ -0,0 +1,58 @@
...
+
+if [ $DISTRIB_RELEASE!="14.04" ]; then
+   sed -i -e '$a\deb http://ppa.launchpad.net/smspillaz/cmake-2.8.12/ubuntu wily main' /etc/apt/sources.list
+fi
+
+
+apt-get update
+apt-get install -y ${packages[@]}
+
+if ! dpkg -s cmake | grep "Version" | grep 2.8.12; then
```
LuxoftAKutsan: @ovvasyliev comment would be useful

setup_env.sh
```diff
@@ -0,0 +1,58 @@
...
+	apt-get remove cmake
```
LuxoftAKutsan: @ovvasyliev probably you should remove cmake and cmake-data

setup_env.sh
```diff
@@ -0,0 +1,58 @@
...
+	echo "INSTALLING CMAKE"
+	CMAKE_CORRECT_VERSION="$(apt-cache show cmake | grep Version | grep '2.8.12' | awk '{print $2}')"
+	apt-get install -y --force-yes cmake=$CMAKE_CORRECT_VERSION cmake-data=$CMAKE_CORRECT_VERSION
+fi
+
```
LuxoftAKutsan: @ovvasyliev 2 empty lines it to much. 1 is enough

setup_env.sh
```diff
@@ -0,0 +1,58 @@
...
+
+. ./tools/Utils/update_gcc.sh
+. ./tools/Utils/install_distcc.sh
+
+python ./tools/infrastructure/install_hooks.py
+
+cp /etc/apt/sources.list.backup /etc/apt/sources.list
+rm -f /etc/apt/sources.list.backup
```
LuxoftAKutsan: @ovvasyliev I would not remove backup file. It can be useful in case of errors during adding ppa

LuxoftAKutsan:
Also please add setting up git credentials 
Would be nice if it would be possible to run scripts as
```
setup_env.sh -c Alexander Kutsan akutsan@luxoft.com 
```
And script would setting up git --email ad git --name

LuxoftAKutsan:
@dpetrov-luxoft please also review

setup_env.sh
```diff
@@ -34,25 +33,37 @@ libgtest-dev
 )
 
 if [ $DISTRIB_RELEASE!="14.04" ]; then
+   #add-apt-repository -y ppa:smspillaz/cmake-2.8.12
```
LuxoftAKutsan: @ovvasyliev please remove commented code.Why you prefer use sed instead of ppa add?

setup_env.sh
```diff
@@ -34,25 +33,37 @@ libgtest-dev
...
 	apt-get install -y --force-yes cmake=$CMAKE_CORRECT_VERSION cmake-data=$CMAKE_CORRECT_VERSION
 fi
 
-
 . ./tools/Utils/update_gcc.sh
 . ./tools/Utils/install_distcc.sh
 
 python ./tools/infrastructure/install_hooks.py
 
+if [ "$1" == "-h" ]; then
```
LuxoftAKutsan: @ovvasyliev Would be nice to move help on the top of the script

LuxoftAKutsan:
@ovvasyliev please also get changes from this review(setup_env part) https://github.com/CustomSDL/sdl_panasonic/pull/632/

setup_env.sh
ovvasyliev: @LuxoftAKutsan there was a problem with ppa adding on Ubunutu 16

setup_env.sh
```diff
@@ -0,0 +1,70 @@
+if [[ $EUID -ne 0 ]]; then
```
AByzhynar: @ovvasyliev Where is sha-bang line?

setup_env.sh
```diff
@@ -0,0 +1,70 @@
...
+libbluetooth3
+libbluetooth-dev
+bluez-tools
+chromium-browser
+libgl1-mesa-dev
+libgtest-dev
+binutils
+)
+
+if [ $DISTRIB_RELEASE!="14.04" ]; then
```
AByzhynar: @ovvasyliev What if someone has 16.04 - we will add ppa for `wily`?

setup_env.sh
```diff
@@ -0,0 +1,70 @@
...
+doxygen
+graphviz
+mscgen
+libbluetooth3
+libbluetooth-dev
+bluez-tools
+chromium-browser
+libgl1-mesa-dev
+libgtest-dev
+binutils
```
AByzhynar: @ovvasyliev Can be appended by ccmake & gitk

setup_env.sh
LuxoftAKutsan: @ovvasyliev also please remove redundant line if you will change file

setup_env.sh
ovvasyliev: @AByzhynar line added 456972eb26cba4c8f2d0a34667034e9d83a9761c

setup_env.sh
ovvasyliev: @AByzhynar Ubuntu 15 and up does not have cmake 2.8 in standard packages. This instalation way was tested with Ubuntu 15.10 and 16.04.

setup_env.sh
```diff
@@ -0,0 +1,74 @@
...
+if [ $DISTRIB_RELEASE!="14.04" ]; then
+   sed -i -e '$a\deb http://ppa.launchpad.net/smspillaz/cmake-2.8.12/ubuntu wily main' /etc/apt/sources.list
+fi
+
+apt-get update
+apt-get install -y ${packages[@]}
+
+#if cmake not installed it will be installed. If any version of cmake installed except 2.8.12 it will be removed before installation.
+if ! dpkg -s cmake | grep "Version" | grep 2.8.12; then
+	apt-get remove -y cmake cmake-data cmake-curses-gui
```
JackLivio: @AByzhynar Is there any reason that this is forcing version 2.8.12? Since this is an older version of cmake, I don't think developers would be very happy to see a setup script removing their newer version of cmake and installing an older version. 

Can the check be to see if the version of cmake installed is older than 2.8.12, if so, then remove cmake and install.

AByzhynar: @JackLivio We are processing development and testing on fixed environment to minimize possible problems due to different environments. And `cmake 2.8.12` is fixed in our environment.

LuxoftAKutsan:
@ovvasyliev why this PR is this this not merged?

ovvasyliev:
@LuxoftAKutsan please merge

setup_env.sh
JackLivio: @AByzhynar 

@jacobkeeler and I believe that it should be documented in the README.md what the expected development environment is for setting up SDL, not forcing it on developers through a setup script.

Also this script doesn't give a notice to the developer that changes are being made to their installed packages, or even ask for permission to make these changes, which is not ok.

I also see that the script removes the currently installed versions gcc and g++ and installs it with the version specified for the SDL environment, which is even more not ok!

Any developer who wants to setup SDL Core and runs this script will mess up their development environment, and they'll have no idea why.

The setup_env script should be installing missing packages, not removing existing ones.

LuxoftAKutsan: @JackLivio I agree that environment in README should be specified version of cmake that SDL required. 

According removing existing packages and installed new versions.
I agree that  it is a bit confusing. But SDL requires specific versions of packages to be installed, and in case if installed versions differs, SDL can fail compiling or even works not correctly. So to create appropriate environment for building and running SDL we need to install appropriate version of dependencies.

I can suggest notify user that setup_env.sh will remove existing version of some packages and install new one. And ask user confirmation according each package (same behavior as yaourt package manager in Archlinux). 
Also I suggest adding ```-f --force``` option to the script, that will scip user confirmation and force installing appropriate versions of dependencies.

JackLivio: @LuxoftAKutsan I agree with your suggestion for prompting the user for confirmation on changing packages. That would be a good improvement.

LuxoftAKutsan:
@ovvasyliev is this changes are already in develop?

setup_env.sh
LuxoftAKutsan: @ovvasyliev  please plan you activities for adding prompt functionality  in setup_env.sh

AByzhynar:
@ovvasyliev @LuxoftAKutsan How long this PR will be opened?

PR #930, opened 2016-11-02
Inclusion of PLAY_PAUSE as a subscribable button

jacobkeeler:
Adds the PLAY_PAUSE button name enum value to the Mobile and HMI APIs, as well as the default HMI capabilities.

### DIFF
--- a/src/components/interfaces/MOBILE_API.xml
+++ b/src/components/interfaces/MOBILE_API.xml
@@ -489,6 +489,7 @@
   <enum name="ButtonName">
     <description>Defines the hard (physical) and soft (touchscreen) buttons available from SYNC</description>
     <element name="OK" />
+    <element name="PLAY_PAUSE" />
     <element name="SEEKLEFT" />
     <element name="SEEKRIGHT" />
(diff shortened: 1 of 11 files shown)


src/components/include/utils/semantic_version.h
```diff
@@ -116,6 +117,8 @@ struct SemanticVersion {
   uint16_t minor_version_ = 0;
   uint16_t patch_version_ = 0;
 };
+
+extern const SemanticVersion version_4_5;
```
jacobkeeler: This doesn't appear necessary

src/components/interfaces/MOBILE_API.xml
```diff
@@ -562,6 +562,7 @@
     <enum name="ButtonName" since="1.0">
         <description>Defines the hard (physical) and soft (touchscreen) buttons available from the module</description>
         <element name="OK" />
+        <element name="PLAY_PAUSE" since="5.0"/>
```
jacobkeeler: Add description, since this replaces `OK` in some places

PR #950, opened 2016-11-04
Remove Ford specific policy endpoint

JackLivio:
New endpoint points to a develpment instance of sdl server that will return a policy table snapshot.


JackLivio:
fix for #1252

PR #576, opened 2016-06-07
Sending button press event based on certain conditions(recommended by Justin)

madhuy:
Sending button press event based on below conditions(recommended by Justin)
a. If no 'appID' -> send to FULL apps only.
b. If 'appID' is present -> send to the named app, only if it’s in FULL or LIMITED? 
c. ignore for background & none even if there is 'appID'

PR #952, opened 2016-11-04
Fix conditions under which OnButtonPress notifications are sent

jacobkeeler:
Update of #576 rebased onto the updated master


mghiumiusliu:
@jacobkeeler Jacob will you merge this PR?

jacobkeeler:
@mghiumiusliu Sorry about not responding earlier, I didn't see your comment until a day or two ago (And I did want some form of confirmation that the updated changes were reviewed before merging).

Were you intending to create two separate pull requests (master and develop) for these changes with #1044? We should be handling hotfixes by merging them into master first and then creating a pull request from master to develop. 

Reverting #1044 and following that procedure would probably cause more issues as it stands, so this is just for future reference.

PR #1042, opened 2016-11-14
Fix tests for ssl context

VVeremjova:
Set encrypted certificates instead of decrypted.
Set client and server certificates separately.
When handshake has just started, we shouldn't expect complete initialisation 

Related to [APPLINK-29755](https://adc.luxoft.com/jira/browse/APPLINK-29755)

Please review @dcherniev, @AGritsevich, @Kozoriz, @dtrunov, @pvvasilev

### DIFF
--- a/src/components/security_manager/test/ssl_context_test.cc
+++ b/src/components/security_manager/test/ssl_context_test.cc
@@ -88,15 +88,23 @@ struct ProtocolAndCipher {
 class SSLTest : public testing::Test {
  protected:
   static void SetUpTestCase() {
-    std::ifstream certificate_file("server/spt_credential_unsigned.p12");
+    SetCertificate("server/spt_credential_unsigned.p12.enc",
+                   server_certificate_data_base64_);
+    SetCertificate("client/client_credential_unsigned.p12.enc",
+                   client_certificate_data_base64_);
+  }
+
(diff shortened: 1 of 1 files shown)


src/components/security_manager/test/ssl_context_test.cc
```diff
@@ -378,20 +388,23 @@ TEST_F(SSLTest, OnTSL2Protocol_BrokenHandshake) {
 // TODO {AKozoriz} : Unexpected uncomplited init of SSL component.
 // In this and next tests.
 // Must be fixed after merge to develop.
-TEST_F(SSLTest, DISABLED_OnTSL2Protocol_Positive) {
+TEST_F(SSLTest, OnTSL2Protocol_Positive) {
   ASSERT_EQ(client_ctx->StartHandshake(&kClientBuf, &client_buf_len),
             security_manager::SSLContext::Handshake_Result_Success);
   ASSERT_FALSE(NULL == kClientBuf);
```
dcherniev: @VVeremjova Maybe this check is unnecessary? Since you are checking this condition in the following while block:
   ASSERT_TRUE(NULL != kClientBuf);

VVeremjova: @dcherniev See 91735d7

PR #1045, opened 2016-11-16
Change version of API in MOBILE_API.xml

dcherniev:
Updated API version and date of version change.
Deleted redundant copy of MOBILE_API.xml file from test_generator folder.

Related tasks: [APPLINK-28276](https://adc.luxoft.com/jira/browse/APPLINK-28276)


dcherniev:
Please review: @vlantonov, @AGritsevich, @Kozoriz, @anosach-luxoft, @dtrunov, @VVeremjova

dcherniev:
Please review: @TMelnyk

TMelnyk:
@dcherniev , reviewed

jacobkeeler:
Should this not be a hotfix in master, as it would be part of the 4.2 release?

dcherniev:
@jacobkeeler You mean that I should rename this branch into: "hotfix/Change_version_of_API" ?

jacobkeeler:
@dcherniev My point is that this change should be put in master rather than develop (which comes with the "feature" to "hotfix" rename) as I would assume that the API version should have been changed to 4.2 in the 4.2.0 release, rather than now.

mghiumiusliu:
@jacobkeeler agree that we should provide this change as hotfix. At least due to the fact that MOBILE API of this revision is presetn in master. However hotfix must be merged into develop and master, and minor version must be increased.

@jacobkeeler Small clarification: we usually agreed version of MOBILE API with FORD, and it wasn't aligned with Open SDL versioning. Do you assume that in future MOBILE API should be aligned with Open SDL version? 
Let me explain this in details, taking into account current situation: we have MOBILE API of version 4.2 in master, however version stated in document is incorrect and we have to fix it (like described above). However this is just a coincidence that MOBILE API has the same version as Open SDL version. Or do you have another understanding?

@jacobkeeler What is Livio preference on hotfix: two pull requests to be sent to master and develop respectively? Or you prefer to have one pull request to master (and sync develop with master later on)?
I personally would prefer 1st option (with two pull request) - it is closer to GitFlow. What do you think?

jacobkeeler:
@mghiumiusliu Up until now, we have been using the latter process of merging hotfixes into master first and later merging master into develop. 

However, we have now decided that we should start using the former method as defined in the GitFlow documentation (with 2 pull requests per hotfix) for the sake of reducing any confusion.

dcherniev:
Closing this pull request due to rename of the branch from "feature" into "hotfix".

PR #1067, opened 2016-11-24
Fix issues related to UTs on CI 

dcherniev:
Added missed MockMessageHelper expectations into SpeakRequestTest;
Added check if logger is available during call of flush_logger method;
Eliminated possibility to try stopping uninitialized connection in TransportManager UT.

Related to: [APPLINK-30284](https://adc.luxoft.com/jira/browse/APPLINK-30284)

### DIFF
--- a/src/components/transport_manager/src/transport_adapter/transport_adapter_impl.cc
+++ b/src/components/transport_manager/src/transport_adapter/transport_adapter_impl.cc
@@ -281,9 +281,11 @@ TransportAdapter::Error TransportAdapterImpl::DisconnectDevice(
        j != to_disconnect.end();
        ++j) {
     ConnectionInfo info = *j;
-    if (OK != info.connection->Disconnect()) {
-      error = FAIL;
-      LOG4CXX_ERROR(logger_, "Error on disconnect " << error);
+    if (info.connection) {
+      if (OK != info.connection->Disconnect()) {
+        error = FAIL;
+        LOG4CXX_ERROR(logger_, "Error on disconnect " << error);
+      }
     }
   }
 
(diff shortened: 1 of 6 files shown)


dcherniev:
Please review: @LuxoftAKutsan, @AByzhynar, @AGaliuzov, @LevchenkoS, @okozlovlux, @dtrunov

src/components/transport_manager/src/transport_adapter/transport_adapter_impl.cc
```diff
@@ -281,9 +281,11 @@ TransportAdapter::Error TransportAdapterImpl::DisconnectDevice(
        j != to_disconnect.end();
        ++j) {
     ConnectionInfo info = *j;
-    if (OK != info.connection->Disconnect()) {
-      error = FAIL;
-      LOG4CXX_ERROR(logger_, "Error on disconnect " << error);
+    if (info.connection) {
```
AByzhynar: @dcherniev What if here info.connection is empty and in next moment in different thread it will be fulfilled?

dcherniev: @AByzhynar Well the info.connection field is uninitialized when the info parameters are created, however due to moving of initialization of this field right after creation of connection, we should always be able to pass this check. If info.connection is empty at the moment we try to send dissconnect we should not try to call something for invalid object. This check is just a reminder that this field can be empty for info paramethers, but in current circumstances this check should always be passed. If this field for some reason would not be valid this will mean that connection would not recieve Disconnect() command, but because we init this field right after creating the connection this should be impossible.

PR #1070, opened 2016-11-25
Add emptylines before EOF

LuxoftAKutsan:


LuxoftAKutsan:
@AGritsevich please review

PR #1077, opened 2016-11-30
Fixes policy update status changing

dev-gh:
During sending out of snapshot with OnSystemRequest policy has set status UPDATING and notify HMI in case prior state was different.

Closes-bug: APPLINK-30383


dev-gh:
@LuxoftAKutsan @AByzhynar @AGritsevich 
Please review

PR #1083, opened 2016-12-01
Fixes policy update status in case of non-consented device

dev-gh:
In case application is registered on non-consented device SDL does not
have to set UPDATE_NEEDED status, but only after device is consented.

Relates-to: APPLINK-30368

### DIFF
--- a/src/components/policy/policy_external/test/policy_manager_impl_user_consent_test.cc
+++ b/src/components/policy/policy_external/test/policy_manager_impl_user_consent_test.cc
@@ -68,9 +68,9 @@ TEST_F(
   // To set UP_TO_DATE before registration
   GetPTU(kValidSdlPtUpdateJson);
 
-  TimevalStruct current_time = date_time::DateTime::getCurrentTime();
+  const TimevalStruct current_time = date_time::DateTime::getCurrentTime();
   const int kSecondsInDay = 60 * 60 * 24;
-  int days_after_epoch = current_time.tv_sec / kSecondsInDay;
+  const int days_after_epoch = current_time.tv_sec / kSecondsInDay;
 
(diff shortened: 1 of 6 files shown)


dev-gh:
@LuxoftAKutsan @AByzhynar 
Please review

src/components/policy/policy_external/test/policy_manager_impl_user_consent_test.cc
```diff
@@ -44,15 +45,48 @@ using ::testing::Return;
...
 TEST_F(
     PolicyManagerImplTest2,
+    AddApplication_AddNewApplicationFromDeviceWithoutConsent_ExpectUpToDate) {
+  // Arrange
+  CreateLocalPT(preloadet_pt_filename_);
+
+  // To set UP_TO_DATE before registration
+  GetPTU(kValidSdlPtUpdateJson);
+
+  TimevalStruct current_time = date_time::DateTime::getCurrentTime();
```
LuxoftAKutsan: @dev-gh it also can be const

src/components/policy/policy_external/test/policy_manager_impl_user_consent_test.cc
```diff
@@ -44,15 +45,48 @@ using ::testing::Return;
...
+  const int kSecondsInDay = 60 * 60 * 24;
+  int days_after_epoch = current_time.tv_sec / kSecondsInDay;
```
LuxoftAKutsan: @dev-gh and that const

dev-gh: @LuxoftAKutsan 
Done with 40c2468

PR #1100, opened 2016-12-09
Fix SDL BC.ActivateApp for NONE level

AByzhynar:
Fixed problem that _SDL BC.ActivateApp_ wasn't sent in case of user replies
"NO" for data consent prompt.
The problem was in additional check that _SDL BC.ActivateApp_ could be sent only if
HMI level was **FULL**. But in case of negative user reply HMI level must be **NONE**.
Therefore this check was deleted.

Related: [APPLINK-30617](https://adc.luxoft.com/jira/browse/APPLINK-30617)


AByzhynar:
@AGaliuzov @Kozoriz @orhan-mehmedov @dev-gh @VProdanov Please review, guys

AByzhynar:
@AGaliuzov Please merge this PR.

PR #1101, opened 2016-12-09
Move some useful scripts

istoilovgithub:
Move changes from check_style.sh
Move show_disabled.sh script
Move api_compare.py script

Related taks APPLINK-30589


istoilovgithub:
Please review
@LuxoftAKutsan, @AByzhynar

PR #1103, opened 2016-12-09
Fix HMI level resumption for COMMUNICATION apps

tkireva:
After COMMUNICATION application resumption due
to lost Wi-Fi connection from LIMITED HMI level,
the application resumes in BACKGROUND HMI level

When an AudioSource HMI state is created, if the
application is of type COMMUNICATIONS, it gets
LIMITED HMI level.

Related Issue: [APPLINK-24767](https://adc.luxoft.com/jira/browse/APPLINK-24767)

### DIFF
--- a/src/components/application_manager/src/hmi_state.cc
+++ b/src/components/application_manager/src/hmi_state.cc
@@ -175,7 +175,7 @@ mobile_apis::HMILevel::eType AudioSource::hmi_level() const {
                                         HMILevel::HMI_NONE)) {
     return parent()->hmi_level();
   }
-  if (is_navi_app(app_id_)) {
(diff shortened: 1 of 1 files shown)


tkireva:
Please review: @anosach-luxoft, @AByzhynar, @OHerasym, @vlantonov

anosach-luxoft:
@tkireva with your new implementation all `voice_com` application get `LIMITED` state instead of `BACKGROUND` during audio source. Is it expected?

tkireva:
@anosach-luxoft 
Yes, it is expected, according to: [https://adc.luxoft.com/jira/browse/APPLINK-26499](https://adc.luxoft.com/jira/browse/APPLINK-26499)

PR #1114, opened 2016-12-16
Fix invalid PT after update

vlantonov:
- The calls to CheckSnapshotInitialization, IsPredataPolicy and IsDefaultPolicy methods of the PolicyManager introduced uninitialized values in the local Policy Table which in turn became invalid.
- This failed the following Policy Table Updates.
- The adding of uninitialized values was removed.

Related Issue: [APPLINK-30706](https://adc.luxoft.com/jira/browse/APPLINK-30706)

### DIFF
--- a/src/components/policy/policy_external/src/cache_manager.cc
+++ b/src/components/policy/policy_external/src/cache_manager.cc
@@ -130,13 +130,16 @@ bool CacheManager::CanAppKeepContext(const std::string& app_id) const {
 uint32_t CacheManager::HeartBeatTimeout(const std::string& app_id) const {
   CACHE_MANAGER_CHECK(0);
   uint32_t result = 0;
-  if (IsApplicationRepresented(app_id)) {
-    if (pt_->policy_table.app_policies_section.apps[app_id]
-            .heart_beat_timeout_ms.is_initialized()) {
-      result = *(pt_->policy_table.app_policies_section.apps[app_id]
-                     .heart_beat_timeout_ms);
-    }
+  if (!IsApplicationRepresented(app_id)) {
+    return result;
   }
+
+  const policy_table::ApplicationPolicies::mapped_type& app =
+      pt_->policy_table.app_policies_section.apps[app_id];
+  if (app.heart_beat_timeout_ms.is_initialized()) {
+    result = *(app.heart_beat_timeout_ms);
+  }
+
   return result;
 }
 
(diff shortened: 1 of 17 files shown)


vlantonov:
Please review: @AByzhynar , @AGaliuzov , @OHerasym , @Kozoriz , @anosach-luxoft

vlantonov:
@dev-gh optional: Please also review

vlantonov:
Rebuild required

src/components/policy/policy_external/src/cache_manager.cc
```diff
@@ -1888,13 +1898,19 @@ bool CacheManager::SetPredataPolicy(const std::string& app_id) {
...
   // permissions or check string value from get_string()
-  policy_table::ApplicationParams& pre_data_app =
-      pt_->policy_table.app_policies_section.apps[kPreDataConsentId];
-  policy_table::ApplicationParams& specific_app =
-      pt_->policy_table.app_policies_section.apps[app_id];
+  if (!IsApplicationRepresented(app_id)) {
+    return false;
+  }
+
+  policy_table::ApplicationPolicies& apps =
```
AByzhynar: @vlantonov Can't see that apps is changed afterwards. So why not `const`?

src/components/policy/policy_regular/src/cache_manager.cc
```diff
@@ -1217,13 +1224,19 @@ bool CacheManager::SetPredataPolicy(const std::string& app_id) {
...
   // permissions or check string value from get_string()
-  policy_table::ApplicationParams& pre_data_app =
-      pt_->policy_table.app_policies_section.apps[kPreDataConsentId];
-  policy_table::ApplicationParams& specific_app =
-      pt_->policy_table.app_policies_section.apps[app_id];
+  if (!IsApplicationRepresented(app_id)) {
+    return false;
+  }
+
+  policy_table::ApplicationPolicies& apps =
```
AByzhynar: @vlantonov The same questions as above.`const`?

src/components/policy/policy_external/src/cache_manager.cc
vlantonov: @AByzhynar The use of `operator []` requires that `apps` should not be `const`. The constness may be achieved by using iterators combined with `find()` container method. However this requires a check for existence, which is already done in `IsApplicationRepresented` . So the check should not be performed twice and this implementation was selected. The constness may also be achieved by using the `at()` method - however it is from C++11 standard.

src/components/policy/policy_regular/src/cache_manager.cc
vlantonov: @AByzhynar The use of `operator []` requires that `apps` should not be `const`. The same answer applies like the one for the previous question.

src/components/policy/policy_external/src/cache_manager.cc
AByzhynar: @vlantonov Thank you.

vlantonov:
Mandatory reviewer @Kozoriz was excluded from review of this pull request as he is not available at work.

PR #1116, opened 2016-12-20
Fix sending URLs of unregister apps in GetURLs

tkireva:
Although there is a check for unregister apps, when the
URLs are gathered, their links appear in the response,
because the URL counting index increments independently
of the type of the application.

The URL counting index is decremented if current app is
not registered, so that its links are skipped.

Related-issues: APPLINK-30883


tkireva:
Please review: @anosach-luxoft, @AByzhynar, @OHerasym, @vlantonov

src/components/application_manager/src/commands/hmi/get_urls.cc
```diff
@@ -88,30 +88,34 @@ void GetUrls::ProcessServiceURLs(const policy::EndpointUrls& endpoints) {
 
   size_t index = 0;
```
AByzhynar: @tkireva 
Move it to the place before its first usage

PR #1121, opened 2016-12-21
Fix proprietary PT snapshot validation

VVeremjova:
Add missing mandatory parameters for module_meta
Fix case sensitive parameter in PTS 
Remove from PTS for proprietary flow parameters from External polices

Related to APPLINK-30870


VVeremjova:
@dev-gh, @AGaliuzov, @AByzhynar
Please review

PR #1127, opened 2016-12-23
Fix sending same URL in OnSystemRequest retry

tkireva:
Fix sending same URL in OnSystemRequest retry
    
    SDL sends same URL in every OnSystemRequest retry, because
    for every message is used first url of the first application.
    
    As a fix, for each message is used the next URL from the current
    application URLs vector, starting from first valid application.
    It there is no more URLs for this application, the function looks
    for the next registered application (or default) and begin cycling
    through its URLs.
    
    Related-issues: [APPLINK-30859](https://adc.luxoft.com/jira/browse/APPLINK-30859)

### DIFF
--- a/src/components/application_manager/src/policies/policy_handler.cc
+++ b/src/components/application_manager/src/policies/policy_handler.cc
@@ -1287,6 +1287,7 @@ void PolicyHandler::OnSnapshotCreated(const BinaryMessage& pt_string) {
   static size_t current_app = 0;
   static size_t current_url = 0;
   if (current_url >= urls.at(current_app).url.size()) {
+    bool is_default, is_registered, has_urls;
     ApplicationSharedPtr app;
     current_url = 0;
     do {
@@ -1295,8 +1296,11 @@ void PolicyHandler::OnSnapshotCreated(const BinaryMessage& pt_string) {
       }
       app =
           application_manager_.application_by_policy_id(urls.at(current_app).app_id);
-    } while (!((urls.at(current_app).app_id == policy::kDefaultId) ||
-        (app && app->IsRegistered() && !urls.at(current_app).url.empty())));
+
+      is_default = urls.at(current_app).app_id == policy::kDefaultId;
+      is_registered = app && app->IsRegistered();
+      has_urls = !urls.at(current_app).url.empty();
+    } while (!(is_default || (is_registered && has_urls)));
   }
 
   SendMessageToSDK(pt_string, urls.at(current_app).url.at(current_url));
--- a/src/components/application_manager/src/policies/policy_handler.cc
+++ b/src/components/application_manager/src/policies/policy_handler.cc
@@ -1283,7 +1283,24 @@ void PolicyHandler::OnSnapshotCreated(const BinaryMessage& pt_string) {
     LOG4CXX_ERROR(logger_, "Service URLs are empty! NOT sending PT to mobile!");
     return;
   }
-  SendMessageToSDK(pt_string, urls.front().url.front());
+
+  static size_t current_app = 0;
+  static size_t current_url = 0;
+  if (current_url >= urls.at(current_app).url.size()) {
+    ApplicationSharedPtr app;
+    current_url = 0;
+    do {
+      if (++current_app >= urls.size()) {
+        current_app = 0;
+      }
+      app =
+          application_manager_.application_by_policy_id(urls.at(current_app).app_id);
+    } while (!((urls.at(current_app).app_id == policy::kDefaultId) ||
+        (app && app->IsRegistered() && !urls.at(current_app).url.empty())));
+  }
+
+  SendMessageToSDK(pt_string, urls.at(current_app).url.at(current_url));
+  current_url++;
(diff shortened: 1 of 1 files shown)


tkireva:
Please review:
@anosach-luxoft, @AByzhynar, @OHerasym

src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -1283,7 +1283,24 @@ void PolicyHandler::OnSnapshotCreated(const BinaryMessage& pt_string) {
...
+    ApplicationSharedPtr app;
+    current_url = 0;
+    do {
+      if (++current_app >= urls.size()) {
+        current_app = 0;
+      }
+      app =
+          application_manager_.application_by_policy_id(urls.at(current_app).app_id);
+    } while (!((urls.at(current_app).app_id == policy::kDefaultId) ||
+        (app && app->IsRegistered() && !urls.at(current_app).url.empty())));
```
AByzhynar: @tkireva For me this `whule` condition is too complicated and unreadable. I would suggest to create several bool variables and therefore simplify it.

src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -1283,7 +1283,28 @@ void PolicyHandler::OnSnapshotCreated(const BinaryMessage& pt_string) {
     LOG4CXX_ERROR(logger_, "Service URLs are empty! NOT sending PT to mobile!");
     return;
   }
-  SendMessageToSDK(pt_string, urls.front().url.front());
+
+  static size_t current_app = 0;
+  static size_t current_url = 0;
+  if (current_url >= urls.at(current_app).url.size()) {
+    bool is_default, is_registered, has_urls;
```
AByzhynar: @tkireva 
- Move variables to the place before their first usage. 
- Please make them `const`
- Use assignment during declaration.

src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -1283,7 +1283,28 @@ void PolicyHandler::OnSnapshotCreated(const BinaryMessage& pt_string) {
...
+    ApplicationSharedPtr app;
+    current_url = 0;
+    do {
+      if (++current_app >= urls.size()) {
+        current_app = 0;
+      }
+      app =
+          application_manager_.application_by_policy_id(urls.at(current_app).app_id);
+
+      is_default = urls.at(current_app).app_id == policy::kDefaultId;
```
AByzhynar: @tkireva I would suggest to add round brackets after `=` ( expr == var) ;

src/components/application_manager/src/policies/policy_handler.cc
tkireva: @AByzhynar 
I can't make them const, because I change their value

AByzhynar: @tkireva Yeah. I see. Ok.

src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -1283,7 +1283,31 @@ void PolicyHandler::OnSnapshotCreated(const BinaryMessage& pt_string) {
     LOG4CXX_ERROR(logger_, "Service URLs are empty! NOT sending PT to mobile!");
     return;
   }
-  SendMessageToSDK(pt_string, urls.front().url.front());
+
+  static size_t current_app = 0;
```
anosach-luxoft: @tkireva could you please give some explanation for next code. It's not very clear what do you want to do. Why variables are `static`?

tkireva: @anosach-luxoft 
On every OnSystemRequest retry SDL must change the URL it sends, walking sequentially through the URLs of all registered applications and the default, so that it must remember which is the next URL between the function callings. This is the reason why the variables are static.

anosach-luxoft: @tkireva it's really hard to understand this code. It's unreadable. I'm not insist hardly but it would be better to rework this loop to decrease conditions complexity

tkireva: @anosach-luxoft 
I agree with you that the loop condition is complex, but there are 2 conditions while the loop should iterate:
1) the default application is found
or
2) registered application with urls is found.
I think that after requested changes from AByzhynar (temporary boolean variables) it is more readable.

src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -1283,7 +1283,34 @@ void PolicyHandler::OnSnapshotCreated(const BinaryMessage& pt_string) {
...
+  static size_t current_url = 0;
+  if (current_url >= urls.at(current_app).url.size()) {
```
LuxoftAKutsan: @tkireva please change implementation of this code, it is rely confusing. If you have no Ideas how to implement it in other way (without static variables at least), please contact me or Orhan or Ivo or Galiuzov or Vladislav. Feel free to discuss you technical solutions with other developers. Probably create additional task to re factor this  piece of code.

PR #1133, opened 2016-12-27
Change copy constructor and operator=

dtrunov:
Change copy constructor and operator= in order to exclude problem
with removing value state during copying integer object


dtrunov:
@dev-gh @VVeremjova please review

dtrunov:
@VVeremjova @dev-gh Please review the lattest changes

PR #1137, opened 2016-12-28
Fix RegisterApp flow

VVeremjova:
According [APPLINK-17927](https://adc.luxoft.com/jira/browse/APPLINK-17927) only after response to registration and sending OnAppRegistered notification PTU should be started

Related to [APPLINK-30884](https://adc.luxoft.com/jira/browse/APPLINK-30884)

### DIFF
--- a/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
+++ b/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
@@ -638,12 +638,11 @@ void RegisterAppInterfaceRequest::SendRegisterAppInterfaceResponseToMobile() {
   SendOnAppRegisteredNotificationToHMI(
       *(application.get()), resumption, need_restore_vr);
 
-  // Start PTU after successfull registration
-  GetPolicyHandler().AddApplication(application->policy_app_id());
   // Default HMI level should be set before any permissions validation, since it
   // relies on HMI level.
   application_manager_.OnApplicationRegistered(application);
 
+  // Start PTU after successfull registration
   // Sends OnPermissionChange notification to mobile right after RAI response
   // and HMI level set-up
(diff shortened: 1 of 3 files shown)


VVeremjova:
@dev-gh, @dtrunov, @AByzhynar Please review

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -638,6 +638,8 @@ void RegisterAppInterfaceRequest::SendRegisterAppInterfaceResponseToMobile() {
   SendOnAppRegisteredNotificationToHMI(
       *(application.get()), resumption, need_restore_vr);
 
+  // Start PTU after successfull registration
+  GetPolicyHandler().AddApplication(application->policy_app_id());
```
dev-gh: @VVeremjova 
I'm thinking of moving that to PolicyHandler::OnAppRegisteredOnMobile. What do you think?

VVeremjova: @dev-gh I'm not sure that it will be more suitable - after starting PTU we should send OnApplicationRegistered and then OnPermissionChange, so if I move AddApplication to OnAppRegisteredOnMobile then I have to move also OnApplicationRegistered, or the flow will become invalid.

dev-gh: @VVeremjova 
As I understand BC.OnApplicationRegistered is sent by SendOnAppRegisteredNotificationToHMI which is called before AddApplication, so it won't affect anything.

VVeremjova: @dev-gh See  	c30e526

PR #1146, opened 2016-12-30
Fixes choosing of application to request backend for policy update

dev-gh:
SDL must choose application randomly from application having different
from NONE HMI level. In case no such apps - random from having NONE will
be choosen.

Closes-bug: APPLINK-31036, APPLINK-31037

### DIFF
--- a/src/components/application_manager/include/application_manager/policies/policy_handler.h
+++ b/src/components/application_manager/include/application_manager/policies/policy_handler.h
@@ -41,6 +41,7 @@
 #include <stdint.h>
 
 #include "policy/policy_manager.h"
+#include "application_manager/application.h"
 #include "application_manager/policies/policy_handler_interface.h"
 #include "application_manager/policies/policy_event_observer.h"
 #include "application_manager/policies/delegates/statistics_delegate.h"
@@ -294,8 +295,9 @@ class PolicyHandler : public PolicyHandlerInterface,
   void OnSystemError(int code) OVERRIDE;
 
   /**
-   * @brief Choose application id to be used for snapshot sending
-   * @return Application id or 0, if there are no applications registered
+   * @brief Chooses random application id to be used for snapshot sending
+   * considering HMI level
+   * @return Application id or 0, if there are no suitable applications
    */
   uint32_t GetAppIdForSending() const OVERRIDE;
 
@@ -475,6 +477,17 @@ class PolicyHandler : public PolicyHandlerInterface,
    * @brief Link all currently registered applications
    */
   void LinkAppsToDevice();
+
+  typedef std::vector<application_manager::ApplicationSharedPtr> Applications;
+
+  /**
+   * @brief Chooses application suitable for forwarding request for policy
+   * update to backend by checking its registration status and device consent
+   * @param app_list List of applications to choose from, list can be modified
+   * in the process of lookup
+   * @return Application id or 0 if no suitable application was found
+   */
+  uint32_t GetApplicationIdForPolicyUpdate(Applications app_list) const;
 
  private:
   class StatisticManagerImpl : public usage_statistics::StatisticsManager {
--- a/src/components/application_manager/include/application_manager/policies/policy_handler.h
+++ b/src/components/application_manager/include/application_manager/policies/policy_handler.h
@@ -481,13 +481,20 @@ class PolicyHandler : public PolicyHandlerInterface,
   typedef std::vector<application_manager::ApplicationSharedPtr> Applications;
 
   /**
-   * @brief Chooses application suitable for forwarding request for policy
-   * update to backend by checking its registration status and device consent
-   * @param app_list List of applications to choose from, list can be modified
-   * in the process of lookup
-   * @return Application id or 0 if no suitable application was found
+   * @brief Checks application registration status (SDL4.0) and device consent
+   * to find out whether application is suitable
+   * @param value Item from applications collection
+   * @return true if application is suitable, otherwise - false
    */
-  uint32_t GetApplicationIdForPolicyUpdate(Applications app_list) const;
+  bool IsAppSuitableForPolicyUpdate(const Applications::value_type value) const;
+
+  /**
+   * @brief Chooses random application from list using
+   * IsAppSuitableForPolicyUpdate
+   * @param app_list Application collection
+   * @return Application id if suitable is found, otherwise - zero
+   */
+  uint32_t ChooseRandomAppForPolicyUpdate(Applications& app_list) const;
 
  private:
   class StatisticManagerImpl : public usage_statistics::StatisticsManager {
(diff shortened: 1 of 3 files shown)


dev-gh:
@AByzhynar @AGaliuzov @LuxoftAKutsan
Reminder

src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -379,40 +373,47 @@ bool PolicyHandler::ClearUserConsent() {
...
+      accessor.begin(),
+      accessor.end(),
+      std::inserter(apps_with_none_level, apps_with_none_level.begin()),
+      has_none_level);
+
+  LOG4CXX_DEBUG(
+      logger_,
+      "Number of apps with NONE level: " << apps_with_none_level.size());
+
+  std::random_shuffle(apps_with_none_level.begin(), apps_with_none_level.end());
```
alexkutsan: @dev-gh do we rely need this randomization? I can cause floating bugs.

dev-gh: @LuxoftAKutsan 
This is a requirement statement. I don't quite understand what bug do you mean? Apps protected by accessor, so they all will be alive during choosing, if that was your concern.

src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -112,7 +112,20 @@ const policy::DeviceParams GetDeviceParams(
   device_params.device_handle = device_handle;
   return device_params;
 }
-}
+
+struct HMILevelPredicate
+    : public std::unary_function<ApplicationSharedPtr, bool> {
+  HMILevelPredicate(const mobile_api::HMILevel::eType level) : level_(level) {}
```
AByzhynar: @dev-gh Looks that here should be added `explicit` keyword according SDL coding style

src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -112,7 +112,20 @@ const policy::DeviceParams GetDeviceParams(
   device_params.device_handle = device_handle;
   return device_params;
 }
-}
+
+struct HMILevelPredicate
+    : public std::unary_function<ApplicationSharedPtr, bool> {
```
AByzhynar: @dev-gh 
Missed appropriate header file include `< functional >`

src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -379,40 +373,47 @@ bool PolicyHandler::ClearUserConsent() {
...
-      const uint32_t app_id = (*first)->app_id();
-      DeviceParams device_params = GetDeviceParams(
-          (*first)->device(),
-          application_manager_.connection_handler().get_session_observer());
+  HMILevelPredicate has_none_level(mobile_api::HMILevel::HMI_NONE);
+  Applications apps_without_none_level;
+  std::copy_if(
+      accessor.begin(),
+      accessor.end(),
+      std::inserter(apps_without_none_level, apps_without_none_level.begin()),
```
AByzhynar: @dev-gh 
 - `std::back_inserter` returns `std::back_insert_iterator` that uses `Container::push_back()`.
 - `std::inserter` returns `std::insert_iterator` that uses `Container::insert()`
So if we are using `std::vector` - why not to use `std::back_inserter` ?

src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -379,40 +373,47 @@ bool PolicyHandler::ClearUserConsent() {
...
+  if (app_id) {
+    return app_id;
   }
-  return 0;
+
+  Applications apps_with_none_level;
+  std::copy_if(
+      accessor.begin(),
+      accessor.end(),
+      std::inserter(apps_with_none_level, apps_with_none_level.begin()),
```
AByzhynar: @dev-gh Same as above..

src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -685,6 +686,49 @@ void PolicyHandler::LinkAppsToDevice() {
...
+  Applications::const_iterator it_app = app_list.begin();
+
+  const ApplicationSharedPtr application_ptr = *it_app;
+  const uint32_t app_id = application_ptr->app_id();
+
+  if (!application_ptr->IsRegistered()) {
+    LOG4CXX_DEBUG(logger_,
+                  "Application " << app_id << " is not marked as registered.");
+    app_list.erase(it_app);
+    return GetApplicationIdForPolicyUpdate(app_list);
```
AByzhynar: @dev-gh Do we really need to use recursion? Why not to use just for loop?

src/components/application_manager/src/policies/policy_handler.cc
dev-gh: @AByzhynar 
Fixed with 891a098

src/components/application_manager/src/policies/policy_handler.cc
dev-gh: @AByzhynar 
Fixed with 891a098

src/components/application_manager/src/policies/policy_handler.cc
dev-gh: @AByzhynar 
Fixed with 891a098

src/components/application_manager/src/policies/policy_handler.cc
dev-gh: @AByzhynar 
Fixed with 891a098

src/components/application_manager/src/policies/policy_handler.cc
dev-gh: @AByzhynar 
I think code is cleaner in that way.

AByzhynar: @dev-gh I thin we can just loop over container elements even **without erasing**. We just need to find first app which suits our needs. If found - exit from the loop and proceed it, if not - next loop iteration.

src/components/application_manager/src/policies/policy_handler.cc
LuxoftAKutsan: @dev-gh I mean that in case of some but, it will be difficult to reproduce it in the same flow.

src/components/application_manager/include/application_manager/policies/policy_handler.h
```diff
@@ -476,6 +478,17 @@ class PolicyHandler : public PolicyHandlerInterface,
...
+  typedef std::vector<application_manager::ApplicationSharedPtr> Applications;
+
+  /**
+   * @brief Chooses application suitable for forwarding request for policy
+   * update to backend by checking its registration status and device consent
+   * @param app_list List of applications to choose from, list can be modified
+   * in the process of lookup
+   * @return Application id or 0 if no suitable application was found
+   */
+  uint32_t GetApplicationIdForPolicyUpdate(Applications app_list) const;
```
AGaliuzov: @dev-gh collection by value?

src/components/application_manager/src/policies/policy_handler.cc
dev-gh: @AByzhynar 
Check out improved version 6d53b62

PR #1148, opened 2017-01-03
Fix incorrect choice set with duplicate vrCommands

indlu:
Fix incorrect successfully creation ChoiseSet with duplicate vrCommands inside it.

 Related task APPLINK-25597

### DIFF
--- a/src/components/application_manager/src/commands/mobile/create_interaction_choice_set_request.cc
+++ b/src/components/application_manager/src/commands/mobile/create_interaction_choice_set_request.cc
@@ -129,22 +129,31 @@ mobile_apis::Result::eType CreateInteractionChoiceSetRequest::CheckChoiceSet(
   const SmartArray* choice_set =
       (*message_)[strings::msg_params][strings::choice_set].asArray();
 
-  SmartArray::const_iterator choice_set_it = choice_set->begin();
+  SmartArray::const_iterator current_choice_set_it = choice_set->begin();
+  SmartArray::const_iterator next_choice_set_it;
 
-  for (; choice_set->end() != choice_set_it; ++choice_set_it) {
+  for (; choice_set->end() != current_choice_set_it; ++current_choice_set_it) {
     std::pair<std::set<uint32_t>::iterator, bool> ins_res =
-        choice_id_set.insert((*choice_set_it)[strings::choice_id].asInt());
+        choice_id_set.insert(
+            (*current_choice_set_it)[strings::choice_id].asInt());
     if (!ins_res.second) {
       LOG4CXX_ERROR(logger_,
                     "Choise with ID "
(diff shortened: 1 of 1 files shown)


indlu:
Please review : @LevchenkoS, @dcherniev, @LuxoftAKutsan

src/components/application_manager/src/commands/mobile/create_interaction_choice_set_request.cc
```diff
@@ -146,6 +147,13 @@ mobile_apis::Result::eType CreateInteractionChoiceSetRequest::CheckChoiceSet(
       LOG4CXX_ERROR(logger_, "Incoming choice set has contains \t\n \\t \\n");
       return mobile_apis::Result::INVALID_DATA;
     }
+    for (choice_set_it_next = choice_set_it + 1;
+         choice_set->end() != choice_set_it_next;
+         ++choice_set_it_next) {
```
dcherniev: @indlu The name of this variable is a bit confusing, considering we increase this iterator in the "for block", I would suggest to rename it into something like: "other_element_of_choice_set"

LevchenkoS: @dcherniev, @indlu, I would rather suggest to rename this variables to `current_choice` and `next_choice`.

indlu:
@LevchenkoS, fixed in 9ae404a

src/components/application_manager/src/commands/mobile/create_interaction_choice_set_request.cc
indlu: @dcherniev, fixed in 9ae404a

indlu:
Rebuild

src/components/application_manager/src/commands/mobile/create_interaction_choice_set_request.cc
dcherniev: @indlu, @LevchenkoS I think it is better to name them exactly current_choice and next_choice since this variant is shorter, and more clear. Also variant in 9ae404a doesn't make much of a difference from the original.

indlu:
Rebuild required

src/components/application_manager/src/commands/mobile/create_interaction_choice_set_request.cc
indlu: @dcherniev, I still prefer not to change the variables names because _set  prefix indicates that we have a choiseSet and _it prefix indicates that we use an iterator.

dcherniev: @indlu OK

indlu:
Rebuild required

indlu:
Rebuild required

PR #1153, opened 2017-01-05
Fix proprietary logic defines for GetURLs

dev-gh:
Also reverts changes done before for APPLINK-30883 since related CRSs are not implemented yet.


dev-gh:
@LuxoftAKutsan 
Please review

Issue #2, opened 2014-10-11
BlueZ 5.x transport adapter


justinjdickow:
Should implement a separate transport adapter that supports OEMs using BlueZ 5.x.  Known issues so far with the current implementation of BlueZ 4.x are that bt-device is not supported for querying paired devices and connecting with the BlueZ 5.x dev APIs.

JackLivio:
Bluez 5 is not supported on the current version of Ubuntu 14. When the sdl core supported environment is moved to Ubuntu 16, Bluez will be upgraded to version 5.

PR #1158, opened 2017-01-06
BlueZ 5.0+ modifications

jacobkeeler:
The AGL platform does not have bluez-utils available, meaning that it cannot use the bt-device command to find paired devices. This PR adds an alternative way of finding paired devices using the bluetoothctl shell which was introduced with BlueZ 5.

Required for AGL Port.

### DIFF
--- a/src/components/transport_manager/src/bluetooth/bluetooth_device_scanner.cc
+++ b/src/components/transport_manager/src/bluetooth/bluetooth_device_scanner.cc
@@ -60,6 +60,23 @@ namespace transport_adapter {
 CREATE_LOGGERPTR_GLOBAL(logger_, "TransportManager")
 
 namespace {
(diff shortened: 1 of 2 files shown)


jacobkeeler:
Fixes #2

src/components/transport_manager/src/bluetooth/bluetooth_device_scanner.cc
```diff
@@ -60,6 +60,23 @@ namespace transport_adapter {
 CREATE_LOGGERPTR_GLOBAL(logger_, "TransportManager")
 
 namespace {
+#ifdef BLUEZ5
+int ValidateAddr(char* dev_list_entry) {
```
LuxoftAKutsan: @jacobkeeler seems that this functions required some unit tests

PR #1159, opened 2017-01-06
AGL Port

jacobkeeler:
Adds CMake configurations which are necessary to cross-compile SDL Core for AGL.

After setting up your cross-compilation environment, SDL Core can now be compiled for AGL (running on the Renesas RCar M2 board) using the provided toolchain.

```bash
cmake ../sdl_core -DCMAKE_TOOLCHAIN_FILE=../sdl_core/rcar_m2_agl_linux.cmake
```


durandaltheta:
As a helpful note to you folks, we at Elektrobit Automotive Americas Inc. have already created a working port of SDL for AGL at https://github.com/pkeb/sdl_core forked from this repository. The fork was created for an internal PoC of SDL on AGL which we demoed at CES2017 and the a Genivi convention. The readme has been updated with extensive documentation on how to get it working.

High level notes to be aware of:
1. The Elektrobit port uses AGL's crosssdk for all compilation toolchain needs (and 99% of libraries). This makes it fit well within the AGL universe but is quite a bit of extra work if the user only needs to compile SDL and nothing else.
2. There were numerous changes that needed to be made to the cmake and 3rd party make builds to make it AGL crosssdk safe (including one or two code changes). These changes may not be backwards compatible with QNX and x86_64 compilation and will almost certainly need to be tweaked if incorporated into the mainline.
3. In particular there is hack to the log4cxx code that forces a particular definition of int32 and int64 structs (as recognized by the compiler).  The problem this fixes may have been caused by my heavy handed copying of the QNX configure arguments during the porting process and may have a cleaner possible fix by tweaking sdl_core/src/3rd_party/apr-cmake/CMakeLists.txt (possibly by removing "ac_cv_sizeof_struct_iovec=8"?)

chinlin0924:
hi @jacobkeeler  is there any plan to have a Yocto Port of SDL Core officially?

jacobkeeler:
@chinlin0924 This port was created before the SDLC was formed, so the process of determining what to include with each release wasn't as strict at the time. It seems to me that we would need to create a SDL Evolution proposal to ever make this port official at this point. Because no such proposal has been submitted yet, there are no plans for an official port as of now. When I have the time, I may be able to create this proposal, but if you strongly need this port you are free to submit a proposal yourself.

chinlin0924:
thanks for the info!

2018-04-06 4:40 GMT+08:00 Jacob Keeler <notifications@github.com>:

PR #1168, opened 2017-01-11
Remove redundant notification for mobile app

dcherniev:
Removed redundant OnAppInterfaceUnregistered notification for mobile application,
which was sent in case SDL recieved OnExitApplication with exit reason - DRIVER_DISTRACTION_VIOLATION.

Related to: [APPLINK-25584](https://adc.luxoft.com/jira/browse/APPLINK-25584)

### DIFF
--- a/src/components/application_manager/src/commands/hmi/on_exit_application_notification.cc
+++ b/src/components/application_manager/src/commands/hmi/on_exit_application_notification.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013, Ford Motor Company
+ * Copyright (c) 2017, Ford Motor Company
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


dcherniev:
Please review: @LuxoftAKutsan, @LevchenkoS, @okozlovlux, @dtrunov

src/components/application_manager/src/commands/hmi/on_exit_application_notification.cc
```diff
@@ -62,17 +62,21 @@ void OnExitApplicationNotification::Run() {
     return;
   }
 
+  HMILevel::eType app_hmi_level =
+      application_manager_.GetDefaultHmiLevel(app_impl);
   Common_ApplicationExitReason::eType reason;
```
dtrunov: @dcherniev I don't understand logic of this fix. Why did you use default hmi level?

dcherniev: @dtrunov You right, I misunderstood what was needed to be done according to requirements, I have removed notification since it is redundant also I made minor cosmetic change at the end of the method. Also I think I need to rename the branch, should I do this after the reviews, or it is better to rename it now?

dcherniev:
Closing this pull request due to rename of the branch.

PR #1170, opened 2017-01-12
Fix SDL omits URL during retry sequence

indlu:
Fix SDL omits one of the URLs while cyclically changing during retry sequence

Related defect APPLINK-31325

### DIFF
--- a/src/components/application_manager/src/policies/policy_handler.cc
+++ b/src/components/application_manager/src/policies/policy_handler.cc
@@ -1388,8 +1388,8 @@ void PolicyHandler::OnSnapshotCreated(const BinaryMessage& pt_string) {
   }
 
   SendMessageToSDK(pt_string, urls.at(current_app).url.at(current_url));
-  if(!(urls.at(current_app).app_id == policy::kDefaultId))
-  current_url++;
+  if (!(urls.at(current_app).app_id == policy::kDefaultId))
+    current_url++;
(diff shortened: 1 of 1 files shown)


indlu:
Please review : @dcherniev,  @LuxoftAKutsan, @dev-gh

src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -1388,7 +1388,8 @@ void PolicyHandler::OnSnapshotCreated(const BinaryMessage& pt_string) {
   }
 
   SendMessageToSDK(pt_string, urls.at(current_app).url.at(current_url));
-  current_url++;
+  if (!(urls.at(current_app).app_id == policy::kDefaultId))
```
LuxoftAKutsan: @indlu use breakets according gstyle

LuxoftAKutsan: @indlu at can throw exeption, can't it?

LuxoftAKutsan: @indlu 
```if (x != y)``` is better then ```if (!(x ==y))```

dev-gh: @indlu 
What if 'default' will have more than one URL? I suppose only first of default will be used and other will be skipped.

indlu:
Closed due to planned changes in functionality.

PR #1197, opened 2017-01-20
Fix TTS capabilities to array

VVeremjova:
According HMI_API.xml TTS capabilities should be array

Related to [APPLINK-31660](https://adc.luxoft.com/jira/browse/APPLINK-31660)

### DIFF
--- a/src/appMain/hmi_capabilities.json
+++ b/src/appMain/hmi_capabilities.json
@@ -343,7 +343,7 @@
 },
     "TTS":
 	{
-		"capabilities":"TEXT",
+		"capabilities":["TEXT"],
 		"language":"EN-US",
 	    "languages":
 	    [


VVeremjova:
Please review @LuxoftAKutsan

jacobkeeler:
@LuxoftAKutsan @VVeremjova This PR results in an error when loading the HMICapabilities file, since the TTS capabilities field is treated as a String rather than an Array when it is loaded. This change requires corresponding modifications to hmi_capabilities_impl.cc in order to work correctly.

LuxoftAKutsan:
@jacobkeeler according HMI_API TTS.capabilities should be array. But according json specifications
```"capabilities":"TEXT"``` is string but not array. Do you think that SDL should manage string as array with one element in it if according HMI_API.xml there should be an array?

jacobkeeler:
@LuxoftAKutsan I'm not saying that it shouldn't be an array, I am saying that it is currently treated as a string by hmi_capabilities_impl.cc (in [this block](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/application_manager/src/hmi_capabilities_impl.cc#L997-L1000)). This means that this component will fail when loading this modified `hmi_capabilities.json`, resulting in every "HMI Capabilities" related RPC (such as SubscribeButton) returning `UNSUPPORTED_RESOURCE`.

hmi_capabilities_impl.cc needs to be modified to read this value as an array rather than a string in order for the change in this PR to not break current functionality.

LuxoftAKutsan:
@jacobkeeler you is right. Appropriate code in SDL is parsing this field as string. It will be fixed.

LuxoftAKutsan:
@jacobkeeler  here is issue according this bug : 
https://github.com/smartdevicelink/sdl_core/issues/1321
And there is pull requests that fixes it : 
 https://github.com/smartdevicelink/sdl_core/pull/1320
https://github.com/smartdevicelink/sdl_core/pull/1322

PR #1246, opened 2017-01-31
Fix missing OnInteriorVehicleData msg to mobile

tkireva:
The reason SDL was not sending OnInteriorVehicleData message to the
mobile application is that the vector with applications it must notify
is empty (with matching UID). It is empty because the data accessor to
all applications set is not initialized. Initializing the accessor
variable fixed the problem.

Related-issues: [SDLOPEN-600](https://adc.luxoft.com/jira/browse/SDLOPEN-600)


tkireva:
Please review:
@vlantonov @okozlovlux @LuxoftAKutsan  @istoilovgithub

okozlovlux:
@indlu , please review

PR #953, opened 2016-11-07
[WiP][Testing]Add sub menu icon parameter

VVeremjova:
PR previously was created and reviewed on #921.  
Please see @AByzhynar, @dcherniev, @AGritsevich, @Kozoriz, @dtrunov, @pvvasilev

Related to [ APPLINK-23379](https://adc.luxoft.com/jira/browse/APPLINK-23379)

PR #1260, opened 2017-02-07
Send INVALID_DATA in case SubMenuIcon is invalid

VVeremjova:
Related to #953
After questions the requirement was clarified

### DIFF
--- a/src/components/application_manager/src/commands/mobile/add_sub_menu_request.cc
+++ b/src/components/application_manager/src/commands/mobile/add_sub_menu_request.cc
@@ -95,8 +95,12 @@ void AddSubMenuRequest::Run() {
       (*message_)[strings::msg_params][strings::menu_name];
   msg_params[strings::app_id] = app->app_id();
 
-  if (((*message_)[strings::msg_params].keyExists(strings::sub_menu_icon)) &&
-      CheckSubMenuIcon()) {
+  if (((*message_)[strings::msg_params].keyExists(strings::sub_menu_icon))) {
+    if (!CheckSubMenuIcon()) {
+      LOG4CXX_ERROR(logger_, "Sub-menu icon is not valid.");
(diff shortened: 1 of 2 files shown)


VVeremjova:
Please review @dtrunov

src/components/application_manager/src/commands/mobile/add_sub_menu_request.cc
```diff
@@ -95,8 +95,12 @@ void AddSubMenuRequest::Run() {
       (*message_)[strings::msg_params][strings::menu_name];
   msg_params[strings::app_id] = app->app_id();
 
-  if (((*message_)[strings::msg_params].keyExists(strings::sub_menu_icon)) &&
-      CheckSubMenuIcon()) {
+  if (((*message_)[strings::msg_params].keyExists(strings::sub_menu_icon))) {
+    if (!CheckSubMenuIcon()) {
+      LOG4CXX_ERROR(logger_, "Sub-menu icon is not valid.");
```
dtrunov: @VVeremjova CheckSubMenuIcon contain redundant checking (*message_)[strings::msg_params].keyExists(strings::sub_menu_icon) please remove it.

VVeremjova: @dtrunov See  1e0e153

LevchenkoS: @VVeremjova, why do you decide to move check for `sub_menu_icon` existence out from `CheckSubMenuIcon()` function?

VVeremjova:
Rebuild required

src/components/application_manager/src/commands/mobile/add_sub_menu_request.cc
VVeremjova: @LevchenkoS If icon does not exist then it is ok. If icon exists and invalid then we should send INVALID_DATA.

PR #1272, opened 2017-02-09
Revert "Send INVALID_DATA in case SubMenuIcon is invalid"

VVeremjova:
Reverts smartdevicelink/sdl_core#1260
Reverts smartdevicelink/sdl_core#953


LuxoftAKutsan:
@jacobkeeler this PR reverts changes that were merged to develop by mistake. In case if revert won't be done this changes will be released within 4.4.0 WP, But it should be in scope of 4.3.1

PR #1284, opened 2017-02-14
Fix sending incorrect path to dynamic image

tkireva:
SDL used to send only icon name in the value of dynamic image
parameter when it is transfered to the HMI in AudioPassThru.
Fix just allowes the image verification funcion to change the value
of the given name.


tkireva:
Please review:
@vlantonov @okozlovlux @istoilovgithub @LevchenkoS

PR #1289, opened 2017-02-15
Fix wrong info parameter AudioPassThru

tkireva:
Fix wrong info parameter AudioPassThru
    
    In case received from mobile side audioPassThruIcon does not exist in
    the application storage folder SDL used to add string "Reference image(s)
    not found" to the info parameter in the response. After clarification it
    turned out that this info must be generated from the HMI, so that the 
    member variable that was used for storing the state of the icon is removed.


tkireva:
Please review:
@dtrunov @vlantonov @okozlovlux @istoilovgithub

PR #1291, opened 2017-02-15
Fix incorrect result code to mobile

tkireva:
SDL used to return wrong result code if only mandatory parameters
are sent from HMI.

### DIFF
--- a/src/components/application_manager/src/commands/mobile/perform_audio_pass_thru_request.cc
+++ b/src/components/application_manager/src/commands/mobile/perform_audio_pass_thru_request.cc
@@ -435,7 +435,8 @@ PerformAudioPassThruRequest::PrepareAudioPassThruResultCodeForResponse(
   const hmi_apis::Common_Result::eType tts_result = tts_response.result_code;
 
   if ((ui_result == hmi_apis::Common_Result::SUCCESS) &&
-      (tts_result != hmi_apis::Common_Result::SUCCESS)) {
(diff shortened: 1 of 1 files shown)


tkireva:
Please review:
@okozlovlux @vlantonov @istoilovgithub @dcherniev

vlantonov:
@tkireva Is this fix covered with Unit test, which succeeds now and fails before changes?

tkireva:
@vlantonov 
Currently this situation is not covered by any test case.

src/components/application_manager/src/commands/mobile/perform_audio_pass_thru_request.cc
```diff
@@ -435,7 +435,8 @@ PerformAudioPassThruRequest::PrepareAudioPassThruResultCodeForResponse(
   const hmi_apis::Common_Result::eType tts_result = tts_response.result_code;
 
   if ((ui_result == hmi_apis::Common_Result::SUCCESS) &&
-      (tts_result != hmi_apis::Common_Result::SUCCESS)) {
+      ((tts_result != hmi_apis::Common_Result::SUCCESS) &&
```
dcherniev: @tkireva I think the additional round brackets here are unnecessary.

tkireva: @dcherniev 
I agree. It is fixed in e285062292aa3bda1ec0e3a166887b02dd0b0dde

PR #1297, opened 2017-02-16
Fix/tts aborted processing

tkireva:
Fix incorrect ABORTED code processing

ABORTED code received from TTS HMI part used to be processed
as exception of other result codes - SDL was transfering it to
the mobile application. After clarification it turned out that
it must be processed as all other error codes - resolved to
WARNINGS


tkireva:
Please review:
@vlantonov @okozlovlux @dcherniev @istoilovgithub

PR #1298, opened 2017-02-16
Disable protocol handler tests

LuxoftAKutsan:
Protocol handler tests are Async and sometimes fail. It should be fixed before enabling


ovvasyliev:
Rebuild required

rememberkarl:
Rebuild? 

What do you mean that rebuild? 

I tried a few times to build and remove and build so on...

It's always happened but I couldn't figure out why...

I'm sorry but do you have any another solution on this problem?

LuxoftAKutsan:
"Rebuild required" - is command for build server to rebuild current pull request.

PR #1308, opened 2017-02-20
Fix missing GENERIC_ERROR message

tkireva:
SDL sends GENERIC_ERROR when the watchdog timeout expires in
PerformAudioPassThru RPC. First problem was that the value of the
timeout is a sum of default timeout value and the value send in
maxDuration parameter. After clarification it turned out that the
maxDuration parameter doesn't have to be added, so the sum statement
is removed. The other problem was that on every TTS.Speak_response
from the HMI the timeout was reset. The clarified behavior defines
that SDL must expect onResetTimeout notification from the HMI and it
doesn't have to reset the timeout when receives TTS.Speak_response.


tkireva:
Please review:
@vlantonov @dcherniev @okozlovlux @istoilovgithub

PR #1311, opened 2017-02-21
Add rejection of PT if it does not support default language

dcherniev:
Added check if PT section consumer_friendly_messages supports default language for all of it's messages,
if this check is failed the PT will be rejected.


dcherniev:
Please review: @LuxoftAKutsan, @okozlovlux

dcherniev:
Rebuild required

dcherniev:
Updated my changes according to solution in policy external, so please review again: @LuxoftAKutsan, @okozlovlux

dcherniev:
Rebuild required

PR #1314, opened 2017-02-22
Fix defect param seconds_between_retries

pold500:
PM should verify that "seconds_between_retries" array has maxlength 5

Changed:
* src/components/policy/policy_regular/src/cache_manager.cc
  Now SDL exits if seconds_between_retries array in sdl_preloaded_pt.json contain more then 5 elements

@AByzhynar please review
@LuxoftAKutsan please review
@dcherniev please review


src/components/policy/policy_regular/src/cache_manager.cc
```diff
@@ -585,6 +585,13 @@ bool CacheManager::SecondsBetweenRetries(std::vector<int>& seconds) {
 
   const std::size_t size =
       pt_->policy_table.module_config.seconds_between_retries.size();
+
+  const std::size_t max_seconds_between_retries_size = 5u;
+  if (size > max_seconds_between_retries_size) {
+    // must exit app
+    exit(EXIT_FAILURE);
```
LuxoftAKutsan: @pold500 I suppose that it is very bad idea to immediately exit application on that case. 
First because this function can be called not only on startup.
Second because you didn't log anything. Just imagine that SDL stops immediately after Start. How would you find issue why it is dound it?

LuxoftAKutsan: I suppose that you should validate length of this array in policy table validation and return is_valid  = falce. And in LifeCycle start you shuld check if policy table validation failed SDL should not be started with appropriate error message

src/components/policy/policy_regular/src/cache_manager.cc
```diff
@@ -585,6 +585,13 @@ bool CacheManager::SecondsBetweenRetries(std::vector<int>& seconds) {
 
   const std::size_t size =
       pt_->policy_table.module_config.seconds_between_retries.size();
+
+  const std::size_t max_seconds_between_retries_size = 5u;
```
tkireva: Could you, please define value 5u as a constant.

src/components/policy/policy_regular/src/cache_manager.cc
tkireva: Is it so critical situation, so that exit application is required?

src/components/policy/policy_regular/src/cache_manager.cc
pold500: @tkireva but it's already defined as constant

tkireva: Sorry for misunderstanding, I am not talking about max_seconds_between_retries_size variable, I see it is const, but the value that you assign to it.

pold500: @tkireva, so you want another const that will be assigned to that const? what the point?

pold500:
@AKutsan please review.

pold500:
@LuxoftAKutsan @tkireva @istoilovgithub Please review changes

pold500:
Rebuild continuous-integration/travis-ci/pr

pold500:
@istoilovgithub "but double check with requirement analyst which of the flows this defects applies to."
But requirements are already described in related jira tickets. Should I ask requirement analyst anyway?

src/components/policy/policy_regular/src/cache_manager.cc
```diff
@@ -585,6 +585,7 @@ bool CacheManager::SecondsBetweenRetries(std::vector<int>& seconds) {
 
   const std::size_t size =
       pt_->policy_table.module_config.seconds_between_retries.size();
+
```
LuxoftAKutsan: @pold500 please remove empty string

PR #1317, opened 2017-02-23
Fix responding Warnings for PerformAudioPassThru

VVeremjova:
In case the response for UI.Alert or TTS.Speak is Warnings and other response is successful
then Warnings should be sent on mobile.

### DIFF
--- a/src/components/application_manager/src/commands/mobile/perform_audio_pass_thru_request.cc
+++ b/src/components/application_manager/src/commands/mobile/perform_audio_pass_thru_request.cc
@@ -414,7 +414,6 @@ PerformAudioPassThruRequest::PrepareAudioPassThruResultCodeForResponse(
     const ResponseInfo& ui_response,
     const ResponseInfo& tts_response,
     bool& out_result) {
-  using namespace helpers;
   mobile_apis::Result::eType result_code = mobile_apis::Result::INVALID_ENUM;
 
   hmi_apis::Common_Result::eType common_result =
@@ -428,13 +427,11 @@ PerformAudioPassThruRequest::PrepareAudioPassThruResultCodeForResponse(
     common_result = hmi_apis::Common_Result::WARNINGS;
     out_result = true;
   }
-
   else if (ui_response.is_ok &&
            tts_result == hmi_apis::Common_Result::WARNINGS) {
     common_result = hmi_apis::Common_Result::WARNINGS;
     out_result = true;
   }
-
   else if (ui_result == hmi_apis::Common_Result::INVALID_ENUM) {
     common_result = tts_result;
   } else {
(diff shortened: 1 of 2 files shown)


src/components/application_manager/src/commands/mobile/perform_audio_pass_thru_request.cc
```diff
@@ -426,7 +427,27 @@ PerformAudioPassThruRequest::PrepareAudioPassThruResultCodeForResponse(
       (tts_result != hmi_apis::Common_Result::INVALID_ENUM)) {
     common_result = hmi_apis::Common_Result::WARNINGS;
     out_result = true;
-  } else if (ui_result == hmi_apis::Common_Result::INVALID_ENUM) {
+  }
+
+  else if ((ui_result == hmi_apis::Common_Result::WARNINGS ||
+            tts_result == hmi_apis::Common_Result::WARNINGS) &&
+           Compare<hmi_apis::Common_Result::eType, EQ, ONE>(
```
tkireva: @VVeremjova I think you can use is_ok flag of the ResponseInfo structure instead of the Compare<>. It is set in ResponseInfo constructor. It will make this statement much more readable.

tkireva: @VVeremjova
The case when ui_result is WARNINGS is covered by the last else statement (line 452), so here it must be excluded. I think you must check if ui_result.is_ok and tts_result == WARNINGS.

VVeremjova: @tkireva Please see  31b88b6

src/components/application_manager/src/commands/mobile/perform_audio_pass_thru_request.cc
```diff
@@ -414,6 +414,7 @@ PerformAudioPassThruRequest::PrepareAudioPassThruResultCodeForResponse(
     const ResponseInfo& ui_response,
     const ResponseInfo& tts_response,
     bool& out_result) {
+  using namespace helpers;
```
tkireva: Could you, please remove this namespace if it not necessary any more.

src/components/application_manager/src/commands/mobile/perform_audio_pass_thru_request.cc
```diff
@@ -426,7 +427,15 @@ PerformAudioPassThruRequest::PrepareAudioPassThruResultCodeForResponse(
       (tts_result != hmi_apis::Common_Result::INVALID_ENUM)) {
     common_result = hmi_apis::Common_Result::WARNINGS;
     out_result = true;
-  } else if (ui_result == hmi_apis::Common_Result::INVALID_ENUM) {
+  }
+
```
tkireva: Could you, please remove empty lines

VVeremjova: @tkireva See 6b95b47

PR #1326, opened 2017-02-27
Fix wrong dynamic image parameter

tkireva:
SDL didn't apply the full image path if it doesn't
exist in the file system. VerifyImage wrapper function
is created that overwrite the image name with path + name
if the image is dynamic regardless of its existence.

### DIFF
--- a/src/components/application_manager/src/message_helper/message_helper.cc
+++ b/src/components/application_manager/src/message_helper/message_helper.cc
@@ -2307,16 +2307,16 @@ mobile_apis::Result::eType MessageHelper::VerifyImage(
     smart_objects::SmartObject& image,
     ApplicationConstSharedPtr app,
     ApplicationManager& app_mngr) {
-  smart_objects::SmartObject img = image;
+  smart_objects::SmartObject temp_image = image;
   const uint32_t image_type = image[strings::image_type].asUInt();
-  mobile_apis::ImageType::eType type =
+  const mobile_apis::ImageType::eType type =
       static_cast<mobile_apis::ImageType::eType>(image_type);
-  mobile_apis::Result::eType result;
 
-  result = VerifyImageApplyPath(img, app, app_mngr);
+  const mobile_apis::Result::eType result =
+      VerifyImageApplyPath(temp_image, app, app_mngr);
   if ((mobile_apis::Result::SUCCESS == result) &&
       (mobile_apis::ImageType::DYNAMIC == type)) {
-    image[strings::value] = img[strings::value];
+    image[strings::value] = temp_image[strings::value];
   }
 
   return result;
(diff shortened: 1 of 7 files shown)


tkireva:
Please review:
@vlantonov @okozlovlux @istoilovgithub @VVeremjova

src/components/application_manager/src/message_helper/message_helper.cc
```diff
@@ -2295,15 +2295,33 @@ mobile_apis::Result::eType MessageHelper::VerifyImage(
...
+    smart_objects::SmartObject& image,
+    ApplicationConstSharedPtr app,
+    ApplicationManager& app_mngr) {
+  smart_objects::SmartObject img = image;
+  const uint32_t image_type = image[strings::image_type].asUInt();
+  mobile_apis::ImageType::eType type =
+      static_cast<mobile_apis::ImageType::eType>(image_type);
+  mobile_apis::Result::eType result;
+
+  result = VerifyImageApplyPath(img, app, app_mngr);
```
vlantonov: @tkireva It seems that `result` is not changed after the initialization, so it can be made `const`. The same applies to `type` variable above, where a name like `image_type` will be more informative.

src/components/application_manager/src/message_helper/message_helper.cc
```diff
@@ -2295,15 +2295,33 @@ mobile_apis::Result::eType MessageHelper::VerifyImage(
...
-  image[strings::value] = full_file_path;
-
   return mobile_apis::Result::SUCCESS;
 }
 
+mobile_apis::Result::eType MessageHelper::VerifyImage(
+    smart_objects::SmartObject& image,
+    ApplicationConstSharedPtr app,
+    ApplicationManager& app_mngr) {
+  smart_objects::SmartObject img = image;
```
vlantonov: @tkireva Consider using a more informative name like `work_image`, `temp_image`, etc.

src/components/application_manager/src/message_helper/message_helper.cc
```diff
@@ -2251,7 +2251,7 @@ mobile_apis::Result::eType MessageHelper::VerifyImageFiles(
   return mobile_apis::Result::SUCCESS;
 }
 
-mobile_apis::Result::eType MessageHelper::VerifyImage(
+mobile_apis::Result::eType MessageHelper::VerifyImageApplyPath(
```
vlantonov: @tkireva Please modify the previous Unit test for `VerifyImage` to be suitable for the `VerifyImageApplyPath` method. Also add a new test for the `VerifyImage`.

tkireva: @vlantonov 
Done in c059cdbfbce6f616ba368327f042001f619b29eb

src/components/application_manager/src/message_helper/message_helper.cc
tkireva: @vlantonov 
Done in c059cdbfbce6f616ba368327f042001f619b29eb

src/components/application_manager/src/message_helper/message_helper.cc
tkireva: @vlantonov 
Done in c059cdbfbce6f616ba368327f042001f619b29eb

tkireva:
Rebuild required

PR #1328, opened 2017-02-27
Fix policy table loading with functional_group RPCs greater than 50

istoilovgithub:
SDL rejects to load policy table when sdl_preloaded_pt.json contains at least one functional group with more than 50 RPCs. The fix changes number of allowed RPC's to unlimited (65535).


istoilovgithub:
@okozlovlux @pold500 Please review

Issue #1228, opened 2017-01-24
SDL is using outdated Mobile_API.xml


KhrystynaDubovyk:
## Occurrence:
N/A

## Preconditions:
N/A

## Steps to reproduce
1. Check MOBILE_API and HMI_API in repo SDL built was.
--
 Or run scripts
[018_ATF_P_TC_Sending_PTS_to_mobile_application.lua](https://github.com/smartdevicelink/sdl_atf_test_scripts/blob/develop/test_scripts/Polices/Policy_Table_Update/018_ATF_P_TC_Sending_PTS_to_mobile_application.lua) 
[022_ATF_P_TC_Got_PTU_from_mobile_application.lua](https://github.com/smartdevicelink/sdl_atf_test_scripts/blob/develop/test_scripts/Polices/Policy_Table_Update/022_ATF_P_TC_Got_PTU_from_mobile_application.lua) 
[024_ATF_P_TC_PTU_validation_rules.lua](https://github.com/smartdevicelink/sdl_atf_test_scripts/blob/develop/test_scripts/Polices/Policy_Table_Update/024_ATF_P_TC_PTU_validation_rules.lua)  
[025_ATF_P_TC_PTU_validation_failure.lua](https://github.com/smartdevicelink/sdl_atf_test_scripts/blob/develop/test_scripts/Polices/Policy_Table_Update/025_ATF_P_TC_PTU_validation_failure.lua)  
[053_P_ATF_Register_NewApp_not_exist_inLocalPT_FinishPTU.lua](https://github.com/smartdevicelink/sdl_atf_test_scripts/blob/develop/test_scripts/Polices/Policy_Table_Update/053_P_ATF_Register_NewApp_not_exist_inLocalPT_FinishPTU.lua) 
[054_P_ATF_Register_NewApp_not_exist_inLocalPT_start_PTU_for_NewApp.lua](https://github.com/smartdevicelink/sdl_atf_test_scripts/blob/develop/test_scripts/Polices/Policy_Table_Update/054_P_ATF_Register_NewApp_not_exist_inLocalPT_start_PTU_for_NewApp.lua)  
[032_ATF_P_TC_PTU_Restarting_Policy_Table_Exchange_HTTP.lua](https://github.com/smartdevicelink/sdl_atf_test_scripts/commit/c8060196d50e8178b10bee291dac9b36be0d1e12#diff-9e7b84bc08913ddda5b4b90cc90967ca) 

## Actual result:
[...]

PR #1336, opened 2017-02-28
Make OnSystemRequest.url param not array

LuxoftAKutsan:
According requirements OnSystemRequest.url should not be an array
Related to #1228


LuxoftAKutsan:
For commit 05ab3a5b5d0e0857eb1bd8b7814ba7bdc0ff2a68 was not created any PR to release branch, so it is in scope of this RP

PR #1338, opened 2017-02-28
Fixed SDL behavior  after FACTORY_DEFAULTS

VVeremjova:
1. Backup LPT after reset of UserConsent
2.  After SDL receives BC.OnExitAllApplications(MASTER_RESET) or BC.OnExitAllApplications(FACTORY_DEFAULTS) SDL should clear stored persistent data of all registered applications.

### DIFF
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -2394,22 +2394,38 @@ void ApplicationManagerImpl::HeadUnitReset(
       GetPolicyHandler().UnloadPolicyLibrary();
 
       resume_controller().StopSavePersistentDataTimer();
-      file_system::remove_directory_content(
-          get_settings().app_storage_folder());
+      ClearUserStorage();
       break;
     }
     case mobile_api::AppInterfaceUnregisteredReason::FACTORY_DEFAULTS: {
       GetPolicyHandler().ClearUserConsent();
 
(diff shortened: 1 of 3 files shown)


src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -2440,6 +2439,22 @@ void ApplicationManagerImpl::HeadUnitReset(
...
+  const ApplicationSet& accessor = applications().GetData();
+
+  ApplicationSetConstIt it_app_list = accessor.begin();
+  ApplicationSetConstIt it_app_list_end = accessor.end();
+  const std::string storage_folder = get_settings().app_storage_folder();
+  for (; it_app_list != it_app_list_end; ++it_app_list) {
+    const std::string directory_path =
+        storage_folder + "/" + (*it_app_list).get()->folder_name();
+    file_system::RemoveDirectory(directory_path);
+    const std::string icon_path = get_settings().app_icons_folder() + "/" +
```
tkireva: @VVeremjova 
Could you, please create one more variable before entering the loop that contain get_settings().app_icons_folder() + "/" and concatenate it with the application dependent part in the loop body, because currently get_settings().app_icons_folder() is called for each application. You can also concatenate "/" to storage_folder variable.

VVeremjova: @tkireva Please see c7585c9

PR #1350, opened 2017-03-06
Fix missing OnHMIStatus full

tkireva:
Redundant OnHMIStatus with wrong data was sent before
the real one. This fix will be applied for other policy
flags in different commit as a part of investigation of
another problem.


VVeremjova:
Please see #1348

PR #1352, opened 2017-03-07
Fix incorrect validation of entityId and entityType parameters in Preloaded PT

istoilovgithub:
After the fix SDL considers numeric values in range 0..128 as valid


istoilovgithub:
@okozlovlux Please review

PR #1358, opened 2017-03-09
Fix IGNORED instead of UNSUPPORTED_RESOURCE for UnsubscribeButton

LevchenkoS:
According to requirements for `UnsubscribeButton`,
`SDL` must send `UNSUPPORTED_RESOURCE`,
if button is not supported.
##
@pold500, @indlu, @tkireva, please, review.

### DIFF
--- a/src/components/application_manager/src/commands/mobile/unsubscribe_button_request.cc
+++ b/src/components/application_manager/src/commands/mobile/unsubscribe_button_request.cc
@@ -33,6 +33,8 @@
 
 #include "application_manager/commands/mobile/unsubscribe_button_request.h"
 
+#include <algorithm>
+
 #include "application_manager/application_impl.h"
 
 namespace application_manager {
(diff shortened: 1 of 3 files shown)


src/components/application_manager/src/commands/mobile/unsubscribe_button_request.cc
```diff
@@ -41,13 +43,30 @@ namespace commands {
...
+struct FindButtonName {
+  FindButtonName(const mobile_apis::ButtonName::eType target_name)
+      : target_name_(target_name) {}
+
+  const mobile_apis::ButtonName::eType target_name_;
+
+  bool operator()(const smart_objects::SmartObject& buttonSO) const {
+    using namespace mobile_apis;
+    const ButtonName::eType name = static_cast<ButtonName::eType>(
+        buttonSO.getElement(hmi_response::button_name).asInt());
```
tkireva: @LevchenkoS Maybe you mean .asUint() as in line 81?

LevchenkoS: @tkireva, actually, there should be `asInt()`, because `ButtonName` enum has a `INVALID_ENUM` value of which equal to -1

tkireva: @LevchenkoS OK

LevchenkoS:
Rebuild required

PR #1359, opened 2017-03-09
Fix OnHMIStatus sent before RAI response

tkireva:
After fixing defect for missing BC.ActivateApp calling of
method SendResponse() (RegisterAppInterfaceResponse) is moved
after setting of HMILevel which triggers OnHMIStatus. Applied fix
moves SendResponse() back.

### DIFF
--- a/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
+++ b/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
@@ -816,11 +816,11 @@ void RegisterAppInterfaceRequest::SendRegisterAppInterfaceResponseToMobile() {
                                           &(msg_params[strings::app_hmi_type]));
   }
 #endif  // SDL_REMOTE_CONTROL
+  SendResponse(true, result_code, add_info.c_str(), &response_params);
+
   // Default HMI level should be set before any permissions validation, since it
   // relies on HMI level.
   application_manager_.OnApplicationRegistered(application);
-
-  SendResponse(true, result_code, add_info.c_str(), &response_params);
 
   (*notify_upd_manager)();
 


tkireva:
Please review:

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -816,12 +816,12 @@ void RegisterAppInterfaceRequest::SendRegisterAppInterfaceResponseToMobile() {
                                           &(msg_params[strings::app_hmi_type]));
   }
 #endif  // SDL_REMOTE_CONTROL
+  SendResponse(true, result_code, add_info.c_str(), &response_params);
```
LuxoftAKutsan: @tkireva Please check that in RegisterAppInterfaceResponse there is no activities that should be done before OnAppRegistered.
Note comment in bellow.
@okozlovlux  please also participate in review

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -816,12 +816,12 @@ void RegisterAppInterfaceRequest::SendRegisterAppInterfaceResponseToMobile() {
...
+
   // Default HMI level should be set before any permissions validation, since it
   // relies on HMI level.
   application_manager_.OnApplicationRegistered(application);
 
-  SendResponse(true, result_code, add_info.c_str(), &response_params);
```
LuxoftAKutsan: @tkireva also note that **BC.ActivateApp should be called after OnAppRegistered**, please also be ensured in this behavior.

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
okozlovlux: @tkireva , did you make these changes with `REMOTE_CONTROL OFF` ? if yes, so these fix will affect Reverse SDL

tkireva: @okozlovlux REMOTE_CONTROL is ON in my build configuration

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
tkireva: @LuxoftAKutsan BC.ActivateApp is sent by PolicyHandler::UpdateHMILevel(...) which is called by PolicyHandler::OnUpdateHMILevel(...), triggered by the change of the HMILevel in ApplicationManagerImpl::OnApplicationRegistered(...) --> line 823.
OnAppRegistered is sent by RegisterAppInterfaceRequest::SendOnAppRegisteredNotificationToHMI(...) --> line 811.
As a result BC.ActivateApp is sent after OnAppRegistered.

PR #1361, opened 2017-03-10
Prevent stop of SDL caused by HMI responses

dcherniev:
Replaced the command (which was shutting SDL) by warning, because it could be reached if SDL receives certain HMI response and check it by plugin manager.


dcherniev:
Please review:  @LuxoftAKutsan, @indlu, @VVeremjova, @LitvinenkoIra, @istoilovgithub

PR #1368, opened 2017-03-13
Info parameter not returned for GENERIC_ERROR resultcode

pold500:
Info parameter not returned for GENERIC_ERROR resultcode for PerformAudioPassThru when watchdog timeout from HMI expires.

Changed:
* src/components/application_manager/include/application_manager/commands/command_request_impl.h
* src/components/application_manager/src/commands/command_request_impl.cc
  Added 'info' parameter into 'onTimeOut' virtual method
* src/components/application_manager/src/commands/mobile/perform_audio_pass_thru_request.cc
  Added info parameter if request is out of time
Changed virtual method signature to match descendants

@LuxoftAKutsan 
@tkireva 
@okozlovlux

### DIFF
--- a/src/components/application_manager/include/application_manager/commands/command_request_impl.h
+++ b/src/components/application_manager/include/application_manager/commands/command_request_impl.h
@@ -268,6 +268,31 @@ class CommandRequestImpl : public CommandImpl,
    */
   const CommandParametersPermissions& parameters_permissions() const;
 
+  /**
+   * @brief Adds interface to be awaited for by sdl request command
+     @param interface_id interface which SDL expects to response in given time
+  */
+  void StartAwaitForInterface(const HmiInterfaces::InterfaceID& interface_id);
+
+  /**
+   * @brief Gets interface await state.
+     @param interface_id interface which SDL awaits for response in given time
+     @return true if SDL awaits for response from given interface in interface_id
+  */
+  bool GetInterfaceAwaitState(const HmiInterfaces::InterfaceID& interface_id);
+
+  /**
+   * @brief Sets given HMI interface await status to false
+     @param interface_id interface which SDL no longer awaits for response in given time
+  */
+  void EndAwaitForInterface(const HmiInterfaces::InterfaceID& interface_id);
+
+  /**
+   * @brief This set stores all the interfaces which are awaited by SDL to
+            return a response on some request
+  */
+  std::set<HmiInterfaces::InterfaceID> awaiting_response_interfaces_;
+
   RequestState current_state_;
   sync_primitives::Lock state_lock_;
   CommandParametersPermissions parameters_permissions_;
@@ -296,19 +321,19 @@ class CommandRequestImpl : public CommandImpl,
       const hmi_apis::FunctionID::eType& function_id);
 
   /**
-   * @brief Add information for the component of response in case of timeout
-   * @param response Response message, which info should be extended
-   */
-  void AddTimeOutComponentInfoToMessage(
-      smart_objects::SmartObjectSPtr& response) const;
-
-  /**
    * @brief Method transforms AppHMIType to string
    * @param enum AppHMIType app_hmi_type contains enum value
    * @return string of AppHMIType
    */
   std::string AppHMITypeToString(
       const mobile_apis::AppHMIType::eType app_hmi_type) const;
+
+  /**
+   * @brief Adds information to result message "info" param in case of GENERIC_ERROR
+            about component which not responded in time
+   * @param response Response message, which info should be extended
+   */
+  void AddTimeOutComponentInfoToMessage(smart_objects::SmartObject& response) const;
 };
 
 }  // namespace commands
--- a/src/components/application_manager/src/commands/command_request_impl.cc
+++ b/src/components/application_manager/src/commands/command_request_impl.cc
@@ -32,6 +32,7 @@
 
 #include <algorithm>
 #include <string>
+#include <numeric>
 #include "utils/macro.h"
 #include "utils/make_shared.h"
 #include "application_manager/commands/command_request_impl.h"
@@ -105,7 +106,7 @@ const std::string CreateInfoForUnsupportedResult(
 
 bool CheckResultCode(const ResponseInfo& first, const ResponseInfo& second) {
   if (first.is_ok && second.is_unsupported_resource &&
-      HmiInterfaces::STATE_NOT_AVAILABLE != second.interface_state) {
+      (HmiInterfaces::STATE_NOT_AVAILABLE != second.interface_state)) {
     return true;
   }
   return false;
@@ -225,9 +226,7 @@ void CommandRequestImpl::onTimeOut() {
                                             function_id(),
                                             correlation_id(),
                                             mobile_api::Result::GENERIC_ERROR);
-
-  AddTimeOutComponentInfoToMessage(response);
-
+  AddTimeOutComponentInfoToMessage(*response);
   application_manager_.ManageMobileCommand(response, ORIGIN_SDL);
 }
 
@@ -780,6 +779,30 @@ const CommandParametersPermissions& CommandRequestImpl::parameters_permissions()
   return parameters_permissions_;
 }
 
+void CommandRequestImpl::StartAwaitForInterface(
+    const HmiInterfaces::InterfaceID& interface_id) {
+  awaiting_response_interfaces_.insert(interface_id);
+}
+
+bool CommandRequestImpl::GetInterfaceAwaitState(
+    const HmiInterfaces::InterfaceID& interface_id) {
+  std::set<HmiInterfaces::InterfaceID>::const_iterator it =
+      awaiting_response_interfaces_.find(interface_id);
+  return (it != awaiting_response_interfaces_.end());
+}
+
+void CommandRequestImpl::EndAwaitForInterface(
+    const HmiInterfaces::InterfaceID& interface_id) {
+  std::set<HmiInterfaces::InterfaceID>::const_iterator it =
+      awaiting_response_interfaces_.find(interface_id);
+  if (it != awaiting_response_interfaces_.end()) {
+    awaiting_response_interfaces_.erase(it);
+  } else {
+      LOG4CXX_WARN(logger_, "EndAwaitForInterface called on interface \
+                    which was not put into await state: " << interface_id);
+  }
+}
+
 bool CommandRequestImpl::IsResultCodeUnsupported(
     const ResponseInfo& first, const ResponseInfo& second) const {
   return ((first.is_ok || first.is_invalid_enum) &&
@@ -789,75 +812,59 @@ bool CommandRequestImpl::IsResultCodeUnsupported(
          (first.is_unsupported_resource && second.is_unsupported_resource);
 }
 
+std::string GetComponentNameFromInterface(
+    const HmiInterfaces::InterfaceID& interface) {
+  switch (interface) {
+    case HmiInterfaces::HMI_INTERFACE_Buttons:
+      return "Buttons";
+    case HmiInterfaces::HMI_INTERFACE_BasicCommunication:
+      return "BasicCommunication";
+    case HmiInterfaces::HMI_INTERFACE_VR:
+      return "VR";
+    case HmiInterfaces::HMI_INTERFACE_TTS:
+      return "TTS";
+    case HmiInterfaces::HMI_INTERFACE_UI:
+      return "UI";
+    case HmiInterfaces::HMI_INTERFACE_Navigation:
+      return "Navigation";
+    case HmiInterfaces::HMI_INTERFACE_VehicleInfo:
+      return "VehicleInfo";
+    case HmiInterfaces::HMI_INTERFACE_SDL:
+      return "SDL";
+    default:
+      return "Unknown type";
+  }
+}
+
+std::string InfoInterfaceSeparator(const std::string& sum, const HmiInterfaces::InterfaceID& container_value) {
+    return sum.empty() ? GetComponentNameFromInterface(container_value) : sum + ", "
+                          + GetComponentNameFromInterface(container_value);
+}
+
 void CommandRequestImpl::AddTimeOutComponentInfoToMessage(
-    smart_objects::SmartObjectSPtr& response) const {
+    smart_objects::SmartObject& response) const {
   using NsSmartDeviceLink::NsSmartObjects::SmartObject;
   LOG4CXX_AUTO_TRACE(logger_);
 
-  if (!response) {
-    LOG4CXX_WARN(logger_, "Invalid message object");
+  if (awaiting_response_interfaces_.empty()) {
+    LOG4CXX_ERROR(logger_,
+                 "No interfaces awaiting, info param is empty");
     return;
   }
 
-  const uint32_t app_connection_key = connection_key();
-  const ApplicationSharedPtr app =
-      application_manager_.application(app_connection_key);
-  if (!app) {
-    LOG4CXX_WARN(logger_, "Invalid connection_key: " << app_connection_key);
-    return;
-  }
-
-  const SmartObject* app_type = app->app_types();
-  if (!app_type) {
-    LOG4CXX_WARN(logger_, "Empty application types array!");
-    return;
-  }
-
-  if (NsSmartDeviceLink::NsSmartObjects::SmartType_Array !=
-      app_type->getType()) {
-    LOG4CXX_WARN(logger_, "Application types are not an array!");
-    return;
-  }
-
-  const SmartObject& app_type_so = app_type->getElement(0);
-  if (NsSmartDeviceLink::NsSmartObjects::invalid_object_value == app_type_so) {
-    LOG4CXX_WARN(logger_, "Invalid object for component type!");
-    return;
-  }
-
-  const int64_t app_hmi_type = app_type_so.asInt();
-  const std::string& component_name = AppHMITypeToString(
-      static_cast<mobile_apis::AppHMIType::eType>(app_hmi_type));
-  const std::string component_info =
-      component_name + " component does not respond";
-  (*response)[strings::msg_params][strings::info] = component_info;
-}
-
-std::string CommandRequestImpl::AppHMITypeToString(
-    const mobile_apis::AppHMIType::eType app_hmi_type) const {
-  switch (app_hmi_type) {
-    case mobile_apis::AppHMIType::DEFAULT:
-      return "Default";
-    case mobile_apis::AppHMIType::COMMUNICATION:
-      return "Communication";
-    case mobile_apis::AppHMIType::MEDIA:
-      return "Media";
-    case mobile_apis::AppHMIType::MESSAGING:
-      return "Messaging";
-    case mobile_apis::AppHMIType::NAVIGATION:
-      return "Navigation";
-    case mobile_apis::AppHMIType::INFORMATION:
-      return "Information";
-    case mobile_apis::AppHMIType::SOCIAL:
-      return "Social";
-    case mobile_apis::AppHMIType::BACKGROUND_PROCESS:
-      return "Background Process";
-    case mobile_apis::AppHMIType::TESTING:
-      return "Testing";
-    case mobile_apis::AppHMIType::SYSTEM:
-      return "System";
-    default:
-      return "Invalid Component Type";
+  std::string not_responding_interfaces_string = std::accumulate(
+               awaiting_response_interfaces_.begin(),
+               awaiting_response_interfaces_.end(),
+               std::string(""),
+               InfoInterfaceSeparator
+               );
+  LOG4CXX_DEBUG(
+      logger_,
+      "Not responding interfaces string: " << not_responding_interfaces_string);
+  if (!not_responding_interfaces_string.empty()) {
+    const std::string component_info =
+        not_responding_interfaces_string + " component does not respond";
+    response[strings::msg_params][strings::info] = component_info;
   }
 }
 
--- a/src/components/application_manager/src/commands/mobile/perform_audio_pass_thru_request.cc
+++ b/src/components/application_manager/src/commands/mobile/perform_audio_pass_thru_request.cc
@@ -49,8 +49,6 @@ namespace str = strings;
 PerformAudioPassThruRequest::PerformAudioPassThruRequest(
     const MessageSharedPtr& message, ApplicationManager& application_manager)
     : CommandRequestImpl(message, application_manager)
-    , awaiting_tts_speak_response_(false)
-    , awaiting_ui_response_(false)
     , result_tts_speak_(hmi_apis::Common_Result::INVALID_ENUM)
     , result_ui_(hmi_apis::Common_Result::INVALID_ENUM) {
   subscribe_on_event(hmi_apis::FunctionID::TTS_OnResetTimeout);
@@ -98,7 +96,7 @@ void PerformAudioPassThruRequest::Run() {
   // According with new implementation processing of UNSUPPORTE_RESOURCE
   // need set flag before sending to hmi
   ProcessAudioPassThruIcon(app);
-  awaiting_ui_response_ = true;
+  StartAwaitForInterface(HmiInterfaces::HMI_INTERFACE_UI);
   if ((*message_)[str::msg_params].keyExists(str::initial_prompt) &&
       (0 < (*message_)[str::msg_params][str::initial_prompt].length())) {
     // In case TTS Speak, subscribe on notification
@@ -120,7 +118,7 @@ void PerformAudioPassThruRequest::on_event(const event_engine::Event& event) {
   switch (event.id()) {
     case hmi_apis::FunctionID::UI_PerformAudioPassThru: {
       LOG4CXX_TRACE(logger_, "Received UI_PerformAudioPassThru");
-      awaiting_ui_response_ = false;
+      EndAwaitForInterface(HmiInterfaces::HMI_INTERFACE_UI);
 
       result_ui_ = static_cast<hmi_apis::Common_Result::eType>(
           message[strings::params][hmi_response::code].asUInt());
@@ -143,7 +141,7 @@ void PerformAudioPassThruRequest::on_event(const event_engine::Event& event) {
       result_tts_speak_ = static_cast<hmi_apis::Common_Result::eType>(
           message[strings::params][hmi_response::code].asUInt());
       GetInfo(message, tts_info_);
-      awaiting_tts_speak_response_ = false;
+      EndAwaitForInterface(HmiInterfaces::HMI_INTERFACE_TTS);
       const bool is_tts_speak_success_unsuported =
           Compare<hmi_apis::Common_Result::eType, EQ, ONE>(
               result_tts_speak_,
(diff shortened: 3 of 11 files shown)


src/components/application_manager/include/application_manager/commands/command_request_impl.h
```diff
@@ -114,7 +114,7 @@ class CommandRequestImpl : public CommandImpl,
    * has exceed it's limit
    *
    */
-  virtual void onTimeOut();
+  virtual void onTimeOut(const std::string &info = "");
```
LuxoftAKutsan: @pold500 virtual function with default parameters is very very bad. It even pointed out in google coding style

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -224,9 +224,8 @@ void CommandRequestImpl::onTimeOut() {
                                             function_id(),
                                             correlation_id(),
                                             mobile_api::Result::GENERIC_ERROR);
-
+  (*response)[strings::msg_params][strings::info] = info;
```
LuxoftAKutsan: @pold500 with that approach SDL will send empty info parameter for each timed out request

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -224,9 +224,8 @@ void CommandRequestImpl::onTimeOut() {
...
   AddTimeOutComponentInfoToMessage(response);
```
dcherniev: @pold500 Have you checked this method? Purpose of this method is exactly to create the "info" param for onTimeOut response, and currently if this method will run successfully, your received info parameter will be simply replaced inside of this method. So technically either you should update expected info inside of this method, or maybe there is something wrong with the script that it does not pass trough this method for some reason.

src/components/application_manager/src/commands/mobile/perform_audio_pass_thru_request.cc
```diff
@@ -62,7 +62,8 @@ void PerformAudioPassThruRequest::onTimeOut() {
   LOG4CXX_AUTO_TRACE(logger_);
 
   FinishTTSSpeak();
-  CommandRequestImpl::onTimeOut();
+  const std::string info = awaiting_ui_response_ ? "UI component does not respond":"";
```
dcherniev: @pold500 I'm a bit confused, is it possible to get inside of this onTimeOut function having false "awaiting_ui_response_" parameter?

src/components/application_manager/src/commands/command_request_impl.cc
dcherniev: @pold500 Yes, this way the "info" parameter will always be present in this response which may be an error behavior for some cases.

src/components/application_manager/src/commands/command_request_impl.cc
pold500: @dcherniev eventually, it does go there, but there's no "info" field afterwards, so I had to add it manually.

src/components/application_manager/src/commands/mobile/perform_audio_pass_thru_request.cc
pold500: @dcherniev I don't know, is there's a way to check it?

src/components/application_manager/src/commands/command_request_impl.cc
dcherniev: @pold500 Well I guess that means that the test hits 1 of the return false conditions in this method, have you checked what condition was it?

tkireva: @pold500 Recently I was working on similar issue with missing info parameter in message with return code GENERIC_ERROR. It turned out that in this test script one of all messages that are sent to the SDL is "broken" and the reason of returning GENERIC_ERROR is different from the expected in the test script and as a result the info parameter also wasn't with the expected value. So, I advise you to get sure about the reason of returning GENERIC_ERROR in your case, it may be different from the expectations in the script.

src/components/application_manager/src/commands/mobile/perform_audio_pass_thru_request.cc
tkireva: @pold500 @dcherniev 
This function is called in case of UI timed out or TTS timed out. It is common for both interfaces and at this point of view it is possible awaiting_ui_response to be false inside. This may be verified by some test script of this CRQ - for missing UI part and missing TTS part.

pold500:
@dcherniev please review
@LuxoftAKutsan please review
@istoilovgithub please review

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -799,68 +821,31 @@ void CommandRequestImpl::AddTimeOutComponentInfoToMessage(
...
+         m_notRespondingInterfaces.push_back(static_cast<HmiInterfaces::InterfaceID>(interface_id));
+      }
   }
+  std::string not_responding_components;
+  std::for_each(std::begin(m_notRespondingInterfaces), std::end(m_notRespondingInterfaces), [&m_notRespondingInterfaces,
+                &not_responding_components]
+                (const HmiInterfaces::InterfaceID& not_responding_interface) {
+      const auto component_name = GetComponentNameFromInterface(not_responding_interface);
+      not_responding_components+= component_name;
+      if(not_responding_interface != *(std::end(m_notRespondingInterfaces) - 1)) {
```
pold500: I think I've stopped a mistake there. Needs to be fixed.

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -790,6 +788,29 @@ bool CommandRequestImpl::IsResultCodeUnsupported(
...
+    case HmiInterfaces::HMI_INTERFACE_VehicleInfo:
+        return "VehicleInfo";
+    case HmiInterfaces::HMI_INTERFACE_SDL:
+        return "SDL";
+    default:
+        return "Unknown type";
+    }
+}
+
 void CommandRequestImpl::AddTimeOutComponentInfoToMessage(
```
istoilovgithub: @pold500 IMO using request id returned by the member Command::function_id you can find exact HMI type that was timed out. Something like this:

HmiInterfaces& hmi_interfaces = application_manager_.hmi_interfaces();
HmiInterfaces::InterfaceID interface =  hmi_interfaces.GetInterfaceFromFunction(function_id());
const std::string& component_name = GetComponentNameFromInterface(interface);
...

istoilovgithub:
@pold500 One more general comments from my side: Please rephrase your commits to follow guidelines. Squash your commits when possible. Prefer rebasing vs merging to resolve conflicts, it keeps commit history simpler.

pold500:
Please review @LuxoftAKutsan @istoilovgithub

VVeremjova:
@pold500 You wrote in description that there was changed command_request_impl.h, which actually doesn't

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -790,6 +788,29 @@ bool CommandRequestImpl::IsResultCodeUnsupported(
          (first.is_unsupported_resource && second.is_unsupported_resource);
 }
 
+std::string GetComponentNameFromInterface(const HmiInterfaces::InterfaceID& interface) {
```
VVeremjova: @pold500 I have checked locally your code with scripts 027_ATF_P_PerformAudioPassThru_UI_timeout_TTS_success.lua
028_ATF_P_PerformAudioPassThru_TTS_timeout_UI_success.lua
and they don't pass. The error is:
expected: UI component does not respond, actual value: Buttons, BasicCommunication, SDL component does not respond
The same situation with any script for TTS, VR not respond. 
Please fix logic and check locally with scripts for reproducing defect

pold500: yes I know about that but that's not logic which is wrong it's script.

pold500:
@VVeremjova please review
@LuxoftAKutsan please review

src/components/application_manager/test/commands/mobile/alert_request_test.cc
```diff
@@ -235,6 +235,8 @@ TEST_F(AlertRequestTest, OnTimeout_GENERIC_ERROR) {
       (*ui_command_result)[am::strings::msg_params][am::strings::result_code]
           .asInt(),
       static_cast<int32_t>(am::mobile_api::Result::GENERIC_ERROR));
+  EXPECT_EQ((*ui_command_result)[am::strings::msg_params][am::strings::info].asString(),
```
VVeremjova: @pold500 It looks incorrect - according requirements info should not be empty if there is GENERIC_ERROR after timeout expired. Please correct this check

src/components/application_manager/test/commands/command_request_impl_test.cc
```diff
@@ -185,6 +185,7 @@ TEST_F(CommandRequestImplTest, OnTimeOut_StateAwaitingHMIResponse_SUCCESS) {
 
   command->onTimeOut();
 
+  EXPECT_TRUE((*dummy_msg)[am::strings::msg_params][am::strings::info].asString().empty());
```
VVeremjova: @pold500 Please add some comment why here info should be empty or change this check

src/components/application_manager/test/commands/mobile/perform_audio_pass_thru_test.cc
```diff
@@ -346,7 +346,7 @@ TEST_F(PerformAudioPassThruRequestTest, OnTimeout_GENERIC_ERROR) {
   command->onTimeOut();
 
   ResultCommandExpectations(
-      msg_mobile_response, NULL, am::mobile_api::Result::GENERIC_ERROR, false);
+      msg_mobile_response, "", am::mobile_api::Result::GENERIC_ERROR, false);
```
VVeremjova: @pold500 Again, after timeout info shouldn't be empty

src/components/application_manager/test/commands/mobile/perform_interaction_test.cc
```diff
@@ -236,6 +237,7 @@ TEST_F(PerformInteractionRequestTest, OnTimeout_VR_GENERIC_ERROR) {
   EXPECT_EQ(
       (*vr_command_result)[strings::msg_params][strings::result_code].asInt(),
       static_cast<int32_t>(am::mobile_api::Result::GENERIC_ERROR));
+  EXPECT_EQ((*vr_command_result)[strings::msg_params][strings::info].asString(), "");
```
VVeremjova: @pold500 According name of test info should be "VR component does not respond"

src/components/application_manager/test/commands/mobile/set_audio_streaming_indicator_test.cc
```diff
@@ -494,18 +494,14 @@ TEST_F(SetAudioStreamingIndicatorRequestTest,
...
   ApplicationSharedPtr mock_app_empty;
   EXPECT_CALL(app_mngr_, application(_))
-      .Times(2)
+      .Times(1)
       .WillRepeatedly(Return(mock_app_empty));
 
   EXPECT_CALL(*mock_app_, RemoveIndicatorWaitForResponse(_)).Times(0);
   EXPECT_CALL(
       mock_message_helper_,
       CreateNegativeResponse(_, _, _, am::mobile_api::Result::GENERIC_ERROR))
```
VVeremjova: @pold500 It doesn't check info

dcherniev:
@pold500 Your pull request had become quite complicated, so it is hard to review it right now, thus I have some generall questions:
Can you squash your changes to few logical blocks for example "changes related to defect fix" and "changes related to unit tests update"?
Please update your pull request description - it's the very first comment in this pull request, and write what changes in what files were made and what purpose each change has in scope of your pull request.
As Ivo has pointed out: please rebase your pull request on latest version of extended_policy_functionality, this will clear your pull request from changes that was not made by you.
If you have changes that are no longer valid and erased in one of your future commits please squash them so they don't uppear in diff history since it is really confusing to read your changes "step by step". This  also applyes to minor fixes like formating code stile fix, it can be squashed in previous commit since there is no point to review changes related to formatting.
Also please be aware that it is a rule to add an answer to each comment that you recieved during the review. It should be a comment with your opinion if you do not aggree, or comment with link to changes that were made according to reviewer request. Do not leave unanswered comments since your reviewer will be unaware that you have completed requested change. Also this will help you remember on what comments you have answered, and what comments still should be checked.
So overall, please update your pull request to make it clear for your reviewers, thank you!

src/components/application_manager/test/commands/mobile/set_audio_streaming_indicator_test.cc
pold500: @VVeremjova fixed in commit https://github.com/smartdevicelink/sdl_core/pull/1368/commits/e1e998ffc82c12f956a5d10777d9061aa13daae2

src/components/application_manager/test/commands/mobile/alert_request_test.cc
pold500: @VVeremjova fixed in https://github.com/smartdevicelink/sdl_core/pull/1368/commits/e1e998ffc82c12f956a5d10777d9061aa13daae2

src/components/application_manager/test/commands/command_request_impl_test.cc
pold500: @VVeremjova fixed in https://github.com/smartdevicelink/sdl_core/pull/1368/commits/e1e998ffc82c12f956a5d10777d9061aa13daae2

src/components/application_manager/test/commands/mobile/perform_audio_pass_thru_test.cc
pold500: @VVeremjova fixed in https://github.com/smartdevicelink/sdl_core/pull/1368/commits/e1e998ffc82c12f956a5d10777d9061aa13daae2

src/components/application_manager/test/commands/mobile/perform_interaction_test.cc
pold500: @VVeremjova fixed in https://github.com/smartdevicelink/sdl_core/pull/1368/commits/e1e998ffc82c12f956a5d10777d9061aa13daae2

src/components/application_manager/test/commands/mobile/set_audio_streaming_indicator_test.cc
pold500: @VVeremjova fixed in https://github.com/smartdevicelink/sdl_core/pull/1368/commits/e1e998ffc82c12f956a5d10777d9061aa13daae2

pold500:
@istoilovgithub squashed commits to form much cleaner and simpler representation

src/components/application_manager/src/commands/command_request_impl.cc
istoilovgithub: @pold500 I am still thinking that this is the right way to fix the issue, did you give it a try?

pold500: @istoilovgithub yeah I tried that, unfortunately, it didn't work.

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -225,13 +225,12 @@ void CommandRequestImpl::onTimeOut() {
...
-
-  AddTimeOutComponentInfoToMessage(response);
-
+  AddTimeOutComponentInfoToMessage(*response, awaiting_response_interfaces_);
   application_manager_.ManageMobileCommand(response, ORIGIN_SDL);
 }
 
-void CommandRequestImpl::on_event(const event_engine::Event& event) {}
+void CommandRequestImpl::on_event(const event_engine::Event& event) {
+}
```
dcherniev: @pold500 I think that this change can be dropped.

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -248,9 +247,11 @@ void CommandRequestImpl::SendResponse(
     current_state_ = kCompleted;
   }
 
-  smart_objects::SmartObjectSPtr result =
-      utils::MakeShared<smart_objects::SmartObject>();
-
+  smart_objects::SmartObjectSPtr result = new smart_objects::SmartObject;
+  if (!result) {
+    LOG4CXX_ERROR(logger_, "Memory allocation failed.");
```
dcherniev: @pold500 Please do not use such allocation checks, it is impossible to reach it by unit tests. Please use 'utils::MakeShared' to allocete your SmartObject instead of 'new' operator.

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -789,75 +811,69 @@ bool CommandRequestImpl::IsResultCodeUnsupported(
...
-    case mobile_apis::AppHMIType::BACKGROUND_PROCESS:
-      return "Background Process";
-    case mobile_apis::AppHMIType::TESTING:
-      return "Testing";
-    case mobile_apis::AppHMIType::SYSTEM:
-      return "System";
-    default:
-      return "Invalid Component Type";
+  std::vector<HmiInterfaces::InterfaceID> not_responding_interfaces_vector;
+  // we copy contents to vector so we can have iterator that will allow us to do
```
dcherniev: @pold500 Please add some spaces here, clarify in your comment what vector is considered here. Is it related to vector above? In that case shouldn't the description be before it?

src/components/application_manager/test/commands/command_request_impl_test.cc
```diff
@@ -164,7 +164,7 @@ TEST_F(CommandRequestImplTest, OnTimeOut_StateCompleted_UNSUCCESS) {
   EXPECT_EQ(RequestState::kCompleted, command->current_state());
 }
 
-TEST_F(CommandRequestImplTest, OnTimeOut_StateAwaitingHMIResponse_SUCCESS) {
+TEST_F(CommandRequestImplTest, DISABLED_OnTimeOut_StateAwaitingHMIResponse_SUCCESS) {
```
dcherniev: @pold500 You have disabled 6 unit tests, please add a description why each of the tests was disabled by either adding a TODO or creating a common description in commit message.

src/components/application_manager/test/commands/mobile/set_audio_streaming_indicator_test.cc
```diff
@@ -494,18 +500,14 @@ TEST_F(SetAudioStreamingIndicatorRequestTest,
...
       .WillOnce(Return(msg_mobile_response));
-  EXPECT_CALL(
-      app_mngr_,
-      ManageMobileCommand(_, am::commands::Command::CommandOrigin::ORIGIN_SDL))
-      .WillOnce(DoAll(SaveArg<0>(&msg_mobile_response), Return(true)));
 
   command->onTimeOut();
```
dcherniev: @pold500 Title of this commit is too big, please reword it to make it more compact and instead extend the commit description.

Itileda:
Hi @LuxoftAKutsan , @istoilovgithub , @tkireva , @VVeremjova , @dcherniev 

Pull request is updated by Pavel. Please review it ASAP.

tkireva:
@pold500 Have you tested this situation for timed out RPC for all(most of) the RPCs. It is common requirement for all of them to send the name of the component. I ask you this, because I see you use method StartAwaitForInterface() only in PerformAudioPassThruRequest and ReadDIDRequest. I also think that Ivo's suggestion is really appropriate. What was the problem with it? Do you need some help?

pold500:
@tkireva I know this but I won't update update all rpc's in one pull request. If there will be corresponding defects, I suggest assign them to me and I fix that rpc's.
About Ivo's solution,it didn't work. Using function_id did not help.

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -789,75 +808,73 @@ bool CommandRequestImpl::IsResultCodeUnsupported(
...
-      return "Background Process";
-    case mobile_apis::AppHMIType::TESTING:
-      return "Testing";
-    case mobile_apis::AppHMIType::SYSTEM:
-      return "System";
-    default:
-      return "Invalid Component Type";
+  // We copy contents to vector so we can have iterator that will allow us to do
+  // operator-() on it, it will be used in the check later
+  std::vector<HmiInterfaces::InterfaceID> not_responding_interfaces_vector;
```
LuxoftAKutsan: @pold500 print is this way.
http://codepad.org/0G4eB69b

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -780,6 +778,27 @@ const CommandParametersPermissions& CommandRequestImpl::parameters_permissions()
...
+  std::set<HmiInterfaces::InterfaceID>::const_iterator it =
+      awaiting_response_interfaces_.find(interface_id);
+  return (it != awaiting_response_interfaces_.end());
+}
+
+void CommandRequestImpl::EndAwaitForInterface(
+    const HmiInterfaces::InterfaceID& interface_id) {
+  std::set<HmiInterfaces::InterfaceID>::const_iterator it =
+      awaiting_response_interfaces_.find(interface_id);
+  if (it != awaiting_response_interfaces_.end()) {
```
pold500: @LuxoftAKutsan you mean that I should check if StartAwawit was called, right?

src/components/application_manager/src/commands/command_request_impl.cc
istoilovgithub: @pold500 My intention is to find as much generic solution as possible, since the current one is command specific. Anyway, if there is no reliable way to find the HMI interface of the timeout command, I am fine with your changes, but what about the other commands, since the same requirement is also applicable to them?

src/components/application_manager/include/application_manager/commands/command_request_impl.h
```diff
@@ -268,6 +268,30 @@ class CommandRequestImpl : public CommandImpl,
    */
   const CommandParametersPermissions& parameters_permissions() const;
 
+  /**
+   * @brief This set stores all the interfaces which are awaited by SDL to
+            return a response on some request
+  */
+  std::set<HmiInterfaces::InterfaceID> awaiting_response_interfaces_;
```
VVeremjova: @pold500 Please move it below functions' declarations

src/components/application_manager/include/application_manager/commands/command_request_impl.h
```diff
@@ -296,19 +320,22 @@ class CommandRequestImpl : public CommandImpl,
...
    * @brief Method transforms AppHMIType to string
    * @param enum AppHMIType app_hmi_type contains enum value
    * @return string of AppHMIType
    */
   std::string AppHMITypeToString(
       const mobile_apis::AppHMIType::eType app_hmi_type) const;
+
+  /**
+   * @brief Adds information to result message "info" param in case of GENERIC_ERROR
+            about component which not repsonded in time
```
VVeremjova: @pold500 typo in responded

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -789,75 +808,73 @@ bool CommandRequestImpl::IsResultCodeUnsupported(
...
-      app_type->getType()) {
-    LOG4CXX_WARN(logger_, "Application types are not an array!");
-    return;
-  }
-
-  const SmartObject& app_type_so = app_type->getElement(0);
-  if (NsSmartDeviceLink::NsSmartObjects::invalid_object_value == app_type_so) {
-    LOG4CXX_WARN(logger_, "Invalid object for component type!");
+  if (not_responding_interfaces.empty()) {
+    LOG4CXX_DEBUG(logger_,
```
VVeremjova: @pold500 Probably it should be Warn

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -789,75 +808,73 @@ bool CommandRequestImpl::IsResultCodeUnsupported(
...
+    const std::string component_name =
+        GetComponentNameFromInterface(not_responding_interface);
+    not_responding_interfaces_string += component_name;
+    if (it != (std::end(not_responding_interfaces_vector) - 1)) {
+      not_responding_interfaces_string += ", ";
+    }
+  }
+  LOG4CXX_DEBUG(
+      logger_,
+      "not_responding_interfaces_string: " << not_responding_interfaces_string);
```
VVeremjova: @pold500 Change message for more readable: e.g. `"Not respond : "<< not_responding_interfaces_string` or print the same string as component_info

src/components/application_manager/src/commands/command_request_impl.cc
pold500: @LuxoftAKutsan about solution with "accumulate": template type of set is enum, not a string, so I still will have to convert set to set of strings or vector of strings for your solution to work.

LuxoftAKutsan:
```
#include <set>
#include <iostream>
#include <string>
#include <numeric>

enum eType {
  X,
  Y
};

std::string convert(eType t) {
  return (X == t) ?  "X": "Y";
}

std::string InfoInterfaceSeparator(const std::string& curr, const eType& val) {
    return curr.empty() ? convert(val) : curr + ", " + convert(val);
}

int main() {
  std::set<eType> myset;
  myset.insert(X);
  std::string res = std::accumulate(
      myset.begin(),
      myset.end(),
      std::string(""),
      InfoInterfaceSeparator
      );
  std::cout << res << std::endl;
}

~                                                                                                                                                                                                                                             
~                                                                                                                                                                                                                                             
~                                                                                                                                                                                                                                             
~     
```

src/components/application_manager/include/application_manager/commands/command_request_impl.h
pold500: @VVeremjova all fixed in https://github.com/smartdevicelink/sdl_core/pull/1368/commits/6cb301a444f95a33a443f5c0438a182f3b6c9e16

src/components/application_manager/src/commands/command_request_impl.cc
pold500: @LuxoftAKutsan fixed in commit https://github.com/smartdevicelink/sdl_core/pull/1368/commits/6cb301a444f95a33a443f5c0438a182f3b6c9e16

src/components/application_manager/include/application_manager/commands/command_request_impl.h
VVeremjova: @pold500 'repsonded' - > 'responded'

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -105,7 +105,7 @@ const std::string CreateInfoForUnsupportedResult(
 
 bool CheckResultCode(const ResponseInfo& first, const ResponseInfo& second) {
   if (first.is_ok && second.is_unsupported_resource &&
-      HmiInterfaces::STATE_NOT_AVAILABLE != second.interface_state) {
+      (HmiInterfaces::STATE_AVAILABLE != second.interface_state)) {
```
VVeremjova: @pold500 Why did you change this variable?

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -789,75 +811,59 @@ bool CommandRequestImpl::IsResultCodeUnsupported(
...
-      return "Social";
-    case mobile_apis::AppHMIType::BACKGROUND_PROCESS:
-      return "Background Process";
-    case mobile_apis::AppHMIType::TESTING:
-      return "Testing";
-    case mobile_apis::AppHMIType::SYSTEM:
-      return "System";
-    default:
-      return "Invalid Component Type";
+  std::string not_responding_interfaces_string = std::accumulate(
```
VVeremjova: @pold500 please add #include <numeric>

pold500: @VVeremjova fixed in https://github.com/smartdevicelink/sdl_core/pull/1368/commits/ab69afffe097fc66b49d474f1e18aad87f4aa6dc

PR #1375, opened 2017-03-15
Correct param values in lower or upper bound cases

dcherniev:
Corrected values for vrSynonyms and ttsName, that are sent in case when we should get their default values from lower or upper bound in json file.

### DIFF
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -78,6 +78,39 @@ namespace {
 int get_rand_from_range(uint32_t from = 0, int to = RAND_MAX) {
   return std::rand() % to + from;
 }
+
+/**
+ * @brief Get parameter data from application's specific language table
+ * @param app_data is SO that contains application data
+ * @param idx is index of array in language table
+ * @param lang_name contains name of language
+ * @param name_so_param is name of parameter to search in language table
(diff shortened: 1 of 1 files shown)


dcherniev:
Please review: @LuxoftAKutsan, @indlu, @VVeremjova, @LitvinenkoIra, @istoilovgithub

src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -2140,51 +2172,73 @@ void ApplicationManagerImpl::PullLanguagesInfo(const SmartObject& app_data,
...
   const SmartObject& languages = app_data[json::languages];
 
+  std::transform(cur_tts_lang.begin(),
+                 cur_tts_lang.end(),
+                 cur_tts_lang.begin(),
+                 ::toupper);
   std::transform(
       cur_vr_lang.begin(), cur_vr_lang.end(), cur_vr_lang.begin(), ::toupper);
 
+  ssize_t specific_tts_idx = -1;
```
istoilovgithub: @dcherniev It would be better to use std::npos here

dcherniev: @istoilovgithub You right, fixed, please check 18232e7

dcherniev:
Rebuild required

dcherniev:
Rebuild required

PR #1378, opened 2017-03-16
Fix INVALID_DATA returned by SDL on request SetGlobalProperties with valid parameters

istoilovgithub:
The issue is caused by the inability to be prepare default help data, after
the fix SDL hail backs to use application name as the last possible option.

### DIFF
--- a/src/components/application_manager/src/commands/mobile/set_global_properties_request.cc
+++ b/src/components/application_manager/src/commands/mobile/set_global_properties_request.cc
@@ -135,7 +135,7 @@ void SetGlobalPropertiesRequest::Run() {
   }
 
   if (is_vr_help_title_present && is_vr_help_present) {
-    LOG4CXX_DEBUG(logger_, "VRHelp params presents");
+    LOG4CXX_DEBUG(logger_, "VRHelp params are present");
 
     if (!CheckVrHelpItemsOrder(msg_params[strings::vr_help])) {
       LOG4CXX_ERROR(logger_,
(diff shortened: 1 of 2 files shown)


istoilovgithub:
@okozlovlux Please review

src/components/application_manager/src/commands/mobile/set_global_properties_request.cc
```diff
@@ -135,7 +135,7 @@ void SetGlobalPropertiesRequest::Run() {
   }
 
   if (is_vr_help_title_present && is_vr_help_present) {
-    LOG4CXX_DEBUG(logger_, "VRHelp params presents");
+    LOG4CXX_DEBUG(logger_, "VRHelp params are present");
```
LuxoftAKutsan: @istoilovgithub please try to avoid handling logic and grammar fixed in one commit

src/components/application_manager/test/commands/mobile/set_global_properties_test.cc
```diff
@@ -90,6 +94,8 @@ class SetGlobalPropertiesRequestTest
   }
 
   ~SetGlobalPropertiesRequestTest() {
+    // Required for proper DataAccessor release
+    Mock::VerifyAndClearExpectations(&app_mngr_);
```
LuxoftAKutsan: @istoilovgithub is app_manager is static? I am not sure that it should be VerifyAndClearExpectations for it? 
@LevchenkoS please clarify this issue

src/components/application_manager/src/commands/mobile/set_global_properties_request.cc
istoilovgithub: @LuxoftAKutsan Honestly forgot for this rule, splitted in separate commit 6ccb49b

src/components/application_manager/test/commands/mobile/set_global_properties_test.cc
istoilovgithub: @LuxoftAKutsan Let me try to put some light on this one since it has a slightly different meaning in this case: this call prevents potential crash while destructing owned lock (see [set_global_properties_test.cc](https://github.com/smartdevicelink/sdl_core/pull/1378/files#diff-33ed62bff3873bac3761ec6756e492fe):474) by forcing mock expectations cleanup at the earliest stage.

PR #1379, opened 2017-03-16
Fix external_consent param out of limit

tkireva:
disallowed_by_external_consent_entities_on / _off parameter
array was defined with size 255. It is changed to 100.


tkireva:
Please review:

PR #1380, opened 2017-03-16
Fix PT array size validation of "default" section in "endpoints"

istoilovgithub:
After the fix PM allows 1 to 3 array elements in "endpoints"
for PROPRIETARY and HTTP flows


istoilovgithub:
@okozlovlux Please review

Issue #1002, opened 2016-11-07
SDL does not send StopAudioStream() if exit app while Video service and Audio service are starting.


DrachenkoAnastasiia:
**Occurrence:**
Always

**Preconditions:**
1.SDL is started
2.HMI is started

**Steps to reproduce:**
1. Register Navi App
2. Activate App
3. Press "Video Service - Show" button
4. Press "Start Service"
5. Press "Ok" button on popup
6. Press "Audio Service - Show" button
7. Press "Start Service" popup
8. Press "Ok" button on popup
9. Press Exit App on menu app

**Actual result:**
SDL only sends StopStream, not send StopAudioStream.

**Expected result:**
StopAudioStream should be sent.

**Environment:**
SDL repo: https://github.com/smartdevicelink/sdl_core.git
SDL branch: master
SDL commit: a6c32b4
HMI: https://github.com/LuxoftSDL/sdl_hmi: 01cb5b49883ed6288e34fcf7db24fd0439d08312
OS: Ubuntu 14.04 LTS

**Attachments:**
[att_issue#1002.zip](https://github.com/smartdevicelink/sdl_core/files/576142/att_issue.1002.zip)

LuxoftAKutsan:
Looks like this issue was fixed with #1638

PR #1381, opened 2017-03-17
Fix SDL does not send StopAudioStream to exit app if Video and Audio services are starting

indlu:
Fixed SDL does not send StopAudioStream to exit app if Video and Audio services are starting

 Related to [Issue-1002](https://github.com/smartdevicelink/sdl_core/issues/1002)


indlu:
@LuxoftAKutsan, @LevchenkoS, @okozlovlux please, review.

PR #1382, opened 2017-03-18
Fix incorrect behavior of SDL when OnPermissionsChange is sent

okozlovlux:
Fix when OnPermissionsChange is sent with incorrect data  in case of SDL functionality is disallowed


okozlovlux:
@LuxoftAKutsan @pold500 @indlu @istoilovgithub please review

PR #1383, opened 2017-03-20
[WIP] Fix case sensitive app name and id

tkireva:
SDL was not case sensitive according to application name and id.
Fix for matchers that compare names and ids and changing of usage
of strcasecmp to strncmp.

### DIFF
--- a/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
+++ b/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
@@ -225,13 +225,11 @@ struct IsSameAppId : public IsSameApp {
     if (AreBothRemoteControl(other) && !IsSameDeviceType(other)) {
       return false;
     }
-    return strncmp(app_id_.c_str(),
-                   other->policy_app_id().c_str(),
-                   app_id_.length()) == 0;
+    return app_id_ == other->policy_app_id();
   }
 
  private:
-  const std::string& app_id_;
+  const std::string app_id_;
 };
 
 struct IsSameAppName : public IsSameApp {
@@ -249,8 +247,7 @@ struct IsSameAppName : public IsSameApp {
       return false;
     }
 
-    bool same =
-        strncmp(name_.c_str(), other->name().c_str(), name_.length()) == 0;
+    bool same = (name_ == other->name().AsMBString());
     if (same) {
       LOG4CXX_AUTO_TRACE(logger_);
       LOG4CXX_ERROR(logger_, "Application name is known already.");
@@ -283,7 +280,7 @@ struct IsSameAppName : public IsSameApp {
   }
 
  private:
-  const std::string& name_;
+  const std::string name_;
   CoincidencePredicateVR matcher_;
   const smart_objects::SmartArray* vr_synonyms_;
 };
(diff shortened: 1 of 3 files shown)


tkireva:
Please review:

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -1191,7 +1199,8 @@ bool RegisterAppInterfaceRequest::IsApplicationWithSameAppIdRegistered() {
       application_manager_.applications().GetData();
 
 #ifdef SDL_REMOTE_CONTROL
-  IsSameAppId matcher(mobile_app_id.AsMBString(),
+  const std::string tmp_mobile_app_id = mobile_app_id.AsMBString();
+  IsSameAppId matcher(tmp_mobile_app_id,
```
okozlovlux: @tkireva , didn't understand.  Why did you create `tmp_mobile_app_id` ?

tkireva: @okozlovlux IsSameAppId holds the app_id as a string reference, but the time of life of the object that AsMBString() returns is not enough, so I create new object tmp_mobile_app_id that will live longer enough. Another approach to fix this problem is changing the IsSameAppId to hold the app_id not as a string reference, but as a string object.

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -225,7 +225,9 @@ struct IsSameAppId : public IsSameApp {
     if (AreBothRemoteControl(other) && !IsSameDeviceType(other)) {
       return false;
     }
-    return strcasecmp(app_id_.c_str(), other->policy_app_id().c_str()) == 0;
+    return strncmp(app_id_.c_str(),
```
istoilovgithub: @tkireva If we are looking for simple lexicographical compare, than std::basic_string::operator== does it, but looks much better in terms of C++, right?

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
istoilovgithub: @tkireva IMO it wold be better to fix matcher instead of making workarounds.

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -246,7 +248,9 @@ struct IsSameAppName : public IsSameApp {
     if (AreBothRemoteControl(other) && !IsSameDeviceType(other)) {
       return false;
     }
-    bool same = strcasecmp(name_.c_str(), other->name().c_str()) == 0;
+
+    bool same =
```
LuxoftAKutsan: @tkireva please make it const, ans use c++ style o string comparator

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -1202,7 +1211,9 @@ bool RegisterAppInterfaceRequest::IsApplicationWithSameAppIdRegistered() {
   ApplicationSetConstIt it = applications.begin();
   ApplicationSetConstIt it_end = applications.end();
   for (; it != it_end; ++it) {
-    if (mobile_app_id.CompareIgnoreCase((*it)->policy_app_id().c_str())) {
```
LuxoftAKutsan: @tkireva please Add additional function to CustonString - "Compare" (case sensitive one)

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
tkireva: @istoilovgithub Done in da077f5e60082acc498089365fb2b794989a001e

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
tkireva: @LuxoftAKutsan Done in 8eab697986f84199ced656a8bb6e066ae6f12e88

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
tkireva: @LuxoftAKutsan Done in 8eab697986f84199ced656a8bb6e066ae6f12e88

tkireva:
Applying [WIP] flag is approved by Igor Kovalenko

PR #1394, opened 2017-03-20
Fixed defect "no external_consent_status_groups section"

pold500:
This pull request fixes a defect about that there's no external_consent_status_groups section 
in policy table in case SDL gets SDL.OnAppPermissionConsent from HMI with externalConsentStatus change from "ON" to "OFF".

Changed files:
* src/components/application_manager/src/smart_object_keys.cc
   fixed name of the key 'ExtendedConsentStatus' to start from the lowercase letter
* src/components/application_manager/src/commands/hmi/on_app_permission_consent_notification.cc
   removed smart pointer from where it should not belong (which caused crash of the application)


pold500:
@LuxoftAKutsan 
@VVeremjova 
@tkireva 
please review

Issue #1010, opened 2016-11-07
[VS][AS] SDL does not send StopStream/StopAudioStream to HMI after unregistering app during streaming


DrachenkoAnastasiia:
## Occurrence
Always

## Preconditions
1. SDL Core and HMI are started.
2. Navi app is registered and activated on HMI.

## Steps to reproduce
1. Start VIdeo/Audio Service.
2. Start File Streaming.
3. Send UnregisterAppInterface while file is streaming.

## Actual result
Core doesn't send StopStream/StopAudioStream to HMI.

## Expected result
SDL sends StopStream to HMI and OnAppUnregistered. 

_**Note:**_
    The same behavior if turn off transport or mobile device.
    The same behavior if exit SPT from mobile (wifi + pipe/socket).

# Environment
SDL repo: https://github.com/smartdevicelink/sdl_core.git
SDL branch: master
SDL commit: a6c32b4
HMI: https://github.com/LuxoftSDL/sdl_hmi: 01cb5b49883ed6288e34fcf7db24fd0439d08312
OS: Ubuntu 14.04 LTS

# Attachments
[att_issue#1010.zip](https://github.com/smartdevicelink/sdl_core/files/576198/att_issue.1010.zip).

# Expected delivery   
- [ ] Source code updates 
- [ ] Code comments 
- [ ] UTs add/update 
- [x] ATF tests add/update ***(not required)***  
- [x] Manual tests ***(not required)*** 
- [x] Add/update CI plans/jobs ***(not required)***  
- [x] SDD updates ***(not required)***  
- [x] Guidelines update ([sdl_core_guides](https://github.com/smartdevicelink/sdl_core_guides)) ***(not required)***  
- [x] Guidelines update ([sdl_hmi_integration_guidelines](https://github.com/smartdevicelink/sdl_hmi_integration_guidelines))  ***(not required)***

t-yoshii:
By https://github.com/smartdevicelink/sdl_core/pull/1407 fix, EndNaviServices() is not called when application is unregistered not by UnregisterAppInterface RPC, such as app is killed or connection is lost. In such cases, ApplicationManager::UnregisterApplication() is called at https://github.com/smartdevicelink/sdl_core/blob/master/src/components/application_manager/src/application_manager_impl.cc#L1251 without calling EndNaviServices().

[...]

PR #1407, opened 2017-03-21
Fix SDL does not send StopAudio/VideoStream after unregistering app

indlu:
Fixed SDL does not send StopStream/StopAudioStream to HMI after unregistering app during streaming

 Related to [Issue-1010](https://github.com/smartdevicelink/sdl_core/issues/1010)


indlu:
@LuxoftAKutsan, @LevchenkoS, @okozlovlux please, review.

PR #1412, opened 2017-03-22
Fix not triggered PTU for NAVI app

tkireva:
When a NAVI application that already persists in the PT
and is registered again, no PTU was trigered. The reason
was in missing check for HMI_TYPE in the right moment.
When this feature is moved from FS, this piece of code is
skipped.

### DIFF
--- a/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
+++ b/src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
@@ -816,6 +816,7 @@ void RegisterAppInterfaceRequest::SendRegisterAppInterfaceResponseToMobile() {
   if ((*message_)[strings::msg_params].keyExists(strings::app_hmi_type)) {
     smart_objects::SmartArray* hmi_types_ptr =
         (*message_)[strings::msg_params][strings::app_hmi_type].asArray();
+    DCHECK_OR_RETURN_VOID(hmi_types_ptr);
     SmartArrayValueExtractor extractor;
     if (hmi_types_ptr && 0 < hmi_types_ptr->size()) {
       std::transform(hmi_types_ptr->begin(),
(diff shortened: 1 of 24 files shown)


tkireva:
Please review:

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -805,9 +812,21 @@ void RegisterAppInterfaceRequest::SendRegisterAppInterfaceResponseToMobile() {
     resumption = resumer.IsApplicationSaved(application->policy_app_id(),
                                             application->mac_address());
   }
+  AppHmiTypes hmi_types;
+  if ((*message_)[strings::msg_params].keyExists(strings::app_hmi_type)) {
+    smart_objects::SmartArray* hmi_types_ptr =
```
okozlovlux: @tkireva check this pointer with DCHECK (app_hmi_type should be array, if it is not - our API parser works not correctly)

tkireva: @okozlovlux Done in 498034cba39a0671642a79409e411003c795b5ba

okozlovlux: @tkireva , I meant something like this :
`DCHECK_OR_RETURN_VOID( (*message_)[strings::msg_params][strings:app_hmi_type].getType() ==
            smart_objects::SmartType_Array);
smart_objects::SmartArray* hmi_types_ptr =
          (*message_)[strings::msg_params][strings::app_hmi_type].asArray();`

tkireva: @okozlovlux I think that if app_hmi_type is not array, this will be detected on schema JSON validation and INVALID_DATA will be returned. Am I wrong?

okozlovlux: @tkireva , i am not sure.
@LuxoftAKutsan , please review

LuxoftAKutsan: @tkireva you are right

Issue #1365, opened 2017-03-10
[Protocol] SDL respond ACK with protocol version 4 for video and audio services if send start service with 2 or 3 protocol version


AStasiuk:
## Occurrence: 
Sometimes (1/5)
## Precondition: 
1. SDL and HMI are started
2. SPT is registered using v2 or v3 protocol
3. Video and Audio services are allowed to be unsecure
4. Wireshark is configured to listen SDL protocol
## Steps:
1. Start VIdeoService.
2. Accept with pop-up on HMI.
3. Start Audio service.
4. Accept with pop-up on HMI.
5. Start RPC service.
## Actual result: 
SDL respond ACK with protocol version 4.
## Expected result: 
SDL should respond ACK with the same protocol as start service was come - version 2.
## Note:
The same behavior if start secure audio/video services.
## Expected delivery

Source code updates
Code comments
UTs add/update
ATF tests add/update (not required)
Manual tests (not required)
Add/update CI plans/jobs (not required)
SDD updates (not required)
Guidelines update (sdl_core_guides) (not required)
Guidelines update (sdl_hmi_integration_guidelines) (not required)

## Attachments
[wireshark.pcapng.txt](https://github.com/smartdevicelink/sdl_core/files/834474/wireshark.pcapng.txt)
[SDL.txt](https://github.com/smartdevicelink/sdl_core/files/834473/SDL.txt)
[HMI.txt](https://github.com/smartdevicelink/sdl_core/files/834475/HMI.txt)

JackLivio:
Duplicate issue https://github.com/smartdevicelink/sdl_core/issues/1034, leaving this one open

jacobkeeler:
Closing with the merge of #1630

PR #1413, opened 2017-03-23
Fix incorrect behavior of SDL during response ACK 

okozlovlux:
Fix incorrect behavior of SDL during response ACK when video and audio services start

  - added condition in SendStartSessionAck to avoid sending max supported protocol version
    when video or audio services start.
    SDL must send negotiated protocol_version from message from mobile app

Related to [Issue-1365](https://github.com/smartdevicelink/sdl_core/issues/1365)


okozlovlux:
@LuxoftAKutsan @pold500 @istoilovgithub @indlu please review

JackLivio:
@LuxoftAKutsan Does this fix #1365

PR #1416, opened 2017-03-24
Fix missing PTU for encrypted service

tkireva:
SDL didn't trigger PTU when a secured service is started and the
certificate is missing because the member variable that contains
the certificate expiration time was not inited and when it checked
it the time was in the future. As fix this variable is inited in
the constructor.


tkireva:
Please review:

PR #1420, opened 2017-03-27
Fix missing PTU by exchg_after_x_days trigger

tkireva:
Wrong fix to old defect related to redundant PTU is removed.
As a fix first pt_exchanged_x_days_after_epoch is checked if
it is initialized.

### DIFF
--- a/src/components/policy/policy_external/src/cache_manager.cc
+++ b/src/components/policy/policy_external/src/cache_manager.cc
@@ -1218,7 +1218,7 @@ bool CacheManager::SetCountersPassedForSuccessfulUpdate(
   return true;
 }
 
-int CacheManager::DaysBeforeExchange(uint16_t current) {
+ReturnValue CacheManager::DaysBeforeExchange(uint16_t current) {
   LOG4CXX_AUTO_TRACE(logger_);
   CACHE_MANAGER_CHECK(0);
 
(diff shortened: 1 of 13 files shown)


tkireva:
Please review:

src/components/policy/policy_regular/src/cache_manager.cc
```diff
@@ -564,16 +564,22 @@ bool CacheManager::SetCountersPassedForSuccessfulUpdate(
 int CacheManager::DaysBeforeExchange(int current) {
   LOG4CXX_AUTO_TRACE(logger_);
   CACHE_MANAGER_CHECK(0);
+
+  const rpc::Optional<rpc::Integer<uint16_t, 0, 65535> > days_after_epoch =
+      (pt_->policy_table.module_meta->pt_exchanged_x_days_after_epoch);
+
```
indlu: @tkireva, you could use Optional from "utils/optional.h" ,
something like optional::Optional<uint32_t> days_after_epoch

src/components/policy/policy_external/src/cache_manager.cc
```diff
@@ -1221,16 +1221,22 @@ bool CacheManager::SetCountersPassedForSuccessfulUpdate(
 int CacheManager::DaysBeforeExchange(uint16_t current) {
```
LuxoftAKutsan: @tkireva if you use enum in return. Please use it in sugnature

src/components/policy/policy_regular/src/cache_manager.cc
LuxoftAKutsan: @indlu , @tkireva should use the same type as in PT. So it is ok that is is Optional from POlicies.

src/components/policy/policy_external/src/cache_manager.cc
tkireva: @LuxoftAKutsan Done in 10ed3ebecd48fbe9dccfb60520d2ebeef411b54b

tkireva:
Rebuild required

tkireva:
Rebuild required

PR #1422, opened 2017-03-27
Fix validation of items in "groups" sub-section from "default"

istoilovgithub:
Fix changes validation of the PT of any type (Snapshot, Preloaded, Updated),
so it is considered as valid if number of elements in the array value of
"groups" sub-section in "default" section are in range 1..<functional_groupings>


istoilovgithub:
@okozlovlux Please review

PR #1424, opened 2017-03-29
SDL sends BC.UpdateDeviceList with out of upper bound size of deviceList

LevchenkoS:
According to requirements, `SDL` should send
info about 100 devices at maximum, even if
`SDL` has more devices connected.
##
@pold500, @indlu, @tkireva, please, review.


pold500:
@LevchenkoS all builds have failed. Have you checked the results? What about unit tests?

LevchenkoS:
Rebuild required

PR #1430, opened 2017-03-30
Fix after successful PTU last status of SDL.OnStatusUpdate is UPDATE_NEEDED

indlu:
Added timeout taken from "timeout_after_x_seconds" filed of LocalPT according to requirements.


indlu:
Closed due to wrong fix

PR #1431, opened 2017-03-30
Fixed copying of icon to icon folder

VVeremjova:
SDL must copy icon from AppStorageFolder to AppIconsFolder ONLY in case of successfull response from HMI

### DIFF
--- a/src/components/application_manager/src/commands/mobile/set_app_icon_request.cc
+++ b/src/components/application_manager/src/commands/mobile/set_app_icon_request.cc
@@ -83,10 +83,6 @@ void SetAppIconRequest::Run() {
     return;
   }
 
-  if (is_icons_saving_enabled_) {
-    CopyToIconStorage(full_file_path);
-  }
-
   smart_objects::SmartObject msg_params =
       smart_objects::SmartObject(smart_objects::SmartType_Map);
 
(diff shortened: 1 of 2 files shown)


VVeremjova:
@LevchenkoS, @pold500 please review

src/components/application_manager/src/commands/mobile/set_app_icon_request.cc
```diff
@@ -253,6 +249,10 @@ void SetAppIconRequest::on_event(const event_engine::Event& event) {
         const std::string& path =
             (*message_)[strings::msg_params][strings::sync_file_name]
                        [strings::value].asString();
+        if (is_icons_saving_enabled_) {
```
tkireva: @VVeremjova The result checked above in line 240 is the result that will be sent to the mobile. Maybe you should check the content of response_info variable

VVeremjova: @tkireva  is_icons_saving_enabled_ is true if it is possible to read and write in app_icons_folder. Only if result is true we can save icon in folder. It doesn't matter which info has response - according requirement "SDL must copy <syncFileName> from "AppStorageFolder" to "AppIconsFolder" ONLY in case of successfull response from HMI"

tkireva: @VVeremjova OK, I understand for this RPC that has only one HMI interface in this case result_code and result will have same values. I agree

VVeremjova:
Rebuild required

PR #1432, opened 2017-03-30
Fix issue when SDL sends NACK instead of PTU trigger

okozlovlux:
- removed redundant notification about PTU
  - fixed issue when SDL sends NACK instead of PTU trigger

### DIFF
--- a/src/components/application_manager/src/policies/policy_handler.cc
+++ b/src/components/application_manager/src/policies/policy_handler.cc
@@ -486,9 +486,9 @@ void PolicyHandler::SendOnAppPermissionsChanged(
 }
 
 void PolicyHandler::OnPTExchangeNeeded() {
+  LOG4CXX_AUTO_TRACE(logger_);
   POLICY_LIB_CHECK_VOID();
-  MessageHelper::SendOnStatusUpdate(policy_manager_->ForcePTExchange(),
-                                    application_manager_);
+  policy_manager_->ForcePTExchange();
 }
 
 void PolicyHandler::GetAvailableApps(std::queue<std::string>& apps) {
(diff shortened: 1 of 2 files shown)


src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -489,9 +489,9 @@ void PolicyHandler::SendOnAppPermissionsChanged(
 }
 
 void PolicyHandler::OnPTExchangeNeeded() {
+  LOG4CXX_AUTO_TRACE(logger_);
   POLICY_LIB_CHECK_VOID();
-  MessageHelper::SendOnStatusUpdate(policy_manager_->ForcePTExchange(),
-                                    application_manager_);
+  policy_manager_->ForcePTExchange();
```
istoilovgithub: @okozlovlux It would be interesting for me to know why did you decide to suppress status update?

src/components/connection_handler/src/connection_handler_impl.cc
```diff
@@ -297,8 +297,9 @@ uint32_t ConnectionHandlerImpl::OnSessionStartedCallback(
   if (!AllowProtection(get_settings(), service_type, is_protected)) {
     return 0;
   }
+  const uint32_t connection_key = KeyFromPair(connection_handle, session_id);
```
istoilovgithub: @okozlovlux Just to let you know that [PR](https://github.com/smartdevicelink/sdl_core/pull/1433) has already contained a similar fix.

src/components/application_manager/src/policies/policy_handler.cc
okozlovlux: @istoilovgithub , because notification about PTU triggered twice

istoilovgithub: @okozlovlux I am perfectly fine with your explanation, but I am concerned about potential functional regressions, so I would like to be sure that you are aware of the impact of this change.

okozlovlux: @istoilovgithub , agree with you. currently looking if possible regression or find other to fix it

okozlovlux:
@tkireva @indlu please review

PR #1436, opened 2017-03-31
Correct deletion of oldest app icon

dcherniev:
Corrected value of file modification time that was recieved from file system.
Previously file system sent only "nanoseconds in one second" value which was an incorrect behaviour.
Provided new way to compare more precisely the file modification time for two files, which includes both values to compare time:
nanoseconds in one second and seconds that represent time since start of epoch.
Updated the UT GetFileModificationTime for file system class.

### DIFF
--- a/src/components/application_manager/src/commands/mobile/set_app_icon_request.cc
+++ b/src/components/application_manager/src/commands/mobile/set_app_icon_request.cc
@@ -190,7 +190,8 @@ void SetAppIconRequest::CopyToIconStorage(
 void SetAppIconRequest::RemoveOldestIcons(const std::string& storage,
                                           const uint32_t icons_amount) const {
   const std::vector<std::string> icons_list = file_system::ListFiles(storage);
-  std::map<uint64_t, std::string> icon_modification_time;
+  std::set<file_system::FileDescriptor, file_system::FileDescriptor>
+      info_about_icons;
   std::vector<std::string>::const_iterator it = icons_list.begin();
   for (; it != icons_list.end(); ++it) {
     const std::string file_name = *it;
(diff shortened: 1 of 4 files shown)


dcherniev:
Please review: @indlu, @VVeremjova, @LitvinenkoIra, @istoilovgithub, @AGaliuzov

src/components/utils/include/utils/file_system.h
```diff
@@ -241,11 +241,22 @@ const std::string ConvertPathForURL(const std::string& path);
...
+struct timespec GetFileModificationTime(const std::string& path);
+
+/**
+ * @brief Compare time stamps for file A and file B
+ * @param file_a contains name and timestamp for file A
+ * @param file_b contains name and timestamp for file B
+ * @return true if file A is older than file B, false otherwise
+ */
+bool CompareFilesByModificationTime(
+    std::pair<std::string, struct timespec> file_a,
```
AGaliuzov: @dcherniev I would rather create the meaningful structure like `FileDescriptor` instead of `pair`
```c++
struct FileDescriptor {
std::string name;
timespec modification_time;
}
```

dcherniev: @AGaliuzov Good point, fixed in 68bae9f.

src/components/application_manager/src/commands/mobile/set_app_icon_request.cc
```diff
@@ -190,31 +190,41 @@ void SetAppIconRequest::CopyToIconStorage(
 void SetAppIconRequest::RemoveOldestIcons(const std::string& storage,
                                           const uint32_t icons_amount) const {
   const std::vector<std::string> icons_list = file_system::ListFiles(storage);
-  std::map<uint64_t, std::string> icon_modification_time;
+  std::vector<file_system::FileDescriptor> info_about_icons;
```
istoilovgithub: @dcherniev What do you think about using std::set, it keeps elements ordered and you do not need to sort them later.

dcherniev: @istoilovgithub Well the main reason I use std::sort for this std::vector is that I need to use custom compare function for this std::sort (to compare time stamps), is it possible to use such function with std::set?

istoilovgithub: @dcherniev Of course you can, but you have to change slightly your code and wrap already implemented function in a functor.

dcherniev: @istoilovgithub Sorry, I am not familiar with functors, could you provide me small example on how should I do that?

dcherniev: @istoilovgithub Do you mean I should override "<" operator for FileDescriptor struct?

dcherniev: @istoilovgithub Fixed, please take a look at changes in 3231391 and then in 895b1da.

PR #1437, opened 2017-03-31
Fix missing PTU for missing certificate

tkireva:


tkireva:
Please review:
@okozlovlux @LitvinenkoIra

src/components/security_manager/src/crypto_manager_impl.cc
```diff
@@ -234,7 +234,7 @@ bool CryptoManagerImpl::OnCertificateUpdated(const std::string& data) {
 }
 
 SSLContext* CryptoManagerImpl::CreateSSLContext() {
-  if (NULL == context_ || IsCertificateUpdateRequired()) {
+  if (NULL == context_) {
```
LevchenkoS: @tkireva, why you remove `IsCertificateUpdateRequired()` from check?

tkireva: @LevchenkoS  @okozlovlux If IsCertificateUpdateRequired() is not removed here, PTU for missing certificate or for expired certificate won't be triggered - protocol handler will send startServiceACK/NACK. Currently I am working on a fix that in this situation will trigger PTU, wait PTU to be finished and after that will sent startServiceACK/NACK.

tkireva: Currently the main problem is that the SDL should wait PTU finished before send any answer.

PR #1439, opened 2017-04-03
Fix SDL crushes when HMI response with wrong type of method

indlu:
Added type check of the received parameter

### DIFF
--- a/src/components/functional_module/src/plugin_manager.cc
+++ b/src/components/functional_module/src/plugin_manager.cc
@@ -175,20 +175,22 @@ ProcessResult PluginManager::ProcessHMIMessage(
   if (application_manager::ProtocolVersion::kHMI == msg->protocol_version()) {
     std::string msg_method;
     // Request or notification from HMI
-    if (value.isMember("method")) {
+    if (value.isMember("method") && value["method"].isString()) {
       msg_method = value["method"].asCString();
       // Response from HMI
-    } else if (value.isMember("result") && value["result"].isMember("method")) {
+    } else if (value.isMember("result") && value["result"].isMember("method") &&
+               value["result"]["method"].isString()) {
       msg_method = value["result"]["method"].asCString();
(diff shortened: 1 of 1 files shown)


indlu:
@istoilovgithub , @LevchenkoS, @okozlovlux, @dcherniev  please, review.

src/components/functional_module/src/plugin_manager.cc
```diff
@@ -175,14 +175,16 @@ ProcessResult PluginManager::ProcessHMIMessage(
...
+               value["result"]["method"].isString()) {
       msg_method = value["result"]["method"].asCString();
       // Error response from HMI
     } else if (value.isMember("error") && value["error"].isMember("data") &&
-               value["error"]["data"].isMember("method")) {
+               value["error"]["data"].isMember("method") &&
+               value["error"]["data"]["method"].isString()) {
       msg_method = value["error"]["data"]["method"].asCString();
     } else {
       LOG4CXX_WARN(logger_,
```
dcherniev: @indlu I would suggest to extend text in this warning this way:  "... 'method' field was not found, or was containing an invalid string."

src/components/functional_module/src/plugin_manager.cc
```diff
@@ -232,14 +234,16 @@ bool PluginManager::IsHMIMessageForPlugin(application_manager::MessagePtr msg) {
...
```
dcherniev: @indlu Here is the same case as above.

src/components/functional_module/src/plugin_manager.cc
indlu: @dcherniev , fixed in a7bb34b

src/components/functional_module/src/plugin_manager.cc
indlu: @dcherniev , fixed in a7bb34b

PR #1454, opened 2017-04-06
Compare size with updated value of icon storage

VVeremjova:
SDL should save icons after removing old ones. Compare max size with updated size after removing


VVeremjova:
@pold500, @LevchenkoS  Please review

PR #1460, opened 2017-04-07
Fix missing ACK/NACK after PTU

tkireva:
Missing check is certificate is expired(missing) after PTU.
When it is missing or expired, NACK should be sent for
nonNavi apps and ACK(encryption false) for Navi.

### DIFF
--- a/src/components/protocol_handler/src/protocol_handler_impl.cc
+++ b/src/components/protocol_handler/src/protocol_handler_impl.cc
@@ -1105,8 +1105,16 @@ RESULT_CODE ProtocolHandlerImpl::HandleControlMessageStartSession(
   if (si.start_protected_ && security_manager_) {
     const uint32_t connection_key =
         session_observer_.KeyFromPair(connection_id, session_id);
+    struct SessionInfo session_info = SessionInfo(connection_id,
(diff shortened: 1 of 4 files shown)


tkireva:
Please review:
@okozlovlux @LitvinenkoIra

src/components/protocol_handler/src/protocol_handler_impl.cc
```diff
@@ -1130,30 +1146,29 @@ RESULT_CODE ProtocolHandlerImpl::HandleControlMessageStartSession(
...
     }
-    struct SessionInfo session_info = SessionInfo(connection_id,
-                                                  packet.session_id(),
-                                                  protocol_version,
-                                                  packet.service_type(),
-                                                  si.hash_id_,
-                                                  si.service_exists_);
+    // Certificate is existing and doesn't rquire update,
+    // try to start encrypted service
     StartEncryptedService(session_info);
```
VVeremjova: @tkireva Please be mentioned that in OnSessionStartedCallback service has been already started

tkireva: @VVeremjova Maybe the comment and the method name must be edited

PR #1461, opened 2017-04-07
Fix responding StartServiceNACK in case of ForceProtectedService and protected service cannot be started

LevchenkoS:
@VVeremjova, @tkireva, @indlu, @pold500, please, review.

### DIFF
--- a/src/components/connection_handler/src/connection_handler_impl.cc
+++ b/src/components/connection_handler/src/connection_handler_impl.cc
@@ -285,6 +285,7 @@ uint32_t ConnectionHandlerImpl::OnSessionStartedCallback(
     const uint8_t session_id,
     const protocol_handler::ServiceType& service_type,
     const bool is_protected,
+    const bool can_start,
     struct ExistingSessionInfo* out_session_info) {
   LOG4CXX_AUTO_TRACE(logger_);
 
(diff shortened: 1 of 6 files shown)


src/components/connection_handler/src/connection_handler_impl.cc
```diff
@@ -263,8 +263,8 @@ bool AllowProtection(const ConnectionHandlerSettings& settings,
                      const protocol_handler::ServiceType& service_type,
                      const bool is_protected) {
   const std::vector<int>& force_unprotected_list =
-      is_protected ? settings.force_unprotected_service()
-                   : settings.force_protected_service();
+      is_protected ? settings.force_protected_service()
+                   : settings.force_unprotected_service();
```
tkireva: @LevchenkoS I disasgree with these changes, because previous implementation works as expected:
is_protected is a flag that show the requested protection state
If one service is about to be started as a protected(is_protected == true) then SDL should check the .ini file if there is a prohibition for this service to be started as protected (if it is force unprotected), so that it checks in settings.force_unprotected_service().
If requested protection is false (the service is about to be started as unprotected), the SDL should check if it is allowed not to be protected (if it is not force_protected), so that it looks for it in settings.force_protected_service().
In both cases if any match is found, this service will be rejected.
This piece of code works perfectly fine.

PR #1473, opened 2017-04-12
Fix sending UPDATE_NEEDED after failed ptu retry sequence

VVeremjova:
In case it is last PTU sending of notification should be starting right after updating


VVeremjova:

PR #1477, opened 2017-04-12
Fixed bug OnPermissionChange sent twice

pold500:
Fixed bug that OnPermissionChange sent twice when needs to change only one time.
Added checks that if we had and NEED an update in externalConsent properties
then we ignore system permission consent requests.

That fix should finally fix the failed CRQ.

### DIFF
--- a/src/components/application_manager/src/policies/policy_handler.cc
+++ b/src/components/application_manager/src/policies/policy_handler.cc
@@ -540,7 +540,11 @@ void PolicyHandler::OnAppPermissionConsentInternal(
     const uint32_t connection_key, PermissionConsent& out_permissions) {
 #endif
   POLICY_LIB_CHECK_VOID();
-
+#ifdef EXTERNAL_PROPRIETARY_MODE
+  const bool k_isNeedToUpdateExternalConsent =
+      policy_manager_->IsNeedToUpdateExternalConsentStatus(
+          external_consent_status);
+#endif
   if (connection_key) {
     ApplicationSharedPtr app = application_manager_.application(connection_key);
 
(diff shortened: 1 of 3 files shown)


pold500:
@VVeremjova please review

src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -587,16 +596,21 @@ void PolicyHandler::OnAppPermissionConsentInternal(
 
       out_permissions.policy_app_id = it->second;
       out_permissions.device_id = it->first;
-      policy_manager_->SetUserConsentForApp(out_permissions);
+#ifdef EXTERNAL_PROPRIETARY_MODE
+      if (!k_isNeedToUpdateExternalConsent) {
+#endif
```
VVeremjova: @pold500 I think it would be more readable: 
```
#ifdef EXTERNAL_PROPRIETARY_MODE
  if (!k_isNeedToUpdateExternalConsent) {
     policy_manager_->SetUserConsentForApp(out_permissions);
    }
#else 
    policy_manager_->SetUserConsentForApp(out_permissions);
#endif // EXTERNAL_PROPRIETARY_MODE
```

pold500: @VVeremjova maybe, but otherwise we save one line of a code :)

pold500: @VVeremjova please take a look at https://github.com/smartdevicelink/sdl_core/pull/1477/commits/0bbe32d43831b030ee90d6589246fffdc2394083

pold500:
Rebuild required

PR #1482, opened 2017-04-13
Temporary solution for finish work of retry timeout after finishing updating

VVeremjova:
PTU should finish retry cycle only after all update notifications are finished.

According requirements, PTU should be finished and startservice which trigger PTU should be sent only after finish PTU sequence.


VVeremjova:

VVeremjova:
Rebuild required

PR #1502, opened 2017-04-21
Add API Patch Version

### LINKED ISSUE #1494: [SDL 0050] Add API Patch Version
Proposal: [Add API Patch Version](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0050-add-api-patch-version.md)

> This proposal is for the addition of a patch version into the Mobile and HMI APIs, as well as the SyncMsgVersion struct.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/151

Steering Committee Decision:

> The Steering Committee has fully agreed that this proposal is a good idea and should be accepted.

jacobkeeler:
Resolves #1494

### DIFF
--- a/src/components/interfaces/MOBILE_API.xml
+++ b/src/components/interfaces/MOBILE_API.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" standalone="no"?>
 <?xml-stylesheet type="text/xml" href="protocol2html.xsl"?>
 
-<interface name="Ford Sync RAPI" version="4.3" date="2017-03-29">
+<interface name="Ford Sync RAPI" version="4.4.0" date="2017-04-27">
 
   <enum name="Result" internal_scope="base">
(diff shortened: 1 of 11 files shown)


src/components/interfaces/MOBILE_API.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" standalone="no"?>
 <?xml-stylesheet type="text/xml" href="protocol2html.xsl"?>
 
-<interface name="Ford Sync RAPI" version="4.3" date="2017-03-29">
+<interface name="Ford Sync RAPI" version="4.3.1" date="2017-04-20">
```
joeygrover: This should force a minor version increase. Since checking minor version 3 will not yield any useful results.

jacobkeeler: @joeygrover fixed

PR #1510, opened 2017-04-26
Fix wrong if case in policy

OHerasym:
### Pull request changes:

 -Fix wrong if case in policy


OHerasym:
Please review: @indlu

PR #1520, opened 2017-05-03
Fix SDL does not send OnPermissionsChange after PTU

LitvinenkoIra:
Send OnPermissionsChange after app groups updated
Enable sending notifications for apps that have groups without user_consent


LitvinenkoIra:
@okozlovlux @LevchenkoS @LuxoftAKutsan please review

LuxoftAKutsan:
@LitvinenkoIra please try not to mix camel case and snake case in branch naming

PR #1526, opened 2017-05-05
Start PTU if previous one was not failed

VVeremjova:
In case PTU failed new PTU should not start.


VVeremjova:

PR #1528, opened 2017-05-05
Add logs for analyse fails on CI

VVeremjova:


LuxoftAKutsan:
@VVeremjova do you have some success according this issue?

Issue #1057, opened 2016-11-22
[SetGlobalProperties] SDL must support new "autoCompleteList" parameter at "KeyboardProperties" struct


mghiumiusliu:
**Description**
SDL must support and process new "autoCompleteList" param at SetGlobalProperties RPC

**Information**
SDL must respond with _"GENERIC_ERROR, success:false"_ in case HMI *_does NOT respond_* during <DefaultTimeout>

**Notes for HMI expected behavior (required for Guidelines)**
> 1.  "The system shall have the ability to receive and store multiple strings for autoCompleteText per app."
> 2.  "When the system receives a new list of strings for autoCompleteText for a particular app, the system shall delete the previous list and replace it with the new list for that app."
> 3.  "When any of the keyboard layouts are being used, the system shall reference the list of autoCompleteText strings for that app."
> 4. "As the user enters data on the keyboard, the system shall display the autoCompleteText strings which match the entry."
> 5. "The number of matching autoCompleteText strings displayed shall only be limited by the character length constraints of the hmi."
> 6. "The system shall provide the user the ability to select one of the displayed matching autoCompleteText strings without having to enter the entire string."
> 7. "When the user selects one of the displayed matching autoCompleteText string, the system shall submit that entry and not require further user input for submission."

**Impact**
SDL 
HTML5 HMI 
SPT 
TestCases 
Requirements

JackLivio:
@mghiumiusliu Closing since this is a requirement and not an issue. If this CRQ is still being worked on please relocate to https://github.com/smartdevicelink/sdl_requirements

LevchenkoS:
PR #1508

PR #1538, opened 2017-05-11
Feature/add new parameter steering wheel location to register app interface response

LuxoftAKutsan:
SDL must support and process new autoCompleteList param at SetGlobalProperties RPC 
Implements https://github.com/smartdevicelink/sdl_core/issues/1057


LuxoftAKutsan:
@LevchenkoS @Itileda please review

PR #1483, opened 2017-04-13
Convert seconds to miliseconds before adding them to retry timeout

VVeremjova:
@okozlovlux, @LevchenkoS, @LitvinenkoIra please review

PR #1535, opened 2017-05-10
PolicyTableUpdate must be triggered in case no "certificate" exist at "module_config" section of PolicyTable"

LitvinenkoIra:
Implementation of [#1046](https://github.com/smartdevicelink/sdl_core/issues/1046)

PR #1542, opened 2017-05-15
Fix sending UPDATE_NEEDED notification

VVeremjova:
After clarification UPDATE_NEEDED should be sent after retry sequence timeout, not after fixed timeout.
For this purpose there was removed thread for update status manager.

Related to PR #1483


VVeremjova:

VVeremjova:
Will be merged as a part of #1535

Issue #1457, opened 2017-04-06
[SDL 0045] External Policy Manager


theresalech:
Proposal: [External Policy Manager](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0045-external-policy-manager.md)

> This proposal is for the addition of an example external policy manager to the sdl core repository. The external policy manager is an application/service that runs on the vehicle’s head unit and is responsible for encrypting/decrypting policy table data when a policy table update takes place.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/134

Steering Committee Decision:

> This proposal has been accepted by the Steering Committee. When this is implemented, it should include a clear description and background, and call out that it shouldn't be used in an embedded environment, but instead serve as a good reference for encrypting and decrypting a policy table when a policy update takes place. The Steering Committee also supported submitting a separate proposal for a production-ready policy manager, which can be used by OEMs if they so choose.

PR #1543, opened 2017-05-15
Sample External Policy Manager

JackLivio:
Takes care of #1457


src/appMain/CMakeLists.txt
```diff
@@ -221,3 +224,14 @@ if (CMAKE_SYSTEM_NAME STREQUAL "QNX")
...
+
+if (${EXTENDED_POLICY} STREQUAL "EXTERNAL_PROPRIETARY")
+  install(FILES start_external_proprietary.sh DESTINATION bin
+      PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ
+      GROUP_EXECUTE WORLD_READ WORLD_EXECUTE RENAME start.sh)
+else()
+  install(FILES start.sh DESTINATION bin
+      PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE GROUP_READ
+      GROUP_EXECUTE WORLD_READ WORLD_EXECUTE)
+endif()
```
jacobkeeler: Minor issue, it appears that `start.sh` is not updated if the project was built in one configuration, then built in another. I built with `EXTERNAL_PRORIETARY` after an `HTTP` build, and the start.sh present was already marked as `Up to Date` during the install step.

src/appMain/start_external_proprietary.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+python sample_policy_manager.py --pack_port 8088 --unpack_port 8089 &
+POLICY_MANAGER=$!
```
jacobkeeler: Not certain how easy it is to do, but it would be useful for the startup of SDL Core to fail if you don't have the necessary python packages installed (or at least give a warning with what went wrong). The README should also be updated with the python dependencies.

JackLivio: I think Core should still continue running since it is independent of the external policy manager. Also errors from starting the python script will be logged in the console

PR #1557, opened 2017-05-24
Fix GetVehicleData request sending by SDL without mobile request

AByzhynar:
- Removed redundant sending of GetVehicleData request as
  it should be sent only in case of appropriate mobile request
- Fixed appropriate unit tests


AByzhynar:
Rebuild required

PR #1558, opened 2017-05-24
ConsumerFriendlyMessages Null Fix

JackLivio:


JackLivio:
@jacobkeeler Please review

PR #1559, opened 2017-05-25
Fix IsReady acceptance after error response

VVeremjova:
If result code from HMI on .IsReady is errorCode then SDL ) SDL should consider that works incorrectly and not available.

### DIFF
--- a/src/components/application_manager/src/commands/hmi/request_to_hmi.cc
+++ b/src/components/application_manager/src/commands/hmi/request_to_hmi.cc
@@ -46,12 +46,27 @@ bool CheckAvailabilityHMIInterfaces(ApplicationManager& application_manager,
 bool ChangeInterfaceState(ApplicationManager& application_manager,
                           const smart_objects::SmartObject& response_from_hmi,
                           HmiInterfaces::InterfaceID interface) {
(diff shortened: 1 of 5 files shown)


VVeremjova:
@okozlovlux , @LitvinenkoIra Please review

src/components/application_manager/src/commands/hmi/request_to_hmi.cc
```diff
@@ -46,12 +46,27 @@ bool CheckAvailabilityHMIInterfaces(ApplicationManager& application_manager,
 bool ChangeInterfaceState(ApplicationManager& application_manager,
                           const smart_objects::SmartObject& response_from_hmi,
                           HmiInterfaces::InterfaceID interface) {
+  using namespace helpers;
   if (response_from_hmi[strings::msg_params].keyExists(strings::available)) {
+    const hmi_apis::Common_Result::eType result_code =
+        static_cast<hmi_apis::Common_Result::eType>(
+            response_from_hmi[strings::params][hmi_response::code].asInt());
+
+    bool is_successful_resultCode =
```
okozlovlux: @VVeremjova ,  `is_successful_result_code`

PR #1569, opened 2017-05-26
Make Size() function thread safe to avoid core dump

LitvinenkoIra:
Fixes https://github.com/smartdevicelink/sdl_core/issues/2603

### Risk

### Summary
The Size() function calls the CheckSetSizes() function which compares the sizes of the two sets and checks this value in DCHECK. So we must be sure that the function Size() will not be called at the same time when there is addition or removal of requests from sets

### Changelog

* Added lock to Size() function to make it thread safe.


AByzhynar:
@LitvinenkoIra Add please or @LuxoftAKutsan or @dev-gh to code review

dev-gh:
@LitvinenkoIra 
Add description for PR

src/components/application_manager/src/request_info.cc
```diff
@@ -257,6 +257,7 @@ uint32_t RequestInfoSet::RemoveMobileRequests() {
 }
 
 const size_t RequestInfoSet::Size() {
+  sync_primitives::AutoLock lock(this_lock_);
```
dev-gh: @LitvinenkoIra 
Please rename 'this_lock_' to something meaningful.

LitvinenkoIra: @dev-gh do you mean in the whole class? I left it so because of the comments in request_info.h in line 257
`  // the lock called this_lock_, since the class represent collection by itself.
  sync_primitives::Lock this_lock_;`

dev-gh: @LitvinenkoIra 
I saw this comment however from places where it is used you can figure out that it locks *_pending_requests_ containers so I would suggest to rename to something like 'pending_requests_lock_'
To me name 'this_lock_' looks like something temporary or draft as it similar to 'my_lock' , 'tmp' etc. and does not provide any clues for people who read the code.

LitvinenkoIra: @dev-gh
Done ab2b23a553acc63427f9c39e31b4cdc16317982f

jacobkeeler:
@LitvinenkoIra please create issue for this.

HSavynetska:
Сreated tasks because there was a request for it #2603

PR #1571, opened 2017-05-29
Replaces assert macro with logic in case of failed handshake

dev-gh:
Establishing of secure connection requires checking of application name and
application id which can be assigned only to already registered application.
Due to wrong mobile behavior (it starts handshake before RAI response received)
SDL triggers assertion on missing application. So it is replaced with logic
and in case application is not registered yet handshake will fail and no
secure connection will be established. Mobile can try to establish that
connection after application is registered.

### DIFF
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -1267,7 +1267,9 @@ bool ApplicationManagerImpl::OnHandshakeDone(
   using namespace helpers;
 
   ApplicationSharedPtr app = application(connection_key);
-  DCHECK_OR_RETURN(app, false);
+  if (!app) {
+    return false;
+  }
   if (Compare<SSLContext::HandshakeResult, EQ, ONE>(
           result,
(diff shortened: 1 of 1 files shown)


AByzhynar:
@LitvinenkoIra Please review

src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -1267,7 +1267,9 @@ bool ApplicationManagerImpl::OnHandshakeDone(
   using namespace helpers;
 
   ApplicationSharedPtr app = application(connection_key);
-  DCHECK_OR_RETURN(app, false);
+  if (!app) {
+    return false;
```
dev-gh: @LuxoftAKutsan 
There is a logging during call to application() function

jacobkeeler:
@dev-gh please create/attach issue

HSavynetska:
Сreated tasks because there was a request for it #2604

ychernysheva:
This PR is not actual, because corresponding issue has been already fixed in #2921

PR #1572, opened 2017-05-29
Fix UnregisterAppInterface for not found app

pvvasilev:
If an app is not found in SDL but the mobile side
sends UnregisterAppInterface, SDL creates the mobile
request but never runs it because the app is not found.
Then the request times out and the response to the
mobile side is `GENERIC_ERROR`. Now command will be
ran and the proper response of `APPLICATION_NOT_REGISTERED`
will be sent.

### DIFF
--- a/src/components/application_manager/src/commands/command_request_impl.cc
+++ b/src/components/application_manager/src/commands/command_request_impl.cc
@@ -487,9 +487,11 @@ mobile_apis::Result::eType CommandRequestImpl::GetMobileResultCode(
 bool CommandRequestImpl::CheckAllowedParameters() {
   LOG4CXX_AUTO_TRACE(logger_);
 
-  // RegisterAppInterface should always be allowed
-  if (mobile_apis::FunctionID::RegisterAppInterfaceID ==
-      static_cast<mobile_apis::FunctionID::eType>(function_id())) {
+  // RegisterAppInterface and UnregisterAppInterface should always be allowed
+  const mobile_apis::FunctionID::eType func_id =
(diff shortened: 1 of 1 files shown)


pvvasilev:
Please review @AKalinich-Luxoft @anosach-luxoft @VProdanov

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -487,9 +487,11 @@ mobile_apis::Result::eType CommandRequestImpl::GetMobileResultCode(
 bool CommandRequestImpl::CheckAllowedParameters() {
   LOG4CXX_AUTO_TRACE(logger_);
 
-  // RegisterAppInterface should always be allowed
-  if (mobile_apis::FunctionID::RegisterAppInterfaceID ==
-      static_cast<mobile_apis::FunctionID::eType>(function_id())) {
+  // RegisterAppInterface and UnregisterAppInterface should always be allowed
+  const mobile_apis::FunctionID::eType func_id =
+      static_cast<mobile_apis::FunctionID::eType>(function_id());
+  if (mobile_apis::FunctionID::RegisterAppInterfaceID == func_id ||
```
AKalinich-Luxoft: @pvvasilev I propose you to use `helpers::Compare` here

pvvasilev: @AKalinich-Luxoft Done in cb7f7b9ca7e2820e53b03ce516ded6e376f281e8

Issue #989, opened 2016-11-07
[API] SDL responds "UNSUPPORTED_RESOURCE", success= false in case only have "UNSUPPORTED_RESOURCE" to Navigation.AlertManeuver


DrachenkoAnastasiia:
## Occurrence
Always

## Preconditions
-SDL Core and HMI are started.
-Navigation interface is not supported.
-AlertManeuver is allowed by Policy.
-App is registered, HMI level = FULL.

## Steps to reproduce
1. Send AlertManeuver() to SDL.
2. HMI sends SUCCESS response to "TTS.Speak" and UNSUPPORTED_RESOURCE to "Navigation.AlertManeuver".   
_**Note:**_ Please use the attached file for reproduce this defect automatically.

## Actual result
SDL returns UNSUPPORTED_RESOURCE, success: false to mobile app.

## Expected result
SDL should returns UNSUPPORTED_RESOURCE, success: true to mobile app.

# Environment
SDL repo: https://github.com/smartdevicelink/sdl_core.git
SDL branch: master
SDL commit: a6c32b4
HMI: https://github.com/LuxoftSDL/sdl_hmi: 01cb5b49883ed6288e34fcf7db24fd0439d08312
OS: Ubuntu 14.04 LTS

# Attachments
[att_issue#989.zip](https://github.com/smartdevicelink/sdl_core/files/575981/att_issue.989.zip)

# Expected delivery   
- [ ] Source code updates 
- [ ] Code comments 
- [ ] UTs add/update 
- [x] ATF tests add/update ***(not required)***  
- [x] Manual tests ***(not required)*** 
- [x] Add/update CI plans/jobs ***(not required)***  
- [x] SDD updates ***(not required)***  
- [x] Guidelines update ([sdl_core_guides](https://github.com/smartdevicelink/sdl_core_guides)) ***(not required)***  
- [x] Guidelines update ([sdl_hmi_integration_guidelines](https://github.com/smartdevicelink/sdl_hmi_integration_guidelines))  ***(not required)***

PR #1576, opened 2017-05-30
Fixes responses 'success' value in case of UNSUPPORTED_RESOURCE

dev-gh:
There is only one case for 'success' to have 'false' with UNSUPPORTED_RESOURCE - 
when single-interface RPC comes and that specific interface is
not available i.e. <interface>.IsReady returned 'false'
Changes done for base class and unit tests with wrong expectations.
Also renamed Response_info::is_invalid_enum to ResponseInfo::is_not_used as
 INVALID_ENUM value considered as that particular request hadn't been sent to
HMI (missed in original RPC or due to other reasons)

### DIFF
--- a/src/components/application_manager/src/commands/command_request_impl.cc
+++ b/src/components/application_manager/src/commands/command_request_impl.cc
@@ -105,6 +105,14 @@ const std::string CreateInfoForUnsupportedResult(
 }
 
 bool CheckResultCode(const ResponseInfo& first, const ResponseInfo& second) {
+  if (!first.is_ok && second.is_not_used) {
+    return false;
+  }
+
+  if (!second.is_ok && first.is_not_used) {
+    return false;
+  }
+
   if (first.is_unsupported_resource && second.is_not_used &&
       HmiInterfaces::STATE_NOT_AVAILABLE == first.interface_state) {
     return false;
@@ -205,6 +213,7 @@ void CommandRequestImpl::SendResponse(
     const mobile_apis::Result::eType& result_code,
     const char* info,
     const smart_objects::SmartObject* response_params) {
+  LOG4CXX_AUTO_TRACE(logger_);
   {
     sync_primitives::AutoLock auto_lock(state_lock_);
     if (kTimedOut == current_state_) {
@@ -778,6 +787,7 @@ bool CommandRequestImpl::PrepareResultForMobileResponse(
 
 bool CommandRequestImpl::PrepareResultForMobileResponse(
     ResponseInfo& out_first, ResponseInfo& out_second) const {
+  LOG4CXX_AUTO_TRACE(logger_);
   using namespace helpers;
 
   out_first.is_ok = Compare<hmi_apis::Common_Result::eType, EQ, ONE>(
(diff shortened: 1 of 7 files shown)


dev-gh:
@AKalinich-Luxoft @LitvinenkoIra
Please review

src/components/application_manager/test/commands/mobile/alert_maneuver_request_test.cc
```diff
@@ -71,6 +72,9 @@ typedef SharedPtr<AlertManeuverRequest> CommandPtr;
 class AlertManeuverRequestTest
     : public CommandRequestTest<CommandsTestMocks::kIsNice> {
  public:
+  AlertManeuverRequestTest()
+      : mock_message_helper_(*MockMessageHelper::message_helper_mock()) {}
```
AByzhynar: @dev-gh Are you sure that expectations for message helper mock are cleared after other unit tests? They can be still expected as this object is static

dev-gh: @AByzhynar 
See https://github.com/smartdevicelink/sdl_core/pull/1576/files#diff-1b0a4cf17aaacb2e65a92bc0a8b7fadbR121

AByzhynar: @dev-gh I would recommend to make expectations clean up in base class. Not in every RPC.

dev-gh: @AByzhynar 
Added expectations cleanup before run of test case with f9c1c4b

theresalech:
Per Luxoft, this fixes Issue #989

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -708,10 +709,10 @@ bool CommandRequestImpl::PrepareResultForMobileResponse(
...
 
-  out_first.is_invalid_enum =
+  out_first.is_not_used =
       hmi_apis::Common_Result::INVALID_ENUM == out_first.result_code;
 
-  out_second.is_invalid_enum =
+  out_second.is_not_used =
       hmi_apis::Common_Result::INVALID_ENUM == out_second.result_code;
 
   out_first.is_unsupported_resource =
```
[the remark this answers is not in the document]
AByzhynar: @JackLivio As I can see it is used in `bool CheckResultCode()` method which is a part of setting the result.

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -102,18 +102,19 @@ const std::string CreateInfoForUnsupportedResult(
 }
 
 bool CheckResultCode(const ResponseInfo& first, const ResponseInfo& second) {
-  if (first.is_ok && second.is_unsupported_resource &&
```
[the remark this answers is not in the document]
LuxoftAKutsan: @JackLivio if one interface is not available, and the request is only for one interface result code should be true.

src/components/application_manager/src/commands/command_request_impl.cc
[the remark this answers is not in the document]
LuxoftAKutsan: @JackLivio  out_second is output parameter and is_unsupported_resource may be used in several after calling  this function

AByzhynar:
Rebuild required

AByzhynar:
PR rebased to current develop

AByzhynar:
Rebuild required

AByzhynar:
Rebuild required

PR #1586, opened 2017-06-01
Change navigationText field name in UpdateTurnList request

anosach-luxoft:
`turnText` field name is legacy and should be replaced with `navigationText`


iCollin:
`turnText` has been used for a while now and I don't think we should change the field name now.

Issue #1015, opened 2016-11-07
SDL does not send VehicleInfo.GetVehicleData in case HMI responds invalid json for TTS.IsReady and true for VehicleInfo.IsReady


KhrystynaDubovyk:
## Occurrence
Always.

## Preconditions
NA

## Steps to reproduce
Or: (Go to atf folder -> Run command: ./start.sh **\<attached script\>**)

1. Send TTS.IsReady from SDL to HMI
~~~
{"id":3,"jsonrpc":"2.0","method":"TTS.IsReady"}
~~~
2. Send invalid json response from HMI to SDL for TTS.IsReady request
~~~
{"id":3,"jsonrpc";"2.0","result":{"available":true,"method":"TTS.IsReady", "code":0}}
~~~
3. Send VehicleInfo.IsReady from SDL to HMI
~~~
{"id":6,"jsonrpc":"2.0","method":"VehicleInfo.IsReady"}
~~~
4. Send available = true from HMI to SDL for "VehicleInfo.IsReady request
~~~
{"result":{"available":true,"code":0,"method":"VehicleInfo.IsReady"},"jsonrpc":"2.0","id":6} 
~~~

## Actual result
SDL does not send VehicleInfo.GetVehicleData to HMI.

## Expected result
SDL should send VehicleInfo.GetVehicleData to HMI.

# Environment
Release: 4.2.0 (master marge)
Repository: https://github.com/smartdevicelink/sdl_core
Commit: a6c32b45832653d6b851be9b19355fa35f377e07

# Attachments
[att_issue#1015.zip](https://github.com/smartdevicelink/sdl_core/files/576228/att_issue.1015.zip).

# Expected delivery   
- [ ] Source code updates 
- [ ] Code comments 
- [ ] UTs add/update 
- [x] ATF tests add/update ***(not required)***  
- [x] Manual tests ***(not required)*** 
- [x] Add/update CI plans/jobs ***(not required)***  
- [x] SDD updates ***(not required)***  
- [x] Guidelines update ([sdl_core_guides](https://github.com/smartdevicelink/sdl_core_guides)) ***(not required)***  
- [x] Guidelines update ([sdl_hmi_integration_guidelines](https://github.com/smartdevicelink/sdl_hmi_integration_guidelines))  ***(not required)***

PR #1590, opened 2017-06-02
Fix CMessageBroker invalid JSON object processing

AKalinich-Luxoft:
Fixed CMessageBroker invalid JSON object processing. The root cause of the problem is that CMessageBroker drops whole message buffer when it receives some invalid json objects. It seems to be not correct because message buffer could contain also some valid json objects.

For example CMessageBroker receives buffer where:
`{valid_response_1}{invalid_response_2}{valid_response_3}...`
Then it parse JSON objects one by one until buffer become empty. `{valid_response_1}` will be parsed OK and SDL will receive parsed message, but `{invalid_response_2}` will not be parsed OK and CMessageBroker will not continue parsing objects after that. Then if CMessageBroker receives some `{new_valid_response}` and parse it OK, it will drop the whole buffer with `{invalid_response_2}{valid_response_3}...` so SDL will not receive parsed messages of `{valid_response_3}` and all others in buffer and this is a reason why SDL will not react on them.

In this commit:
- In case of invalid JSON object received CMessageBroker tries to find next JSON object after current invalid and if it was found, current invalid JSON object will be dropped and parsing will be continued
- Fixed string formatting in MB debug messages to avoid compile errors when MB logs are enabled

### DIFF
--- a/src/3rd_party-static/message_broker/src/lib_messagebroker/CMessageBroker.cpp
+++ b/src/3rd_party-static/message_broker/src/lib_messagebroker/CMessageBroker.cpp
@@ -458,12 +458,22 @@ void CMessageBroker::onMessageReceived(int fd, std::string& aJSONData, bool tryH
          * It can be a coincidence of course, but we have to give it a try. */
         return;
       } else if ('{' == aJSONData[0]) {
-        DBG_MSG_ERROR((" Incomplete JSON object probably.\n"));
+        const bool is_object = true;
+        const size_t next_object_pos =
+            p->jumpOverJSONObjectOrArray(is_object, aJSONData);
+
+        if (next_object_pos != std::string::npos) {
+          DBG_MSG_ERROR(("Invalid JSON object probably. Skipping.\n"));
+          aJSONData.erase(0, next_object_pos);
+          DBG_MSG(("Buffer after cut is: '%s'\n", aJSONData.c_str()));
+          continue;
+        }
+        DBG_MSG_ERROR(("Incomplete JSON object probably.\n"));
         return;
       } else {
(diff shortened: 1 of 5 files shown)


AKalinich-Luxoft:
@AByzhynar, @LitvinenkoIra, @LuxoftAKutsan, @dev-gh please review

src/3rd_party-static/message_broker/src/lib_messagebroker/CMessageBroker.cpp
```diff
@@ -444,26 +444,37 @@ void CMessageBroker::onMessageReceived(int fd, std::string& aJSONData, bool tryH
...
+        const bool is_object = true;
+        const size_t next_object_pos =
+            p->jumpOverJSONObjectOrArray(is_object, aJSONData);
+
+        if (next_object_pos != std::string::npos) {
+          DBG_MSG_ERROR(("Invalid JSON object probably. Skipping.\n"));
+          aJSONData.erase(0, next_object_pos);
+          DBG_MSG(("Buffer after cut is: '%s'\n", aJSONData.c_str()));
+          continue;
+        } else {
```
AByzhynar: @AKalinich-Luxoft Possibly `else` is redundant

AKalinich-Luxoft: @AByzhynar fixed in https://github.com/smartdevicelink/sdl_core/pull/1590/commits/2e33280a1c6cb077fa5a301f967c859318501e24

theresalech:
Per Luxoft, this fixes issue #1015

PR #1429, opened 2017-03-29
Fix SystemRequest with filename=IVSU but w/o binary data

istoilovgithub:
The fix removes legacy support for 'IVSU' files.

Issue #976, opened 2016-11-07
[IVSU] SDL doesn't reject SystemRequest with filenam=IVSU but w/o binary data.


DrachenkoAnastasiia:
## Occurrence
Always

## Preconditions
1. Core, HMI started.
2. App is registered and activated on HMI.

## Steps to reproduce
1. Send systemRequest. specify name IVSU, don't select a file on sdcard.

## Actual result
SDL replies with SUCCESS and resend this SystemRequest to HMI.

## Expected result
SDL should reject such request.

# Environment
SDL repo: https://github.com/smartdevicelink/sdl_core.git
SDL branch: master
SDL commit: a6c32b4
HMI: https://github.com/LuxoftSDL/sdl_hmi: 01cb5b49883ed6288e34fcf7db24fd0439d08312
OS: Ubuntu 14.04 LTS

# Expected delivery   
- [ ] Source code updates 
- [ ] Code comments 
- [ ] UTs add/update 
- [x] ATF tests add/update ***(not required)***  
- [x] Manual tests ***(not required)*** 
- [x] Add/update CI plans/jobs ***(not required)***  
- [x] SDD updates ***(not required)***  
- [x] Guidelines update ([sdl_core_guides](https://github.com/smartdevicelink/sdl_core_guides)) ***(not required)***  
- [x] Guidelines update ([sdl_hmi_integration_guidelines](https://github.com/smartdevicelink/sdl_hmi_integration_guidelines))  ***(not required)***

jacobkeeler:
Closing with the merge of #1625

PR #1625, opened 2017-06-19
Fix SystemRequest with filename=IVSU but w/o binary data

LuxoftAKutsan:
Resend pr #1429 to develop branch 
SDL should reject SystemRequest with specified filename but without attached binary data. 
github issue : 
https://github.com/smartdevicelink/sdl_core/issues/976

### DIFF
--- a/src/components/application_manager/src/commands/mobile/system_request.cc
+++ b/src/components/application_manager/src/commands/mobile/system_request.cc
@@ -520,19 +520,8 @@ void SystemRequest::Run() {
     if (!file || !file->is_download_complete ||
         !file_system::MoveFile(app_full_file_path, file_dst_path)) {
       LOG4CXX_DEBUG(logger_, "Binary data not found.");
-
-      std::string origin_file_name;
-      if ((*message_)[strings::msg_params].keyExists(strings::file_name)) {
-        origin_file_name =
-            (*message_)[strings::msg_params][strings::file_name].asString();
-      }
-      if (!(mobile_apis::RequestType::HTTP == request_type &&
-            0 == origin_file_name.compare(kIVSU))) {
-        LOG4CXX_DEBUG(logger_, "Binary data required. Invalid data");
-        SendResponse(false, mobile_apis::Result::INVALID_DATA);
-        return;
-      }
-      LOG4CXX_DEBUG(logger_, "IVSU does not require binary data. Continue");
+      SendResponse(false, mobile_apis::Result::REJECTED);
+      return;
     }
     processing_file_ = file_dst_path;
   }
--- a/src/components/application_manager/test/commands/CMakeLists.txt
+++ b/src/components/application_manager/test/commands/CMakeLists.txt
@@ -55,13 +55,13 @@ set(LIBRARIES
   gmock
   Utils
   SmartObjects
-  jsoncpp
   HMI_API
   MOBILE_API
   ApplicationManager
   AMHMICommandsLibrary
   AMMobileCommandsLibrary
   connectionHandler
+  jsoncpp
 )
 
 create_cotired_test("commands_test" "${SOURCES}" "${LIBRARIES}" )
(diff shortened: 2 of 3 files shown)


LuxoftAKutsan:

src/components/application_manager/src/commands/mobile/system_request.cc
```diff
@@ -520,19 +520,8 @@ void SystemRequest::Run() {
...
-            (*message_)[strings::msg_params][strings::file_name].asString();
-      }
-      if (!(mobile_apis::RequestType::HTTP == request_type &&
-            0 == origin_file_name.compare(kIVSU))) {
-        LOG4CXX_DEBUG(logger_, "Binary data required. Invalid data");
-        SendResponse(false, mobile_apis::Result::INVALID_DATA);
-        return;
-      }
-      LOG4CXX_DEBUG(logger_, "IVSU does not require binary data. Continue");
+      SendResponse(false, mobile_apis::Result::REJECTED);
```
dev-gh: @LuxoftAKutsan 
Why exactly this result code?

dev-gh:
@LuxoftAKutsan 
I don't quite understand what is this PR about. 
If it is about IVSU handling (per PR title), than there should not be unit test since it is part of FS functionality.
If it is about SystemRequest and binary data (per PR description) you have to check this flow with unit test.

LuxoftAKutsan:
@dev-gh please review

src/components/application_manager/src/commands/mobile/system_request.cc
LuxoftAKutsan: SDL should reject Putfile with file path but without binary data

dev-gh: @LuxoftAKutsan 
Before it was returning INVALID_DATA

LuxoftAKutsan: @dev-gh according to github issue it should be rejected.
There is [requirements](https://adc.luxoft.com/jira/browse/APPLINK-21782?src=confmacro)

LuxoftAKutsan:
@dev-gh Unit test changes to avoid F-S behaviour

src/components/application_manager/src/commands/mobile/system_request.cc
jacobkeeler: @LuxoftAKutsan This doesn't seem to fix the described problem, the original code seemed to fulfill that purpose (specifically for the file name `IVSU`, albeit the result code needed to change):
```
 -      std::string origin_file_name;
 -      if ((*message_)[strings::msg_params].keyExists(strings::file_name)) {
 -        origin_file_name =
 -            (*message_)[strings::msg_params][strings::file_name].asString();
 -      }
 -      if (!(mobile_apis::RequestType::HTTP == request_type &&
 -            0 == origin_file_name.compare(kIVSU))) {
 -        LOG4CXX_DEBUG(logger_, "Binary data required. Invalid data");
 -        SendResponse(false, mobile_apis::Result::INVALID_DATA);
 -        return;
 -      }
```
Was this supposed to be expanded to any case where binary data is not specified?

It also doesn't seem to do what you stated either (reject if a file path is specified but no binary data is included), since it doesn't check whether a file name is specified.

AKalinich-Luxoft: @jacobkeeler the problem, described in #976 is that SDL replies with SUCCESS on SystemRequest with filename = IVSU and request type = HTTP, however SDL should reject such request. The original code was removed, because it is responsible exactly for this case.
In all other cases, when no binary data present, according to our requirements:
>In case mobile application sends valid SystemRequest with "HTTP" type, no binary data and file with "fileName" is not present in "SystemFilesPath" folder, SDL must respond with resultCode:REJECTED and success:false.

So INVALID_DATA result code is wrong here and it was replaced with correct one.
Also, code responsible for file_name parameter are at lines 463-468. In case when file name is not specified, default value of kSYNC constant will be taken.

PR #1374, opened 2017-03-15
Fix issue with incorrect behaviour in AlertManeuver

okozlovlux:
- fixed issue when in AlertManeuver SDL responds GENERIC_ERROR instead of
    INVALID_DATA when soft button has Type is Image or Both and Text is
    whitespace or \t or \n or empty
  - fixed failed UTs

Related to [Issue-980](https://github.com/smartdevicelink/sdl_core/issues/980)

Issue #980, opened 2016-11-07
[APIs] AlertManeuver: SDL responds GENERIC_ERROR instead of INVALID_DATA when soft button has Type is Image or Both and Text is whitespace or \t or \n or empty


DrachenkoAnastasiia:
## Occurrence
 Always

## Preconditions
- AlertManeuver is allowed by policy for "default" group.
- SDL core and HMI are started.
- App is connected and at the FULL.

## Steps to reproduce
1. Send AlertManeuver sure that soft button has Type is Image or Both and Text is whitespace or \t or \n or empty.
_**Note:**_ Should you the attached lua file and reproduce this defect automatically.

## Actual result
SDL responds with resultCode "GENERIC_ERROR" and success:"false". 

## Expected result
SDL must respond with resultCode "INVALID_DATA" and success:"false".

# Environment
SDL: https://github.com/smartdevicelink/sdl_core.git: a6c32b45832653d6b851be9b19355fa35f377e07
ATF: https://github.com/smartdevicelink/sdl_atf.git: 962e1eff8664d97af5e59ec4e22f39b9ea76f5eb
OS: Ubuntu 14.04 LTS

# Attachments
[att_issue#980.zip](https://github.com/smartdevicelink/sdl_core/files/575882/att_issue.980.zip).

# Expected delivery   
- [ ] Source code updates 
- [ ] Code comments 
- [ ] UTs add/update 
- [x] ATF tests add/update ***(not required)***  
- [x] Manual tests ***(not required)*** 
- [x] Add/update CI plans/jobs ***(not required)***  
- [x] SDD updates ***(not required)***  
- [x] Guidelines update ([sdl_core_guides](https://github.com/smartdevicelink/sdl_core_guides)) ***(not required)***  
- [x] Guidelines update ([sdl_hmi_integration_guidelines](https://github.com/smartdevicelink/sdl_hmi_integration_guidelines))  ***(not required)***

jacobkeeler:
Closing with the merge of #1627

PR #1627, opened 2017-06-20
Fix issue with incorrect behaviour in AlertManeuver

LuxoftAKutsan:
Resend PR #1374 to develop branch
 
Fixed issue when in AlertManeuver SDL responds GENERIC_ERROR instead of
INVALID_DATA when soft button has Type is Image or Both and Text is 
whitespace or \t or \n or empty

Related to #980

### DIFF
--- a/src/components/application_manager/src/commands/mobile/alert_maneuver_request.cc
+++ b/src/components/application_manager/src/commands/mobile/alert_maneuver_request.cc
@@ -216,17 +216,17 @@ bool AlertManeuverRequest::PrepareResponseParameters(
 
 bool AlertManeuverRequest::IsWhiteSpaceExist() {
   LOG4CXX_AUTO_TRACE(logger_);
+  using smart_objects::SmartArray;
 
   if ((*message_)[strings::msg_params].keyExists(strings::tts_chunks)) {
-    const smart_objects::SmartArray* tc_array =
+    const SmartArray* tts_chuncks_arr =
         (*message_)[strings::msg_params][strings::tts_chunks].asArray();
 
-    smart_objects::SmartArray::const_iterator it_tc = tc_array->begin();
-    smart_objects::SmartArray::const_iterator it_tc_end = tc_array->end();
+    SmartArray::const_iterator it_tts_chunk = tts_chuncks_arr->begin();
 
-    for (; it_tc != it_tc_end; ++it_tc) {
-      const char* str = (*it_tc)[strings::text].asCharArray();
-      if (strlen(str) && !CheckSyntax(str)) {
+    for (; it_tts_chunk != tts_chuncks_arr->end(); ++it_tts_chunk) {
+      const char* tts_chunk_text = (*it_tts_chunk)[strings::text].asCharArray();
+      if (strlen(tts_chunk_text) && !CheckSyntax(tts_chunk_text)) {
         LOG4CXX_ERROR(logger_, "Invalid tts_chunks syntax check failed");
         return true;
       }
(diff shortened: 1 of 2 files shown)


LuxoftAKutsan:

src/components/application_manager/src/commands/mobile/alert_maneuver_request.cc
```diff
@@ -226,13 +225,33 @@ bool AlertManeuverRequest::IsWhiteSpaceExist() {
...
-      str = (*it_tc)[strings::text].asCharArray();
+      const char* str = (*it_tc)[strings::text].asCharArray();
       if (strlen(str) && !CheckSyntax(str)) {
         LOG4CXX_ERROR(logger_, "Invalid tts_chunks syntax check failed");
         return true;
       }
     }
   }
+  if ((*message_)[strings::msg_params].keyExists(strings::soft_buttons)) {
+    DCHECK_OR_RETURN(
```
AKalinich-Luxoft: @LuxoftAKutsan maybe it will be better to replace `DCHECK` with `LOG4CXX_ERROR + return true` as it's done in lines 249-250? I don't think that wrong type in rpc message is a reason for fatal error and SDL termination.

LuxoftAKutsan:
@dev-gh please review

src/components/application_manager/src/commands/mobile/alert_maneuver_request.cc
```diff
@@ -226,13 +225,33 @@ bool AlertManeuverRequest::IsWhiteSpaceExist() {
...
+        (*message_)[strings::msg_params][strings::soft_buttons].getType() ==
+            smart_objects::SmartType_Array,
```
dev-gh: @LuxoftAKutsan 
I would suggest to have
-  'using smart_objects'
- don't use abbreviations like 'sb', 'str' as it is completely not understandable but 'soft_buttons' etc.

src/components/application_manager/src/commands/mobile/alert_maneuver_request.cc
LuxoftAKutsan: @AKalinich-Luxoft smart object schema validation is responsible for correct type of this field. It it is wrong - this requests should not be executed. Ant there it is fatal error

LuxoftAKutsan:
@dev-gh please check answers

src/components/application_manager/src/commands/mobile/alert_maneuver_request.cc
LuxoftAKutsan: @dev-gh  see cba30e2

src/components/application_manager/src/commands/mobile/alert_maneuver_request.cc
```diff
@@ -216,23 +216,41 @@ bool AlertManeuverRequest::PrepareResponseParameters(
 
 bool AlertManeuverRequest::IsWhiteSpaceExist() {
   LOG4CXX_AUTO_TRACE(logger_);
-  const char* str = NULL;
+  using smart_objects::SmartArray;
 
   if ((*message_)[strings::msg_params].keyExists(strings::tts_chunks)) {
-    const smart_objects::SmartArray* tc_array =
+    const SmartArray* tts_chuncks_arr =
```
dev-gh: @LuxoftAKutsan 
Chunks

jacobkeeler: @LuxoftAKutsan please fix typo

AByzhynar: @jacobkeeler @dev-gh Fixed in 7b57748b3213b6739f9e700c751f7a3b63c7f027

PR #1628, opened 2017-06-20
Fix unstable unit test 

LuxoftAKutsan:
Unit test OnEvent_WarningsResultCode_SUCCESS was unstable due to missed HMIToMobileResult expectation. 
Due to HMIToMobileResult returns simple type (enum) gmock generates it on some systems, but sometimes it cause gmock exception due to missed specification of return value. 
This PR adds expectation of calling HMIToMobileResult and specify return value for it.


dev-gh:
@LuxoftAKutsan 
Please write PR description that gives answers to:

   - what was the issue? 
   - what was done to resolve the issue
   - why it was done in that way?

LuxoftAKutsan:

PR #1630, opened 2017-06-20
Fix incorrect behaviour of SDL during respose ACK when not rpc service started

LuxoftAKutsan:
Resend #1413 to develop branch

In the case of start not rpc service by low protocol, SDL was answered ACK with highest supported protocol version. It is fairly for RPC service but nit correct for Audio and VIdeo. 

This PR adds check of type service before sending Ack. In case if this is RPC service SDL should respond with maximum supported version. In the case of not RPC service sdl should respond with the same version of protocol as StartService received. 

Fixes : #1365

### DIFF
--- a/src/components/protocol_handler/src/protocol_handler_impl.cc
+++ b/src/components/protocol_handler/src/protocol_handler_impl.cc
@@ -176,13 +176,16 @@ void set_hash_id(uint32_t hash_id, protocol_handler::ProtocolPacket& packet) {
 
 void ProtocolHandlerImpl::SendStartSessionAck(ConnectionID connection_id,
                                               uint8_t session_id,
-                                              uint8_t,
+                                              uint8_t protocol_version,
                                               uint32_t hash_id,
                                               uint8_t service_type,
                                               bool protection) {
   LOG4CXX_AUTO_TRACE(logger_);
-
   uint8_t protocolVersion = SupportedSDLProtocolVersion();
+
+  if (kRpc != service_type) {
+    protocolVersion = protocol_version;
+  }
 
   ProtocolFramePtr ptr(
       new protocol_handler::ProtocolPacket(connection_id,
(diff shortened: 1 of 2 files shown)


LuxoftAKutsan:

LuxoftAKutsan:
@dev-gh  please review

src/components/protocol_handler/src/protocol_handler_impl.cc
```diff
@@ -176,14 +176,17 @@ void set_hash_id(uint32_t hash_id, protocol_handler::ProtocolPacket& packet) {
...
-                                              uint8_t,
+                                              uint8_t protocol_version,
                                               uint32_t hash_id,
                                               uint8_t service_type,
                                               bool protection) {
   LOG4CXX_AUTO_TRACE(logger_);
-
   uint8_t protocolVersion = SupportedSDLProtocolVersion();
 
+  if (kRpc != service_type) {
```
dev-gh: @LuxoftAKutsan 
I believe we don't have to rely on version coming from packet header but instead on .ini file settings like max_supported_protocol_version()

LuxoftAKutsan:
@dev-gh @AByzhynar  please review

src/components/protocol_handler/src/protocol_handler_impl.cc
LuxoftAKutsan: @dev-gh this is related to rpc service but not for others

dev-gh: @LuxoftAKutsan 
So if there will be StartService(Audio, proto = 100) SDL will reply with 100, however it does not support that.

LuxoftAKutsan:
@dev-ghplease review

LitvinenkoIra:
@jacobkeeler I suppose that these changes are already included in 9df5e02922703f00c9cfeaf7606de71d85f6345f

jacobkeeler:
@LitvinenkoIra Not quite, that doesn't fix the behavior for Audio and Video services, it still sends Core's max version in these cases. I fixed the merge conflicts in this PR and merged.

PR #1635, opened 2017-06-21
Fix SDL does not send StopAudioStream

AByzhynar:
Fixed problem that SDL does not send StopAudioStream when exiting app if Video and Audio services are started and approved.
Both StopAudioStreaming and StopNaviStreaming should be sent.
The requirend functionality was missed.

Related PR in another branch: [#1381](https://github.com/smartdevicelink/sdl_core/pull/1381)
Related to [Issue-1002](https://github.com/smartdevicelink/sdl_core/issues/1002)

### DIFF
--- a/src/components/application_manager/src/application_impl.cc
+++ b/src/components/application_manager/src/application_impl.cc
@@ -40,6 +40,7 @@
 #include "interfaces/MOBILE_API.h"
 #include "utils/file_system.h"
 #include "utils/logger.h"
+#include "utils/helpers.h"
 #include "utils/gen_hash.h"
 #include "utils/make_shared.h"
 #include "utils/timer_task_impl.h"
@@ -442,11 +443,17 @@ void ApplicationImpl::StopStreaming(
 
   SuspendStreaming(service_type);
 
-  if (service_type == ServiceType::kMobileNav && video_streaming_approved()) {
-    StopNaviStreaming();
-  } else if (service_type == ServiceType::kAudio &&
-             audio_streaming_approved()) {
+  if (!helpers::Compare<ServiceType, helpers::EQ, helpers::ONE>(
+          service_type, ServiceType::kMobileNav, ServiceType::kAudio)) {
+    return;
+  }
+
+  if (audio_streaming_approved()) {
     StopAudioStreaming();
+  }
+
+  if (video_streaming_approved()) {
+    StopNaviStreaming();
   }
 }
 


AByzhynar:
@dev-gh @LuxoftAKutsan @AKalinich-Luxoft @LitvinenkoIra Please review

src/components/application_manager/src/application_impl.cc
```diff
@@ -442,7 +442,16 @@ void ApplicationImpl::StopStreaming(
 
   SuspendStreaming(service_type);
 
-  if (service_type == ServiceType::kMobileNav && video_streaming_approved()) {
+  if (service_type == ServiceType::kMobileNav && video_streaming_approved() &&
```
dev-gh: @AByzhynar 
```
if (!Compare(service, Audio, Video)) {
   return;
}
if (audio_approved()) {
   StopAudio();
}
if (video_approved()) {
   StopVideo()
}
```

dev-gh:
@AByzhynar 
Description does not explain what was the cause and why this solution.

AByzhynar:
@dev-gh Please re-check. Implementation was changed.

src/components/application_manager/src/application_impl.cc
```diff
@@ -442,11 +443,20 @@ void ApplicationImpl::StopStreaming(
...
-    StopNaviStreaming();
-  } else if (service_type == ServiceType::kAudio &&
-             audio_streaming_approved()) {
-    StopAudioStreaming();
+  if (!helpers::Compare<ServiceType, helpers::EQ, helpers::ONE>(
+          service_type, ServiceType::kMobileNav, ServiceType::kAudio)) {
+      return;
+  }
+
+  const bool as_approved = audio_streaming_approved();
```
dev-gh: @AByzhynar 
I would prefer direct calls to the methods as their names are clear as opposed to variables

AByzhynar: @dev-gh @AKalinich-Luxoft Removed redundant variables.

AByzhynar:
Rebuild required

src/components/application_manager/src/application_impl.cc
```diff
@@ -442,12 +443,18 @@ void ApplicationImpl::StopStreaming(
...
+    return;
+  }
+
+  if (audio_streaming_approved()) {
     StopAudioStreaming();
   }
+
+  if (video_streaming_approved()) {
+    StopNaviStreaming();
+  }
```
JackLivio: Because there is only one service type parameter being passed to this function I think it would be cleaner to only stop the service passed to the stop streaming function, instead of trying to stop both audio and video services in the same function.

AByzhynar: @JackLivio Agree. It would be cleaner. But in case we exit all applications - we have to stop all types of streamings approved.

LuxoftAKutsan:
Opened in https://github.com/smartdevicelink/sdl_core/pull/1835

PR #1641, opened 2017-06-22
4.3 release candidate coverity fixes

JackLivio:
Fixes for coverity issues found on the new 4.3 release candidate.


src/components/policy/policy_external/src/cache_manager.cc
```diff
@@ -1082,9 +1082,6 @@ bool CacheManager::SetUserPermissionsForApp(
       permissions.group_permissions.begin();
   std::vector<FunctionalGroupPermission>::const_iterator iter_end =
       permissions.group_permissions.end();
-  if (out_app_permissions_changed) {
-    *out_app_permissions_changed = false;
-  }
```
jacobkeeler: The assignment should still be kept. While it is initialized to false in the one case where this function is used, this really shouldn't be assumed.

PR #1648, opened 2017-06-28
Remove all references to App Launching from README

### LINKED ISSUE #1547: README should not reference "v4tester" application
In the README, it is recommended that people should use an iOS "v4tester" application that links to a random box account download. This is _absolutely not_ how people should get started. They should use the example app in the iOS library, or hello_android, or pretty much anything other than that. The "v4tester" application was never built for this purpose or even to be shared, is not supported or updated, and I have no idea how it got into distribution.

joeljfischer:
Fixes #1547

App Launching is largely done a different way now, so remove references to SDL Server, QUERY_APPS, and the v4tester application. None of these are necessary or recommended for getting started.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,54 +116,6 @@ There are several RPCs that are "required" to be implemented in order for SDL to
   * OnSystemRequest
   * Speak
 
-## App Launching
-
-Below are instructions for testing app launching and query with a full system set up.
-
-### SDL Server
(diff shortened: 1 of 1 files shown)


jacobkeeler:
Closing due to being duplicate of https://github.com/smartdevicelink/sdl_core/pull/1780 (even though this one was opened first, the other contains another important README change).

Issue #1663, opened 2017-07-14
 TransportManager that incorrectly deletes a connection object within a C++ map iterator loop


madhuy:
erasing [connection](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/transport_manager/src/transport_manager_impl.cc#L630) in connections map iterator loop causing not to release  transport adapter [connection](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/transport_manager/src/transport_adapter/transport_adapter_impl.cc#L771)

**IN SDL core logs we see below messages**

`
TRACE [12 Jul 2017 14:24:57,536][Utils] Enter
DEBUG [12 Jul 2017 14:24:57,536][Utils] Stopping thread #0 "TM DiscRoutine"
DEBUG [12 Jul 2017 14:24:57,536][Utils] Stopped thread #0 "TM DiscRoutine "
TRACE [12 Jul 2017 14:24:57,536][Utils] Exit
TRACE [12 Jul 2017 14:24:57,536][Utils] Exit
DEBUG [12 Jul 2017 14:24:57,538][Utils] Timer TM DiscRoutine has been destroyed
TRACE [12 Jul 2017 14:24:57,538][Utils] Exit
TRACE [12 Jul 2017 14:24:57,538][TransportManager] Enter
Madhu  [12 Jul 2017 14:24:57,538][TransportManager] Device_id: 0xb6bfe6d4, app_handle: 0xb695e168 connection not finalized
TRACE [12 Jul 2017 14:24:57,538][TransportManager] Exit
TRACE [12 Jul 2017 14:24:57,538][TransportManager] Exit
DEBUG [12 Jul 2017 14:24:57,538][TransportManager] eevent_type = ON_UNEXPECTED_DISCONNECT
`

[Here](http://en.cppreference.com/w/cpp/container/map/erase) is best way to erase object from map in iterator loop.

PR #1674, opened 2017-07-20
Fix/issue 1663

davidswi:
This PR fixes https://github.com/smartdevicelink/sdl_core/issues/1663. It also includes a fix for a bug in connection_handler_impl.cc where the lock is not released in the case where the end of the collection is reached in the iterator loop.


JackLivio:
@davidswi Editing your changes to connection handler impl.cc since that change has since been merged into master.

Issue #1686, opened 2017-07-27
Invalid memory access in websocket_handler


shoamano83:
Steps to reproduce the issue:
1. Check out develop branch
2. Modify CMakeLists.txt to turn on EXTENDED_MEDIA_MODE
3. Run cmake then build SDL Core
4. Start SDL Core with valgrind:
   LD_LIBRARY_PATH=. valgrind ./smartDeviceLinkCore > valgrind_memory.txt 2>&1
5. Launch sdl_hmi
6. Wait for a couple of seconds then verify output of valgrind

Expected behavior: valgrind does not detect invalid memory access in NsMessageBroker::CWebSocketHandler::parseWebSocketData()
Actual behavior: valgrind detects invalid memory access in the method

Tested environment:
Ubuntu 16.04 (x86_64)
 - SDL Core: develop branch (commit 2cf2b12f6bd60215eb9ebfcfc15e13beabc864e3) with EXTENDED_MEDIA_MODE enabled
 - sdl_hmi: latest master branch (commit 18460356c02e985df598b94f5a6bdbba1c451fff)

Notes:
- An unit test for this fix is not available since invalid memory access cannot be detected with an unit test.
- FindGstreamer-1.0.cmake needed to be modified on Ubuntu 16.04 to detect gstconfig.h
- valgrind_memory.txt is attached for reference

[valgrind_memory.txt](https://github.com/smartdevicelink/sdl_core/files/1179935/valgrind_memory.txt)

JackLivio:
@shoamano83 Verified issue exists in valgrind

PR #1687, opened 2017-07-27
fix: invalid memory access in websocket_handler

shoamano83:
This is to fix https://github.com/smartdevicelink/sdl_core/issues/1686. Using the value of 'size' seems wrong, it should be 'length'.
Note: an unit test for this fix is not available since invalid memory access cannot be detected with an unit test.


shoamano83:
@JackLivio Please kindly review.

shoamano83:
The fix was incomplete and needs some more work.

shoamano83:
This is **ready for review** now.

JackLivio:
@shoamano83 verified fix removes issue from valgrind

PR #1690, opened 2017-07-28
fix: memory leak in FromMicToFileRecorderThread

shoamano83:
This is to fix issue https://github.com/smartdevicelink/sdl_core/issues/1682.
Note: an unit test for this fix is not available since memory leaks cannot be detected with an unit test.


shoamano83:
@JackLivio Please kindly review.

jacobkeeler:
Changed base to release/4.4.0 so that it can be included in the release

PR #1707, opened 2017-08-04
Use `MaxSupportedProtocolVersion` instead of `EnableProtocolX` to determine protocol version

jacobkeeler:
An ini configuration was already implemented for `MaxSupportedProtocolVersion`.

This commit removes the `EnableProtocolX` options and instead uses the `MaxSupportedProtcolVersion` parameter to determine what features are supported.

### DIFF
--- a/src/appMain/smartDeviceLink.ini
+++ b/src/appMain/smartDeviceLink.ini
@@ -30,6 +30,8 @@ AudioStreamingPort = 5080
 ; SDL source version 
 ; represents as a git commit hash value
 SDLVersion = {GIT_COMMIT}
+; SDL supported protocol version
+MaxSupportedProtocolVersion = 5
 ; All logging event could be dropped by setting $LogsEnabled to false
 LogsEnabled = true
 ; Contains .json/.ini files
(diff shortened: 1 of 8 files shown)


jacobkeeler:
@fronneburg My understanding of the feature is that if no MaximumPayloadSize is set, then it will be automatically set to the `MAXIMUM_FRAME_DATA_V2_SIZE` you mentioned. I suppose I agree that this is an odd behavior (since I don't believe that a dynamic MTU can be set on the proxy side before v5), but this could be created as a separate issue since this PR isn't related to that behavior.

Also, you say that `MaximumControlPayloadSize` needs to be set? Did you just mean the default `MaximumPayloadSize`? I don't get any issues with rejected messages when the either of these are not set.

fronneburg:
Yes, @jacobkeeler, I mis-typed that. I meant MaximumPayloadSize. If "MaximumPayloadSize" is not set to 131072, and "EnableProtocol4" is set to "true", messages are rejected.

I do agree that it should be a separate issue, but I don't really have a feeling one way or the other about how important it is and whether we actually need to take the time to write it up.

Issue #1718, opened 2017-08-12
Build fails in protocol_packet_test.cc


shoamano83:
### Step to reproduce:
1. Check out latest develop branch (commit 4e7e56009daf055d3a5ef51d1d5c1587417390c2)
2. Apply fix of PR https://github.com/smartdevicelink/sdl_core/pull/1716
3. Enable BUILD_TESTS in CMakeLists.txt
4. create a build directory ($ cd .. && mkdir cmake_build && cd cmake_build)
5. Give a build ($ cmake ../sdl_core && make). Type your sudo password if asked

### Expected Result:
SDL Core should build without any issue

### Actual Result:
Build fails with following error message:
```
[ 27%] Building CXX object src/components/protocol_handler/test/CMakeFiles/protocol_handler_test.dir/protocol_header_validator_test.cc.o
In file included from /home/samano/projects/sdl_core_build/sdl_core/src/components/protocol_handler/test/protocol_header_validator_test.cc:33:0:
/home/samano/projects/sdl_core_build/sdl_core/src/3rd_party-static/gmock-1.7.0/gtest/include/gtest/gtest.h: In instantiation of 'testing::AssertionResult testing::internal::CmpHelperEQ(const char*, const char*, const T1&, const T2&) [with T1 = int; T2 = long unsigned int]':
/home/samano/projects/sdl_core_build/sdl_core/src/3rd_party-static/gmock-1.7.0/gtest/include/gtest/gtest.h:1524:23:   required from 'static testing::AssertionResult testing::internal::EqHelper<true>::Compare(const char*, const char*, const T1&, const T2&, typename testing::internal::EnableIf<(! testing::internal::is_pointer<T2>::value)>::type*) [with T1 = int; T2 = long unsigned int; typename testing::internal::EnableIf<(! testing::internal::is_pointer<T2>::value)>::type = void]'
/home/samano/projects/sdl_core_build/sdl_core/src/components/protocol_handler/test/protocol_header_validator_test.cc:71:3:   required from here
/home/samano/projects/sdl_core_build/sdl_core/src/3rd_party-static/gmock-1.7.0/gtest/include/gtest/gtest.h:1448:16: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
   if (expected == actual) {
                ^
cc1plus: all warnings being treated as errors
[...]

PR #1719, opened 2017-08-12
fix: build break in protocol_packet_test.cc

shoamano83:
This commit fixes https://github.com/smartdevicelink/sdl_core/issues/1718.

### DIFF
--- a/src/components/protocol_handler/test/protocol_packet_test.cc
+++ b/src/components/protocol_handler/test/protocol_packet_test.cc
@@ -125,7 +125,8 @@ TEST_F(ProtocolPacketTest, SerializePacketWithDiffServiceType) {
   for (size_t i = 0; i < serv_types.size(); ++i) {
     RawMessagePtr res =
         GetRawMessage(PROTOCOL_VERSION_3, FRAME_TYPE_CONTROL, serv_types[i]);
-    EXPECT_EQ(PROTOCOL_VERSION_3, res->protocol_version());
+    EXPECT_EQ(static_cast<uint32_t>(PROTOCOL_VERSION_3),
+              res->protocol_version());
     EXPECT_EQ(serv_types[i], res->service_type());
     EXPECT_EQ(PROTOCOL_HEADER_V2_SIZE, res->data_size());
   }
@@ -146,7 +147,8 @@ TEST_F(ProtocolPacketTest, SerializePacketWithWrongServiceType) {
   for (size_t i = 0; i < serv_types.size(); ++i) {
     RawMessagePtr res =
         GetRawMessage(PROTOCOL_VERSION_3, FRAME_TYPE_CONTROL, serv_types[i]);
-    EXPECT_EQ(PROTOCOL_VERSION_3, res->protocol_version());
+    EXPECT_EQ(static_cast<uint32_t>(PROTOCOL_VERSION_3),
+              res->protocol_version());
     EXPECT_EQ(kInvalidServiceType, res->service_type());
   }
 }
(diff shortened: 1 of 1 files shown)


shoamano83:
@jacobkeeler @JackLivio Please kindly review and merge.

jacobkeeler:
Another case where I the failure doesn't happen on my machine. I'm a little confused as to why this casting needs to be done where you've done it, but not in this test: https://github.com/shoamano83/sdl_core/blob/63eb7c774cd77589f7bff6d36301d174a0458070/src/components/protocol_handler/test/protocol_packet_test.cc#L104

shoamano83:
The compiler is complaining about comparing between signed and unsigned values.

We have no issue at line 104 since 'version' is declared as uint8_t value and res->protocol_version() returns uint8_t as well.
In the case of other tests, PROTOCOL_VERSION_3 is defined as an enum (which is integer) so it is not treated as an unsigned value. So, comparing it with res->protocol_version() causes "comparison between signed and unsigned".

By the way, I'm working on Ubuntu 16.04 machine, not 14.04. This might be the reason I'm getting extra errors.

jacobkeeler:
Meant to point to this line as well: https://github.com/shoamano83/sdl_core/blob/63eb7c774cd77589f7bff6d36301d174a0458070/src/components/protocol_handler/test/protocol_packet_test.cc#L107
```
if (res->protocol_version() == PROTOCOL_VERSION_1) {
```

This would involve a comparison between an unsigned and signed value, correct? Maybe it only throws the warning due to something in the `EXPECT_EQ` macro?

shoamano83:
I don't fully understand this, but it looks like we get this error when we explicitly specify the type of enum value.

As you pointed out, we do not get the error with this code:
```
    if (res->protocol_version() == PROTOCOL_VERSION_1) {
```
However, if I change the code like this, then the compiler throws same error:
```
diff --git a/src/components/protocol_handler/test/protocol_packet_test.cc b/src/components/protocol_handler/test/protocol_packet_test.cc
index 7819c4b..d226f7a 100644
--- a/src/components/protocol_handler/test/protocol_packet_test.cc
+++ b/src/components/protocol_handler/test/protocol_packet_test.cc
@@ -104,7 +104,9 @@ TEST_F(ProtocolPacketTest, SerializePacketWithDiffVersions) {
     EXPECT_EQ(version, res->protocol_version());
     EXPECT_EQ(kControl, res->service_type());
     EXPECT_EQ(some_connection_id_, res->connection_key());
-    if (res->protocol_version() == PROTOCOL_VERSION_1) {
+//    if (res->protocol_version() == PROTOCOL_VERSION_1) {
+    protocol_handler::MajorProtocolVersion tempVersion = PROTOCOL_VERSION_1;
+    if (res->protocol_version() == tempVersion) {
       EXPECT_EQ(PROTOCOL_HEADER_V1_SIZE, res->data_size());
     } else {
       EXPECT_EQ(PROTOCOL_HEADER_V2_SIZE, res->data_size());
```
```
[ 29%] Building CXX object src/components/protocol_handler/test/CMakeFiles/protocol_handler_test.dir/protocol_packet_test.cc.o
/home/samano/projects/sdl_core/sdl_core/src/components/protocol_handler/test/protocol_packet_test.cc: In member function ‘virtual void test::components::protocol_handler_test::ProtocolPacketTest_SerializePacketWithDiffVersions_Test::TestBody()’:
/home/samano/projects/sdl_core/sdl_core/src/components/protocol_handler/test/protocol_packet_test.cc:109:33: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
     if (res->protocol_version() == tempVersion) {
                                 ^
```
When we use EXPECT_EQ macro, types of the two values are passed to CmpHelperEQ() function. I think that is why we get the error.

shoamano83:
Thanks Jacob!

PR #1729, opened 2017-08-18
Fix version negotiation for pre-v5 apps

jacobkeeler:
The previous expected behavior was to send the maximum protocol version supported by Core in the first StartSessionACK. Now a hard limit of `PROTOCOL_VERSION_4` is set if the proxy doesn't support version 5, as v5 involves a change to the payload structure.

### DIFF
--- a/src/components/protocol_handler/src/protocol_handler_impl.cc
+++ b/src/components/protocol_handler/src/protocol_handler_impl.cc
@@ -243,6 +243,17 @@ void ProtocolHandlerImpl::SendStartSessionAck(
 
   uint8_t maxProtocolVersion = SupportedSDLProtocolVersion();
 
+  const bool proxy_supports_v5_protocol =
+      protocol_version >= PROTOCOL_VERSION_5 ||
+      (ServiceTypeFromByte(service_type) == kRpc &&
+       full_version.majorVersion >= PROTOCOL_VERSION_5);
+
+  // We can't send a V5+ packet if the proxy doesn't support it,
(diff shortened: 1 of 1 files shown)


mrapitis:
@jacobkeeler to be consistent with previous functionality, core should respond with the level supported by the proxy (as opposed to a hard coded value of version 4).  For example, if a proxy connecting supports a max version of protocol level of 3, core should reply back with version 3 messaging.

jacobkeeler:
@mrapitis It doesn't appear that this was ever done for `StartSessionACK`. Here is the current code for master, and it doesn't use the received version from the `StartSession` message at all (`protocolVersion` is just set by checking ini config options to determine the version that Core supports): https://github.com/smartdevicelink/sdl_core/blob/master/src/components/protocol_handler/src/protocol_handler_impl.cc#L185-L196

The old flow, to my understanding, was that Core is supposed to send it's maximum supported version in it's `StartSessionACK`, and the proxy then determines which protocol version to proceed with from then on. The current code reflects this.

All other control messages reciprocate the same protocol version that is sent.

The problem I see with what you are describing is that it doesn't work for the first `StartSession` message (which is defined as a v1 message), because before protocol v5 there was no way of knowing the maximum supported version of the proxy based on that first message.

mrapitis:
@jacobkeeler thanks for the follow up, what you have described makes sense.  I have also tested the fix/version_negotiation_pre_v5 branch on a couple of pre v5 apps on iOS and they are connecting successfully (they used to crash before this commit).

PR #1730, opened 2017-08-21
Fix the build error while DBUS is enabled

chinlin0924:
Change-Id: I17234605c6693c0b34194d05d11185243dd0e510

### DIFF
--- a/src/components/dbus/CMakeLists.txt
+++ b/src/components/dbus/CMakeLists.txt
@@ -74,7 +74,7 @@ add_custom_command(
 )
 
 add_library("DBus" ${SOURCES})
-add_dependecnies("DBus" Utils install-3rd_party_dbus)
+add_dependencies("DBus" Utils install-3rd_party_dbus)
 
 target_link_libraries("DBus" "${LIBRARIES}")
 


chinlin0924:
can you help verify the change when the DBUS is used?
this obviously is a typo, but it exists for a long time.

chinlin0924:
@JackLivio can you help review & verify the change? this is a typo issue

jacobkeeler:
@chinlin0924 Does this change fully fix the build with the `dbus` option enabled? I am fine with this change obviously, I just want to verify that this is all that is necessary to fix the build for that option. 

To my understanding, this option is used for building with the QT HMI, which appears to be quite incredibly outdated (looks like there haven't been any meaningful changes to it since 2015). For this reason, it seems likely to me that there would be even more issues with this build option, given that is hasn't likely been checked any time recently.

chinlin0924:
yes! this is verified with QT HMI option, cmake -DHMI=qt -DHMI2=ON 
in head unit, it is impossible to use web hmi as solution using browser. I wonder why no one fixes this, are they use HTML as HMI? or still no one adopt the SDL core into their product?

chinlin0924:
oh! I think QT_HMI_API.xml need to be updated from HMI_API.xml
I am not sure why we need 2 xml to maintain?

chinlin0924:
you are right. but at least, let 's fix the error step by step.
BTW, is there any component owner to sync this part (QT_HMI_API.xml) and
relevant RPC  & QT/QML files?

2017-08-22 3:58 GMT+08:00 Jacob Keeler <notifications@github.com>:

jacobkeeler:
@chinlin0924 The existing HMIs are really only there to be used for example/demo purposes for now, currently each OEM writes their own HMI (which also allows them to keep their own branding in the car).

jacobkeeler:
Thinking over this more, I believe a lot more deliberation will likely need to be done regarding how to proceed with the QT HMI. I will likely write a proposal soon to determine whether we should drop support for this HMI and remove it from the repository or whether we should update this HMI and begin supporting it again. 

In any case, I can say that the QT HMI has not been supported in years, and is likely just a remnant of before this project became open source. As such, any fixes to this component will likely not be prioritized until we determine how to proceed with it (and if it is supported in the future, it will likely be in a separate repository, as is the case with the other HMIs now).

chinlin0924:
I think QT HMI support is quite important for OEM to integration, if they
cannot be maintained as well as DBUS support. Then this project will not be
practical since head unit cannot use browser to do the HMI rendering for
production.
SmartDeviceLink itself is a good technology, but the reference
implementation took vendor too much resource and time to be a product, and
have too less useful and easier document to quick implementation, even the
road map and target isn't clear. I am worried about it since I have saw the
SDL project for 2 years, and now I synced again to this, and it still
remained the same problems. :(


2017-08-23 1:49 GMT+08:00 Jacob Keeler <notifications@github.com>:

chinlin0924:
it seems a bit late to review, it is 2 month ago. is it possible to speed up the code review & verify, so that the merge will not fail again & again then rebase. I have contributes some chanes, but non of them succeed in time, it takes too long to review.

jacobkeeler:
@chinlin0924 As I said before, this will likely not be prioritized until we determine how to proceed with the QT HMI. Even if we end up supporting it again, I suspect it will be moved into it's own repository, so this fix probably won't be relevant at that time (but I'm not entirely certain, which is why I'm keeping this PR open).

I am planning on writing a proposal for this soon (shortly after SDL Core 4.4.0 is released, which should be at the end of this month), afterward we should have a clearer path forward on this.

chinlin0924:
good to know. thanks!


2017-10-20 2:38 GMT+08:00 Jacob Keeler <notifications@github.com>:

jacobkeeler:
Closing this PR because the issue is fixed in develop, will be in Release 4.5.0

PR #1750, opened 2017-09-07
Remove RC artifact

JackLivio:
Equipment should have been removed from RC baseline


JackLivio:
@LuxoftAKutsan @jacobkeeler

PR #1752, opened 2017-09-08
Replace Functional module library

LuxoftAKutsan:
Replace Adding functional module libraries before application manager library in message helper test
It fixes linkage issue.


LuxoftAKutsan:
@JackLivio @jacobkeeler  please take a look

Issue #1756, opened 2017-09-12
AudioPassThru stops when another app is unregistered


shoamano83:
### Precondition:
1. Check out latest develop branch (commit da4b5766ca3f1fca4caa4b4815073f72dfe37182) and give a build
2. Start SDL Core by ./start.sh
3. Start generic_hmi
4. Open Google Chrome browser and show http://localhost:3000

### Step to reproduce:
1. Connect RPCBuilder app from an Android phone via Wi-Fi
2. Connect another app (let's say "App 2") from another phone via Wi-Fi
3. On the browser, click on App 2 and make it FULL level
4. Let App 2 start AudioPassThru with long maxDuration (e.g. 30 seconds)
5. While AudioPassThru is running, unregister RPCBuilder (by tapping the gear icon on upper right)

### Expected Result:
Audio Pass Thru operation continues because unregistered app is not related to the operation.

### Actual Result:
After RPCBuilder is unregistered, SDL Core stops sending audio data to App 2.

### Tested environment:
Ubuntu 16.04 (x86_64) on a VirtualBox machine
RPC Builder running on Nexus 6 (Android 5.0.1)
"App 2" is a proprietary sample app, running on iPhone 5s (iOS 8.4.1)

jacobkeeler:
Closing with the merge of #1757

PR #1757, opened 2017-09-12
Fix/unregister while audio pass thru

shoamano83:
Fixes https://github.com/smartdevicelink/sdl_core/issues/1756

### Summary of the changes
- In ApplicationManagerImpl::UnregisterApplication(), do not call StopAudioPassThru() if AudioPassThru is not started by the app.
  - ApplicationManager::BeginAudioPassThrough() and ApplicationManager::EndAudioPassThrough() are updated to receive app_id. EndAudioPassThrough() verifies app_id and returns true only when it is same as the ID provided to BeginAudioPassThrough().

### API changes
- This PR updates two public methods in ApplicationManager.
  - Old signatures are marked as "DUPLICATED".

### Risk
- Normal sequence of AudioPassThru operation is affected.

### Testing Plan
- Added an unit test that fails with existing Core and passes with this fix.

### DIFF
--- a/src/components/application_manager/include/application_manager/application_manager_impl.h
+++ b/src/components/application_manager/include/application_manager/application_manager_impl.h
@@ -478,7 +478,7 @@ class ApplicationManagerImpl
   uint32_t GetNextHMICorrelationID() OVERRIDE;
 
   /* @brief Starts audio passthru process
-   * @deprecated Use BeginAudioPassThru(int32_t session_key) instead
+   * @deprecated Use BeginAudioPassThru(uint32_t app_id) instead
    *
    * @return true on success, false if passthru is already in process
    */
@@ -486,14 +486,14 @@ class ApplicationManagerImpl
 
   /**
    * @brief Starts AudioPassThru process by given application
-   * @param session_key Session key of connection for Mobile side
+   * @param app_id ID of the application which starts the process
    * @return true if AudioPassThru can be started, false otherwise
    */
-  bool BeginAudioPassThru(int32_t session_key) OVERRIDE;
+  bool BeginAudioPassThru(uint32_t app_id) OVERRIDE;
 
   /*
    * @brief Finishes already started audio passthru process
-   * @deprecated Use EndAudioPassThru(int32_t application_key) instead
+   * @deprecated Use EndAudioPassThru(uint32_t app_id) instead
    *
    * @return true on success, false if passthru is not active
    */
@@ -501,11 +501,11 @@ class ApplicationManagerImpl
 
   /**
    * @brief Finishes already started AudioPassThru process by given application
-   * @param application_key ID of the application which started the process
+   * @param app_id ID of the application which started the process
    * @return true if AudioPassThru process has been started with given
    * application and thus it can be stopped, false otherwise
    */
-  bool EndAudioPassThru(int32_t application_key) OVERRIDE;
+  bool EndAudioPassThru(uint32_t app_id) OVERRIDE;
 
   /*
    * @brief Retrieves driver distraction state
@@ -1719,7 +1719,7 @@ class ApplicationManagerImpl
   std::map<uint32_t, TimevalStruct> tts_global_properties_app_list_;
 
   bool audio_pass_thru_active_;
-  int32_t audio_pass_thru_app_key_;
+  uint32_t audio_pass_thru_app_id_;
   sync_primitives::Lock audio_pass_thru_lock_;
   sync_primitives::Lock tts_global_properties_app_list_lock_;
   hmi_apis::Common_DriverDistractionState::eType driver_distraction_state_;
--- a/src/components/application_manager/include/application_manager/application_manager_impl.h
+++ b/src/components/application_manager/include/application_manager/application_manager_impl.h
@@ -481,7 +481,8 @@ class ApplicationManagerImpl
    *
    * @return true on success, false if passthru is already in process
    */
-  DEPRECATED bool BeginAudioPassThrough() OVERRIDE;
+  // DEPRECATED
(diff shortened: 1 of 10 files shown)


shoamano83:
@jacobkeeler @JackLivio Please kindly review.

This is not urgent (i.e. hotfix isn't necessary) but I prefer to be included in Core 4.4.
Thanks!

shoamano83:
Updated the destination of this PR to release/4.4.0 branch.

jacobkeeler:
@shoamano83 please rebase on current develop branch (this includes using the new `DEPRECATED` macro, rather than just adding a comment).

shoamano83:
@jacobkeeler Rebasing completed. I can force-push it to this branch if the project accepts it. Otherwise I will merge latest `develop` branch into this feature branch to resolve conflicts.

shoamano83:
Hi @jacobkeeler, @JackLivio I rebased the fixes onto latest `develop` and force-pushed my changes. Please kindly review.
Thanks.

src/components/include/application_manager/application_manager.h
```diff
@@ -393,15 +393,32 @@ class ApplicationManager {
    *
    * @return true on success, false if passthru is already in process
    */
+  // DEPRECATED
   virtual bool BeginAudioPassThrough() = 0;
```
jacobkeeler: In my experience, the `DEPRECATED` macro occasionally causes issues if it is used in both the header and implementation. The best solution I've found is just to use the macro in the highest level (which would be this interface in this case). In any case, using this setup it shouldn't be present in any `.cc` files (which I will need to fix in the current develop).

shoamano83: Thanks for your comment. I reverted the last change and put DEPRECATED macro only in application_manager.h. I also put it in the mock file.

src/components/application_manager/include/application_manager/application_manager_impl.h
```diff
@@ -481,15 +481,32 @@ class ApplicationManagerImpl
    *
    * @return true on success, false if passthru is already in process
    */
+  // DEPRECATED
```
jacobkeeler: Should now be a property `@deprecated` in the method description rather than a standalone comment. The same will need to be done in `application_manager.h`.

src/components/application_manager/include/application_manager/application_manager_impl.h
```diff
@@ -481,15 +481,32 @@ class ApplicationManagerImpl
...
   bool EndAudioPassThrough() OVERRIDE;
 
+  /**
+   * @brief Finishes already started AudioPassThru process by given application
+   * @param application_key ID of the application which started the process
+   * @return true if AudioPassThru process has been started with given
+   * application and thus it can be stopped, false otherwise
+   */
+  bool EndAudioPassThru(int32_t application_key) OVERRIDE;
```
jacobkeeler: Argument name should be unified, perhaps `session_key` in both cases.

src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -3222,9 +3245,8 @@ void ApplicationManagerImpl::UnregisterApplication(
 
   commands_holder_->Clear(app_to_remove);
 
-  if (audio_pass_thru_active_) {
+  if (EndAudioPassThru(app_id)) {
```
jacobkeeler: `app_id` is separate from `connection_key` correct. `connection_key` is used everywhere else.

src/components/application_manager/include/application_manager/application_manager_impl.h
shoamano83: Thanks, done.

src/components/application_manager/include/application_manager/application_manager_impl.h
shoamano83: My intention was to sync with application_manager.h where its `StartAudioPassThruThread()` takes `session_key` and `StopAudioPassThru()` takes `application_key`. Eventually they are same thing and I'm OK to unify the names. Please refer to the next comment as well.

src/components/application_manager/src/application_manager_impl.cc
shoamano83: My understanding is that "app_id" and "connection_key" are eventually same.
- app_id is acquired through ApplicationImpl::app_id()
- The app_id_ value of ApplicationImpl is set through its constructor
- ApplicationImpl instance is created in ApplicationManagerImpl::RegisterApplication(). app_id is generated by session observer's `GetDataOnSessionKey()`.
- Looking at ConnectionHandlerImpl::GetDataOnSessionKey(), connection_key is broken into conn_handle and session_id by `PairFromKey()` then reconstructed into app_id with `KeyFromPair()`.

Also, if I look at perform_audio_pass_thru_request.cc, there are several places where `connection_key()` is used for app_id.

Actually, I have been confused why there are multiple variables "app_id", "connection_key" and "session_key" that eventually mean same thing. Perhaps, at the early stage of the project, they meant for different things? Anyway, my suggestion is to unify the args of `BeginAudioPassThru()` and `EndAudioPassThru()` to "app_id". I think this will clarify that only one app is running Audio Pass Thru at a time. Then, perhaps in audio_pass_thru_request.cc, we put a code something like
```
  uint32_t app_id = connection_key();
  application_manager_.BeginAudioPassThru(app_id);
```
which will hopefully avoid confusion.

Please let me know if it's OK for you.

shoamano83: Thanks @jacobkeeler . I have renamed the args, and also updated the type to `uint32_t` to align with other codes. Hope this is OK.

PR #1759, opened 2017-09-15
Fix SDL crash on exit/unregistering of non-RC application

AKalinich-Luxoft:
SDL plugin manager is receiving notifications from application manager when any application was deactivated or unregistered. This notification will be transfered from plugin manager to RC plugin. However RC plugin processes events from any application so when it try to process event from non-RC application, SDL crashes. RC plugin should process events only from applications with RC functionality.

Also there is another problem when application is unregistering its app_id() is removing from applications list in AM before OnApplicationEvent() call so when some of plugins will try to get shared_ptr using app_id, it will
receive null pointer, however in AM still exists shared_ptr to this app.
So it will be better to pass shared_ptr of application instead of accessing to it every time from plugins using app_id parameter. Moreover, some of plugins could require more information about application in future.

Following changes were done:
- Updated OnApplicationEvent() with application shared_ptr param for plugin manager and all related plugins
- RC plugin now triggers OnApplicationEvent only for events from RC applications
- Fixed possible NULL pointer assignment in Resource Allocation Manager
- Fixed affected mocks and unit tests logic

### DIFF
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -3011,7 +3011,8 @@ void ApplicationManagerImpl::UnregisterApplication(
 
 #ifdef SDL_REMOTE_CONTROL
   plugin_manager_.OnApplicationEvent(
-      functional_modules::ApplicationEvent::kApplicationUnregistered, app_id);
+      functional_modules::ApplicationEvent::kApplicationUnregistered,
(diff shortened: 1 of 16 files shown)


JackLivio:
Closing this pr in favor of keeping #1760. Only need the fix to be merged into the 4.4.0 release candidate for now. This fix will be merged back into develop after 4.4.0 is released.

PR #1760, opened 2017-09-15
Fix SDL crash on exit/unregistering of non-RC application

AKalinich-Luxoft:
SDL plugin manager is receiving notifications from application manager when any application was deactivated or unregistered. This notification will be transfered from plugin manager to RC plugin. However RC plugin processes events from any application so when it try to process event from non-RC application, SDL crashes. RC plugin should process events only from applications with RC functionality.

Also there is another problem when application is unregistering its app_id() is removing from applications list in AM before OnApplicationEvent() call so when some of plugins will try to get shared_ptr using app_id, it will
receive null pointer, however in AM still exists shared_ptr to this app.
So it will be better to pass shared_ptr of application instead of accessing to it every time from plugins using app_id parameter. Moreover, some of plugins could require more information about application in future.

Following changes were done:
- Updated OnApplicationEvent() with application shared_ptr param for plugin manager and all related plugins
- RC plugin now triggers OnApplicationEvent only for events from RC applications
- Fixed possible NULL pointer assignment in Resource Allocation Manager
- Fixed affected mocks and unit tests logic

### DIFF
--- a/src/components/remote_control/src/remote_control_plugin.cc
+++ b/src/components/remote_control/src/remote_control_plugin.cc
@@ -321,9 +321,17 @@ ResourceAllocationManager& RemoteControlPlugin::resource_allocation_manager() {
 }
 
 void RemoteControlPlugin::OnApplicationEvent(
-    functional_modules::ApplicationEvent event, const uint32_t application_id) {
+    functional_modules::ApplicationEvent event,
+    application_manager::ApplicationSharedPtr application) {
   LOG4CXX_AUTO_TRACE(logger_);
-  resource_allocation_manager_.OnApplicationEvent(event, application_id);
+  if (false == service()->IsRemoteControlApplication(application)) {
+    LOG4CXX_DEBUG(logger_,
+                  "Application " << application->app_id()
+                                 << " has no remote control functionality."
+                                 << " Event will be ignored for RC plugin");
+    return;
+  }
+  resource_allocation_manager_.OnApplicationEvent(event, application);
 }
 
 void RemoteControlPlugin::OnPolicyEvent(functional_modules::PolicyEvent event) {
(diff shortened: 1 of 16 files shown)


src/components/remote_control/src/resource_allocation_manager_impl.cc
```diff
@@ -162,15 +162,16 @@ void ResourceAllocationManagerImpl::ProcessApplicationPolicyUpdate() {
 RCAppExtensionPtr ResourceAllocationManagerImpl::GetApplicationExtention(
     application_manager::ApplicationSharedPtr application) {
   LOG4CXX_AUTO_TRACE(logger_);
+
+  RCAppExtensionPtr rc_app_extension;
   if (!application) {
-    return NULL;
+    return rc_app_extension;
```
JackLivio: Seems kind of dangerous returning an uninitialized pointer... Please initialize rc_app_extension to `NULL` so the pointer does not point to random garbage in memory.

JackLivio: @LuxoftAKutsan

AKalinich-Luxoft: @JackLivio rc_app_extension is a default constructed shared pointer, so it is already points to NULL, not to random garbage. Also, on bool() operator call for such empty pointer, it will return false as expected. However if you try to init it with NULL explicitly, SDL will crash with DCHECK error so it is much safer to return just an empty shared pointer. For example you could look at FindApp() function in application_manager_impl.cc

src/components/remote_control/src/remote_control_plugin.cc
```diff
@@ -321,9 +321,17 @@ ResourceAllocationManager& RemoteControlPlugin::resource_allocation_manager() {
...
+    application_manager::ApplicationSharedPtr application) {
   LOG4CXX_AUTO_TRACE(logger_);
-  resource_allocation_manager_.OnApplicationEvent(event, application_id);
+    if (false == service()->IsRemoteControlApplication(application)) {
+      LOG4CXX_DEBUG(logger_,
+                    "Application " << application->app_id()
+                                   << " has no remote control functionality."
+                                   << " Event will be ignored for RC plugin");
+      return;
+    }
```
jacobkeeler: Style issues, run `. tools/infrastructure/check_style.sh --fix` in source folder

AKalinich-Luxoft: @jacobkeeler yes, sorry. Style issues was fixed in https://github.com/smartdevicelink/sdl_core/pull/1760/commits/86974f42574b6a1427192bc2842fbbdf82425841

Issue #1762, opened 2017-09-20
Navigation/phone call/video streaming supported flags are not enabled when HMI capabilities are read from file


shoamano83:
### Description:
When HMI capabilities are read from file, `is_navigation_supported_`, `is_phone_call_supported_` and `is_video_streaming_supported_` flags in `HMICapabilitiesImpl` class are not updated even if these capabilities are available.
Because of this, a mobile app will consider that the head unit does not support navigation/phone call/video streaming features.

I have updated an unit test so that we can detect this issue as test failure.

### Step to reproduce:

1. Please check out my updated unit test. (https://github.com/shoamano83/sdl_core/commit/1f9c8f892086c74542ef993a6f2d2d2e89601033)
2. Give a build and run test ($ make test)

### Expected Result:
`Test #14: application_manager_test` should pass.

### Actual Result:
`Test #14: application_manager_test` fails.

### Tested environment:
Ubuntu 16.04 (x86_64) on a VirtualBox machine

shoamano83:
The fix has been merged to release/4.4.0 branch. I'm closing this issue now.

PR #1763, opened 2017-09-20
Fix/enable system capabilities from file

shoamano83:
Fixes https://github.com/smartdevicelink/sdl_core/issues/1762

### Summary of the changes
- In HMICapabilitiesImpl::load_capabilities_from_file(), update `is_navigation_supported_`, `is_phone_call_supported_` and `is_video_streaming_supported_` flags according to the availability of capability information.

### Risk
- This PR only impacts the flags when HMI capabilities are read from a file.

### Testing Plan
- An unit test is updated to fail with existing implementation and pass with the fix.
- Extra unit tests are added.

### DIFF
--- a/src/components/application_manager/src/hmi_capabilities_impl.cc
+++ b/src/components/application_manager/src/hmi_capabilities_impl.cc
@@ -1068,6 +1068,7 @@ bool HMICapabilitiesImpl::load_capabilities_from_file() {
       }
       if (check_existing_json_member(ui, "systemCapabilities")) {
         Json::Value system_capabilities = ui.get("systemCapabilities", "");
+        set_navigation_supported(false);
         if (check_existing_json_member(system_capabilities,
                                        "navigationCapability")) {
           Json::Value navigation_capability =
@@ -1076,7 +1077,11 @@ bool HMICapabilitiesImpl::load_capabilities_from_file() {
           Formatters::CFormatterJsonBase::jsonValueToObj(
               navigation_capability, navigation_capability_so);
           set_navigation_capability(navigation_capability_so);
+          if (!navigation_capability_so.empty()) {
+            set_navigation_supported(true);
+          }
         }
+        set_phone_call_supported(false);
         if (check_existing_json_member(system_capabilities,
                                        "phoneCapability")) {
           Json::Value phone_capability =
@@ -1085,7 +1090,11 @@ bool HMICapabilitiesImpl::load_capabilities_from_file() {
           Formatters::CFormatterJsonBase::jsonValueToObj(phone_capability,
                                                          phone_capability_so);
           set_phone_capability(phone_capability_so);
+          if (!phone_capability_so.empty()) {
+            set_phone_call_supported(true);
+          }
         }
+        set_video_streaming_supported(false);
         if (check_existing_json_member(system_capabilities,
                                        "videoStreamingCapability")) {
           Json::Value vs_capability =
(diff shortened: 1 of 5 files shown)


shoamano83:
@jacobkeeler @JackLivio Please kindly review.

I think this is necessary for 4.4 release. Please let me know if I should create a PR to release/4.4.0 branch also.

src/components/application_manager/src/hmi_capabilities_impl.cc
```diff
@@ -1076,7 +1077,11 @@ bool HMICapabilitiesImpl::load_capabilities_from_file() {
           Formatters::CFormatterJsonBase::jsonValueToObj(
               navigation_capability, navigation_capability_so);
           set_navigation_capability(navigation_capability_so);
+          if (!navigation_capability_so.empty()) {
+            set_navigation_supported(true);
+          }
         }
+        set_phone_call_supported(false);
```
JackLivio: I think setting these to false  before checking for the capability is unnecessary since those values are already initialized to false by the constructor.

src/components/application_manager/src/hmi_capabilities_impl.cc
```diff
@@ -1085,7 +1090,11 @@ bool HMICapabilitiesImpl::load_capabilities_from_file() {
           Formatters::CFormatterJsonBase::jsonValueToObj(phone_capability,
                                                          phone_capability_so);
           set_phone_capability(phone_capability_so);
+          if (!phone_capability_so.empty()) {
+            set_phone_call_supported(true);
+          }
         }
+        set_video_streaming_supported(false);
```
JackLivio: same here

src/components/application_manager/src/hmi_capabilities_impl.cc
JackLivio: @shoamano83

src/components/application_manager/src/hmi_capabilities_impl.cc
```diff
@@ -1068,6 +1068,7 @@ bool HMICapabilitiesImpl::load_capabilities_from_file() {
       }
       if (check_existing_json_member(ui, "systemCapabilities")) {
         Json::Value system_capabilities = ui.get("systemCapabilities", "");
+        set_navigation_supported(false);
```
JackLivio: Same here

JackLivio:
Otherwise looks good

src/components/application_manager/src/hmi_capabilities_impl.cc
shoamano83: Thanks for your comment.
This call is added to reset the flag when HMICapabilitiesImpl::load_capabilities_from_file() is called more than once. I understand that such situation won’t happen in production, but it does in unit testing. It is called during HMICapabilitiesTest() constructor, then during hmi_capabilities_test->LoadCapabilitiesFromFile() call in each test.

I’ll try if I can remove this double-loading issue in the unit tests.

shoamano83: @JackLivio Fixed with https://github.com/smartdevicelink/sdl_core/pull/1763/commits/fc201151a0fda0eaef0b8cf300afc4a96239255b. Please kindly review again.

src/components/application_manager/src/hmi_capabilities_impl.cc
shoamano83: Fixed with https://github.com/smartdevicelink/sdl_core/pull/1763/commits/fc201151a0fda0eaef0b8cf300afc4a96239255b.

src/components/application_manager/src/hmi_capabilities_impl.cc
shoamano83: Fixed with https://github.com/smartdevicelink/sdl_core/pull/1763/commits/fc201151a0fda0eaef0b8cf300afc4a96239255b.

shoamano83:
Thanks. The destination of this PR is changed to release/4.4.0 branch.

Issue #1773, opened 2017-09-22
RC related RPCs are DISALLOWED whenever an app has specifically defined policies


AKalinich-Luxoft:
**Steps to reproduce**
1. Edit PTU - Add "RemoteControl" group to specific app permissions
2. Edit PTU - Set moduleType to [“RADIO”, “CLIMATE”] for specific app permissions
3. Register mobile app with REMOTE_CONTROL HMI type and bring it to FULL
4. Send SetInteriorVehicleData(CLIMATE)
5. Expected DISALLOWED - OK
6. Perform PTU sequence using PT edited in steps 1-2
7. After successful PTU send SetInteriorVehicleData(CLIMATE) again

**Expected result**
SDL allows SetInteriorVehicleData() request and sends it to HMI

**Actual result**
SDL responds with DISALLOWD and does not send request to HMI

This same scenario ends with SUCCESS when no explicit app policy is specified and all of these steps are performed on the default policies. Adding explicit policies for an app causes this issue.

jacobkeeler:
Closing with the merge of #1774

PR #1774, opened 2017-09-22
Fix DISALLOWED response when app has specifically defined policies

AKalinich-Luxoft:
This problem mostly related to RC functionality. When SDL receives any RC request, it checks is sending app have `REMOTE_CONTROL` HMI type. If app does not have such HMI type, SDL disallows such RPC for it.

The problem was in `CacheManager::GetHMITypes()` function, which returns `appHMIType` array even if it was not initialized(param is not present in policy) so SDL tries to find HMI type in an empty array and that always evaluates to false so SDL decides that current app is not RC app. If `appHMITypes` is not specified in policies, SDL should search HMI type in array of HMI types, provided by application on its registration.
This check works correctly if application has default policies.

To fix that problem there was added check in `GetHMITypes()` to avoid returning of uninitialized array.

Fixes https://github.com/smartdevicelink/sdl_core/issues/1773


AByzhynar:
Rebuild required

AByzhynar:
Rebuild required

PR #1776, opened 2017-09-22
Require nav or projection app for SendHapticData

JackLivio:
@jacobkeeler


src/components/interfaces/MOBILE_API.xml
```diff
@@ -5450,6 +5450,10 @@
       <element name="GENERIC_ERROR"/>
       <element name="DISALLOWED"/>
     </param> 
+
+    <param name="info" type="String" maxlength="1000" mandatory="false" platform="documentation">
+      <description>Provides additional human readable info regarding the result.</description>
+    </param>
```
jacobkeeler: This will be added as part of https://github.com/smartdevicelink/sdl_core/pull/1775, it can be removed from this PR

Issue #1777, opened 2017-09-25
Core generates a lot of logs during streaming


shoamano83:
Currently SDL Core generates quite a few log lines each time it receives a frame from mobile. Because of this, we will get a large amount of logs during video streaming.
- Log files can be flushed away quickly.
- If the system has slow log output, log messages will queued up in memory and might cause system's RAM to exhaust.

jacobkeeler:
Closing with the merge of #1778

PR #1778, opened 2017-09-25
fix: reduce logs during streaming

shoamano83:
Fixes https://github.com/smartdevicelink/sdl_core/issues/1777.

I thought that removing particular logs completely will confuse developers and hence it will be a bad idea. So instead of eliminating logs, log levels are decreased.
Head unit systems with video streaming feature can set log output level to DEBUG (or higher) to get rid of the logs.


shoamano83:
@JackLivio @jacobkeeler Please kindly review.

JackLivio:
@shoamano83 I like this change. Could you please retarget for the 4.4.0 release candidate branch? I would like to put this into the next release since most of the new features revolve around video streaming.

JackLivio:
@shoamano83 We are still discussing if these logs are needed at all, or if trace is the best log level for these logs.

shoamano83:
Thanks for your comment. Let me know if you want to remove some of the logs, I will update PR accordingly.

jacobkeeler (approved):

Issue #1772, opened 2017-09-22
Default app policies are never updated after a PTU


AKalinich-Luxoft:
## Occurence:
Always

## Steps to reproduce:
1. Make sure SDL is built with default flags
2. Start SDL and HMI
3. Connect mobile device
4. Register mobile application
5. Perform Policy Table Update sequence: add new functional group for default app with some additional RPC enabled (which was disabled previously), e.g. ‘GetVehicleData’. This RPC is allowed for FULL HMI level.
6. Activate application in order to switch it to FULL HMI level
7. Send ‘GetVehicleData’ RPC

## Expected result:
RPC is processed successfully

## Actual result:
SDL responds with DISALLOWED and not sends request to HMI

## Test script:
[1772_update_default_section.lua](https://github.com/smartdevicelink/sdl_atf_test_scripts/blob/develop/test_scripts/Defects/4_5/1772_update_default_section.lua)

jacobkeeler:
Needs to be fixed for PROPRIETARY mode

1. Add Location-1 to the functional_group list in the default app policy on the PTU server
2. Connect test app
3. Perform PTU (app id is not present in update)
4. Bring test app to FULL
5. Send GetVehicleData(gps:true)
6. Get DISALLOWED response

The test app used is left with an outdated version of the `default` permissions in this case. Any new app that is connected will have the updated permissions.

jacobkeeler:
App policies in
1. Local PT:
```
        "app_policies": {
            "default": {
                "keep_context": false,
                "steal_focus": false,
                "priority": "NONE",
                "default_hmi": "NONE",
                "groups": ["Base-4", "RemoteControl"],
		"moduleType": [
			"RADIO",
			"CLIMATE"
		]
            },
            "device": {
                "keep_context": false,
                "steal_focus": false,
                "priority": "NONE",
                "default_hmi": "NONE",
                "groups": ["DataConsent-2"]
            },
            "pre_DataConsent": {
                "keep_context": false,
                "steal_focus": false,
                "priority": "NONE",
                "default_hmi": "NONE",
                "groups": ["BaseBeforeDataConsent"]
            }
        }
```
2. Policy Server:
```
        "app_policies": {
            "default": {
                "keep_context": false,
                "steal_focus": false,
                "priority": "NONE",
                "default_hmi": "NONE",
                "groups": ["Base-4", "Location-1"]
            },
            "device": {
                "keep_context": false,
                "steal_focus": false,
[...]

PR #1779, opened 2017-09-25
Fix PTU applying for "default" app policies

AKalinich-Luxoft:
This issue is reproduced on EXTERNAL_PROPRIETARY flow only.

There was a problem that "default" policy section was not updated after PTU with changes in this section.
Also applications which have "default" policies in PTU was not updated too.

To fix this issue there was removed code in `CacheManager`, which incorrectly assigns default policies to apps with "default" policies.

Also there was a redundant code because "default" policies is unwrapped in PTU and replaced with actual default policies, so all specific application policies is already have actual new default policy permissions. In this case it is correct to assign to every app his own policies from PTU.

Also there was updated logic in `ProcessAppPolicyCheckResults()` to perform all needed actions once per app, because its possible that results could contain few results for one `app_id`.

Fixes https://github.com/SmartDeviceLink/sdl_core/issues/1772

### DIFF
--- a/src/components/policy/policy_external/include/policy/policy_helper.h
+++ b/src/components/policy/policy_external/include/policy/policy_helper.h
@@ -199,6 +199,23 @@ struct CheckAppPolicy {
   CheckAppPolicyResults& out_results_;
 };
 
+/**
+ * @brief Helper struct for filling actions to be done for processed application
+ * using CheckAppPolicyResults data as a source
+ */
+struct FillActionsForAppPolicies {
+  FillActionsForAppPolicies(
+      ApplicationsPoliciesActions& actions,
+      const policy_table::ApplicationPolicies& app_policies)
+      : actions_(actions), app_policies_(app_policies) {}
+
+  void operator()(const policy::CheckAppPolicyResults::value_type& value);
+
+ private:
+  ApplicationsPoliciesActions& actions_;
+  const policy_table::ApplicationPolicies& app_policies_;
+};
+
 /*
  * @brief Fill permissions data with merged rpc permissions for hmi levels and
  * parameters
--- a/src/components/policy/policy_external/src/cache_manager.cc
+++ b/src/components/policy/policy_external/src/cache_manager.cc
@@ -630,7 +630,7 @@ void CacheManager::ProcessUpdate(
       *(initial_policy_iter->second.RequestType);
 
   const std::string& app_id = initial_policy_iter->first;
-  RequestTypes merged_pt_request_types;
+  bool update_request_types = true;
 
   if (app_id == kDefaultId || app_id == kPreDataConsentId) {
     if (new_request_types.is_omitted()) {
@@ -638,25 +638,28 @@ void CacheManager::ProcessUpdate(
                    "Application " << app_id
                                   << " has omitted RequestTypes."
                                      " Previous values will be kept.");
-      return;
-    }
-    if (new_request_types.empty()) {
+      update_request_types = false;
+    } else if (new_request_types.empty()) {
       if (new_request_types.is_cleaned_up()) {
         LOG4CXX_INFO(logger_,
                      "Application " << app_id
                                     << " has cleaned up all values."
                                        " Previous values will be kept.");
-        return;
+        update_request_types = false;
+      } else {
+        LOG4CXX_INFO(logger_,
+                     "Application " << app_id
+                                    << " has empty RequestTypes."
+                                       " Any parameter will be allowed.");
       }
-      LOG4CXX_INFO(logger_,
-                   "Application " << app_id
-                                  << " has empty RequestTypes."
-                                     " Any parameter will be allowed.");
     }
-    merged_pt_request_types = new_request_types;
-  } else {
-    merged_pt_request_types = new_request_types;
   }
+
+  const RequestTypes merged_pt_request_types =
+      update_request_types
+          ? new_request_types
+          : *(pt_->policy_table.app_policies_section.apps[app_id].RequestType);
+
   pt_->policy_table.app_policies_section.apps[app_id] =
       initial_policy_iter->second;
   *(pt_->policy_table.app_policies_section.apps[app_id].RequestType) =
@@ -682,15 +685,6 @@ bool CacheManager::ApplyUpdate(const policy_table::Table& update_pt) {
       pt_->policy_table.app_policies_section.apps[iter->first].set_to_null();
       pt_->policy_table.app_policies_section.apps[iter->first].set_to_string(
           "");
-    } else if (policy::kDefaultId == (iter->second).get_string()) {
-      policy_table::ApplicationPolicies::const_iterator iter_default =
-          update_pt.policy_table.app_policies_section.apps.find(kDefaultId);
-      if (update_pt.policy_table.app_policies_section.apps.end() ==
-          iter_default) {
-        LOG4CXX_ERROR(logger_, "The default section was not found in PTU");
-        continue;
-      }
-      ProcessUpdate(iter_default);
     } else {
       ProcessUpdate(iter);
     }
--- a/src/components/policy/policy_external/src/policy_helper.cc
+++ b/src/components/policy/policy_external/src/policy_helper.cc
@@ -484,6 +484,41 @@ bool CheckAppPolicy::IsRequestTypeChanged(
   return diff.size();
 }
 
+void FillActionsForAppPolicies::operator()(
+    const policy::CheckAppPolicyResults::value_type& value) {
+  const std::string app_id = value.first;
+  const policy_table::ApplicationPolicies::const_iterator app_policy =
+      app_policies_.find(app_id);
+
+  if (app_policies_.end() == app_policy) {
+    return;
+  }
+
+  if (IsPredefinedApp(*app_policy)) {
+    return;
+  }
+
+  switch (value.second) {
+    case RESULT_APP_REVOKED:
+    case RESULT_NICKNAME_MISMATCH:
+      actions_[app_id].is_notify_system = true;
+      return;
+    case RESULT_CONSENT_NEEDED:
+    case RESULT_PERMISSIONS_REVOKED_AND_CONSENT_NEEDED:
+      actions_[app_id].is_consent_needed = true;
+      break;
+    case RESULT_CONSENT_NOT_REQIURED:
+    case RESULT_PERMISSIONS_REVOKED:
+    case RESULT_REQUEST_TYPE_CHANGED:
+      break;
+    case RESULT_NO_CHANGES:
+    default:
+      return;
+  }
+  actions_[app_id].is_notify_system = true;
+  actions_[app_id].is_send_permissions_to_app = true;
+}
+
 FillNotificationData::FillNotificationData(Permissions& data,
                                            GroupConsent group_state,
                                            GroupConsent undefined_group_consent,
--- a/src/components/policy/policy_external/src/policy_manager_impl.cc
+++ b/src/components/policy/policy_external/src/policy_manager_impl.cc
@@ -385,53 +385,45 @@ CheckAppPolicyResults PolicyManagerImpl::CheckPermissionsChanges(
 void PolicyManagerImpl::ProcessAppPolicyCheckResults(
     const CheckAppPolicyResults& results,
     const policy_table::ApplicationPolicies& app_policies) {
-  CheckAppPolicyResults::const_iterator it_results = results.begin();
+  ApplicationsPoliciesActions actions_for_apps_policies;
+  FillActionsForAppPolicies filler(actions_for_apps_policies, app_policies);
 
-  for (; results.end() != it_results; ++it_results) {
-    const policy_table::ApplicationPolicies::const_iterator app_policy =
-        app_policies.find(it_results->first);
+  std::for_each(results.begin(), results.end(), filler);
 
+  ProcessActionsForAppPolicies(actions_for_apps_policies, app_policies);
+}
+
+void PolicyManagerImpl::ProcessActionsForAppPolicies(
+    const ApplicationsPoliciesActions& actions,
+    const policy_table::ApplicationPolicies& app_policies) {
+  ApplicationsPoliciesActions::const_iterator it_actions = actions.begin();
+  for (; it_actions != actions.end(); ++it_actions) {
+    policy_table::ApplicationPolicies::const_iterator app_policy =
+        app_policies.find(it_actions->first);
     if (app_policies.end() == app_policy) {
       continue;
     }
 
-    if (IsPredefinedApp(*app_policy)) {
-      continue;
-    }
+    if (it_actions->second.is_consent_needed) {
+      // Post-check after ExternalConsent consent changes
+      const std::string& policy_app_id = app_policy->first;
+      if (!IsConsentNeeded(policy_app_id)) {
+        sync_primitives::AutoLock lock(app_permissions_diff_lock_);
 
-    switch (it_results->second) {
-      case RESULT_NO_CHANGES:
-        continue;
-      case RESULT_APP_REVOKED:
-        NotifySystem(*app_policy);
-        continue;
-      case RESULT_NICKNAME_MISMATCH:
-        NotifySystem(*app_policy);
-        continue;
-      case RESULT_CONSENT_NEEDED:
-      case RESULT_PERMISSIONS_REVOKED_AND_CONSENT_NEEDED: {
-        // Post-check after ExternalConsent consent changes
-        const std::string policy_app_id = app_policy->first;
-        if (!IsConsentNeeded(policy_app_id)) {
-          sync_primitives::AutoLock lock(app_permissions_diff_lock_);
+        PendingPermissions::iterator app_id_diff =
+            app_permissions_diff_.find(policy_app_id);
 
-          PendingPermissions::iterator app_id_diff =
-              app_permissions_diff_.find(policy_app_id);
-
-          if (app_permissions_diff_.end() != app_id_diff) {
-            app_id_diff->second.appPermissionsConsentNeeded = false;
-          }
+        if (app_permissions_diff_.end() != app_id_diff) {
+          app_id_diff->second.appPermissionsConsentNeeded = false;
         }
-      } break;
-      case RESULT_CONSENT_NOT_REQIURED:
-      case RESULT_PERMISSIONS_REVOKED:
-      case RESULT_REQUEST_TYPE_CHANGED:
-        break;
-      default:
-        continue;
+      }
     }
-    NotifySystem(*app_policy);
-    SendPermissionsToApp(*app_policy);
+    if (it_actions->second.is_notify_system) {
+      NotifySystem(*app_policy);
+    }
+    if (it_actions->second.is_send_permissions_to_app) {
+      SendPermissionsToApp(*app_policy);
+    }
   }
 }
 
--- a/src/components/policy/policy_external/include/policy/policy_manager_impl.h
+++ b/src/components/policy/policy_external/include/policy/policy_manager_impl.h
@@ -719,14 +719,28 @@ class PolicyManagerImpl : public PolicyManager {
 
   /**
    * @brief Processes results from policy table update analysis done by
-   * CheckPermissionsChanges() by sending OnPermissionChange and
-   * OnAppPermissionChanged notifications
+   * CheckPermissionsChanges() by filling ApplicationsPoliciesActions struct
+   * with actions which should be done for every application and passes them to
+   * ProcessActionsForAppPolicies()
    * @param results Collection of per-application results
    * @param app_policies Reference to updated application policies section as
    * a data source for generating notifications data
    */
   void ProcessAppPolicyCheckResults(
       const CheckAppPolicyResults& results,
+      const policy_table::ApplicationPolicies& app_policies);
+
+  /**
+   * @brief Processes actions filled in ProcessAppPolicyCheckResults() for every
+   * application by sending OnPermissionChange and OnAppPermissionChanged
+   * notifications and by checking consent where it needed
+   * @param actions Reference to map with actions to be done or not for every
+   * application
+   * @param app_policies Reference to updated application policies section as
+   * a data source for generating notifications data
+   */
+  void ProcessActionsForAppPolicies(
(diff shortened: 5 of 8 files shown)


src/components/policy/policy_external/include/policy/policy_helper.h
```diff
@@ -178,6 +178,14 @@ struct CheckAppPolicy {
    */
   void AddResult(const std::string& app_id, PermissionsCheckResult result);
   /**
+   * @brief Adds result of check of applications related to current application
```
LuxoftAKutsan: @AKalinich-Luxoft please describe this function better. It is not clear what does in means - related application to application.

src/components/policy/policy_external/include/policy/policy_helper.h
```diff
@@ -178,6 +178,14 @@ struct CheckAppPolicy {
...
+   * policy with updated one
+   * @param app_id Application id
+   * @param result Result value
+   */
+  void AddRelatedAppsResult(const std::string& app_id,
+                            PermissionsCheckResult result);
```
LuxoftAKutsan: @AKalinich-Luxoft is it output parameter? So rename it

src/components/policy/policy_external/src/cache_manager.cc
```diff
@@ -705,6 +708,7 @@ bool CacheManager::ApplyUpdate(const policy_table::Table& update_pt) {
   pt_->policy_table.consumer_friendly_messages.assign_if_valid(
       update_pt.policy_table.consumer_friendly_messages);
 
+  UnwrapAppPolicies(pt_->policy_table.app_policies_section.apps);
```
dev-gh: @AKalinich-Luxoft 
I didn't catch why you call this method again in several places. AFAIK the purpose of this method is to replace permissions records in PTU (not in current PT) like
```
"app_id" : "default"
```
with the content of "default" permissions section.
So the only intention was to call it once PTU is received and before processing updated permissions i.e. current PT should not be 'unwrapped' again.

src/components/policy/policy_external/src/policy_helper.cc
```diff
@@ -342,6 +359,8 @@ bool CheckAppPolicy::operator()(const AppPoliciesValueType& app_policy) {
   if (!IsPredefinedApp(app_policy)) {
     SetPendingPermissions(app_policy, result);
     AddResult(app_id, result);
+  } else {
+    AddRelatedAppsResult(app_id, result);
```
dev-gh: @AKalinich-Luxoft 
This part is inside bool operator()(<application_policy>&) which basically iterates over all applications policies. Why you're iterating over all applications policies (AddRelatedAppsResult call) inside another iteration over same data?
Why you can't reuse  AddResult()?

src/components/policy/policy_external/include/policy/policy_helper.h
AKalinich-Luxoft: @LuxoftAKutsan description was updated in https://github.com/smartdevicelink/sdl_core/pull/1779/commits/c40c209107a48effdaf494913f9e62abd9d92334

src/components/policy/policy_external/include/policy/policy_helper.h
AKalinich-Luxoft: @LuxoftAKutsan param was renamed as well as whole function name. See https://github.com/smartdevicelink/sdl_core/pull/1779/commits/c40c209107a48effdaf494913f9e62abd9d92334

src/components/policy/policy_external/src/cache_manager.cc
AKalinich-Luxoft: @dev-gh as I found out current PT should be re-unwrapped again after PTU applying. Let's imagine such case:
1. SDL is started with preloaded pt
2. App1 is registered and gets "default" policy permissions. Note that ApplicationPolicies class wrapped in Stringifyable class and its predefined_string_ is also set to "default"
3. App1 initiates PTU where only default policy permissions were changed (for example was added new functional group)
4. As you said, before processing updated permissions, policy manager calls UnwrapAppPolicies. This function replaces application policies with default policies for applications which have "default" predefined_string string (see step 2)
5. Before applying changes in PT we have "App1" with copied default permissions and "default" with its own permissions.
6. All PTU changes was applied. Now "default" has updated permissions, however "App1" still have copy of default permissions before update, i.e. default permission changes will not affect App1. This changes will take effect for App1 only after SDL restart. 
7. I think this is not correct behaviour so I add UnwrapAppPolicies call after updates processing, so this call re-assigns updated default policy permissions for specific applications with default permissions, so with this rewrapping it works as expected and updated permissions will take an effect immediately for App1.

This is a reason why I use this function here. I will be thankful if you could suggest better solution here.

src/components/policy/policy_external/src/policy_helper.cc
AKalinich-Luxoft: @dev-gh because in operator() we check permission changes only for current app_id. So in case when you will receive PTU where only "default" policy was changed, you will get out_results like:
default: RESULT_CONSENT_NOT_REQIURED
However if application with App1 was registered and using default policies, it will not receive OnPermissionChange notification later in ProcessAppPolicyCheckResults(), because its permissions were changed indirectly.
This AddRelatedAppsResult performs additional iteration over all applications in case some of predefined application policies permissions were changed, in order to add applications, which permissions were changed indirectly, i.e:
default: RESULT_CONSENT_NOT_REQIURED
App1: RESULT_CONSENT_NOT_REQIURED
In this case App1 will receive OnPermissionChange notification as expected.

src/components/policy/policy_external/src/cache_manager.cc
dev-gh: @AKalinich-Luxoft 
I think p.6 is incorrect, I mean "Now "default" has updated permissions, however "App1" still have copy of default permissions before update, i.e. default permission changes will not affect App1."
If it is so, then there is some other issue with applying of PTU.

Per my understanding flow must be following:
- pp. 1-3 are okay
 - p. 4 - clarification: replacement is done for PTU, not for current PT, so during checking of permissions difference App1 should have old 'default' permissions in PT but new 'default' in PTU 
 - p. 5 - okay
 - p. 6 - during applying of PTU App1 policy must be replaced with the one from PTU i.e.  groups must be updated as well so I don't quite understand why restart is required

From the code in 'develop' branch I see that in case new groups detected and no user consents are required RESULT_CONSENT_NOT_REQIURED code is returned during difference check. Looks like this code doesn't trigger OnPermissionsChange sending and I think it may be related to issue cause.

src/components/policy/policy_external/src/policy_manager_impl.cc
```diff
@@ -139,6 +139,65 @@ struct ConsentsUpdater
...
+        return;
+      case RESULT_APP_REVOKED:
+      case RESULT_NICKNAME_MISMATCH:
+        is_notify_system_ = true;
+        return;
+      case RESULT_CONSENT_NEEDED:
+      case RESULT_PERMISSIONS_REVOKED_AND_CONSENT_NEEDED: {
+        is_consent_needed_ = true;
+        break;
+        case RESULT_CONSENT_NOT_REQIURED:
```
AByzhynar: @AKalinich-Luxoft Fix indents

src/components/policy/policy_external/src/policy_manager_impl.cc
```diff
@@ -139,6 +139,65 @@ struct ConsentsUpdater
...
+          break;
+        default:
+          return;
+      }
+    }
+    is_notify_system_ = true;
+    is_send_permissions_ = true;
+  }
+
+  const bool is_notify_system() const {
```
AByzhynar: @AKalinich-Luxoft What is the sense of making `const` on return by value?

src/components/policy/policy_external/src/policy_manager_impl.cc
```diff
@@ -139,6 +139,65 @@ struct ConsentsUpdater
...
+    using namespace policy;
+    switch (value.second) {
+      case RESULT_NO_CHANGES:
+        return;
+      case RESULT_APP_REVOKED:
+      case RESULT_NICKNAME_MISMATCH:
+        is_notify_system_ = true;
+        return;
+      case RESULT_CONSENT_NEEDED:
+      case RESULT_PERMISSIONS_REVOKED_AND_CONSENT_NEEDED: {
```
AByzhynar: @AKalinich-Luxoft What for `{`?

src/components/policy/policy_external/src/policy_manager_impl.cc
```diff
@@ -385,53 +446,38 @@ CheckAppPolicyResults PolicyManagerImpl::CheckPermissionsChanges(
...
+  for (; iter != iter_end; ++iter) {
+    if (IsPredefinedApp(*iter)) {
       continue;
     }
 
-    if (IsPredefinedApp(*app_policy)) {
-      continue;
-    }
+    AppPolicyResultsChecker checker(iter);
+    std::for_each(results.begin(), results.end(), std::ref(checker));
```
AByzhynar: @AKalinich-Luxoft AFAIK `std::ref` is C++11 feature

src/components/policy/policy_external/src/policy_helper.cc
dev-gh: @AKalinich-Luxoft 
As I responded in my previous comment I suspect there is an issue with the detecting of 'default' groups changes and further results processing. So I think there is no need for additional checks but instead, we have to properly detect changes of 'default' section before applying of PTU, put correct result for that and send the notification. I'll try to check it on my side and let you know.

src/components/policy/policy_external/src/policy_manager_impl.cc
dev-gh: @AByzhynar 
I don't see any sense to avoid of using C++11 as the appropriate proposal was accepted already.

AByzhynar: @dev-gh I don't know about such acception

src/components/policy/policy_external/src/cache_manager.cc
AKalinich-Luxoft: @dev-gh p.4 - what if we send PTU, where section with App1 will not be present? How to check permission changes for App1 in this case?
regarding p.6 - it is okay only if App1 is present at least in PTU, but what if it will not be present there as in case above?

AKalinich-Luxoft: @dev-gh restart of SDL is required because in local PT permissions will not be updated for App1 with default policy after PTU. However if SDL will be restarted, it will load local PT from storage, and when we register App1 again, it will have updated default permissions as expected.

dev-gh: @AKalinich-Luxoft 
If SDL sends snapshot to backend with some application id then expected to get back specific permissions or "default". Otherwise it is wrong PTU.

src/components/policy/policy_external/src/policy_helper.cc
AKalinich-Luxoft: @dev-gh I suppose we have to add new default permissions for App1 to PTU (if it is absent there) before its applying? In this case all this sequence should work correctly.

src/components/policy/policy_external/src/cache_manager.cc
AKalinich-Luxoft: @dev-gh ok, in this case I will check this issue with PTU with "default" for some application id. Probably some of this logic is redundant in this case.

AKalinich-Luxoft: @dev-gh pull request was updated and redundant logic were removed. Please review again 89ceac249b54ec02c042865b27f4af8b2363d450

src/components/policy/policy_external/src/policy_helper.cc
AKalinich-Luxoft: @dev-gh not actual after 89ceac249b54ec02c042865b27f4af8b2363d450

src/components/policy/policy_external/src/policy_manager_impl.cc
AKalinich-Luxoft: @AByzhynar this code was removed, please review 89ceac249b54ec02c042865b27f4af8b2363d450

src/components/policy/policy_external/src/policy_manager_impl.cc
AKalinich-Luxoft: @AByzhynar this code was removed, please review 89ceac249b54ec02c042865b27f4af8b2363d450

src/components/policy/policy_external/src/policy_manager_impl.cc
AKalinich-Luxoft: @AByzhynar this code was removed, please review 89ceac249b54ec02c042865b27f4af8b2363d450

src/components/policy/policy_external/src/policy_manager_impl.cc
AKalinich-Luxoft: @AByzhynar this code was removed, please review 89ceac249b54ec02c042865b27f4af8b2363d450

dev-gh:
@AKalinich-Luxoft 
Update description according to latest changes

AKalinich-Luxoft:
@dev-gh Done

src/components/policy/policy_external/src/policy_manager_impl.cc
```diff
@@ -139,6 +139,23 @@ struct ConsentsUpdater
...
+ * application policies iterator
+ */
+struct ActionsForApplicationPolicies {
+  ActionsForApplicationPolicies()
+      : app_policies(NULL)
+      , is_notify_system(false)
+      , is_send_permissions_to_app(false)
+      , is_consent_needed(false) {}
+
+  policy_table::ApplicationPolicies::const_iterator app_policies;
```
LuxoftAKutsan: @AKalinich-Luxoft Why iterator? Why not reference to object? Iterator may be invalidated.

src/components/policy/policy_external/src/policy_manager_impl.cc
```diff
@@ -399,39 +421,58 @@ void PolicyManagerImpl::ProcessAppPolicyCheckResults(
...
+  // Perform all needed actions once per app_id
+  for (ActionsForApplications::const_iterator it = actions_for_apps.begin();
+       it != actions_for_apps.end();
+       ++it) {
+    policy_table::ApplicationPolicies::const_iterator app_policy =
+        it->second.app_policies;
+
+    if (it->second.is_consent_needed) {
+      // Post-check after ExternalConsent consent changes
+      const std::string policy_app_id = app_policy->first;
```
LuxoftAKutsan: @AKalinich-Luxoft reference

src/components/policy/policy_external/src/policy_manager_impl.cc
```diff
@@ -399,39 +421,58 @@ void PolicyManagerImpl::ProcessAppPolicyCheckResults(
...
+  }
+
+  // Perform all needed actions once per app_id
+  for (ActionsForApplications::const_iterator it = actions_for_apps.begin();
+       it != actions_for_apps.end();
+       ++it) {
+    policy_table::ApplicationPolicies::const_iterator app_policy =
+        it->second.app_policies;
+
+    if (it->second.is_consent_needed) {
```
LuxoftAKutsan: @AKalinich-Luxoft a lot of not understandable names and a lot of code in one function. Please perform decomposition and variables renaming

src/components/policy/policy_external/src/policy_manager_impl.cc
```diff
@@ -387,9 +404,14 @@ void PolicyManagerImpl::ProcessAppPolicyCheckResults(
     const policy_table::ApplicationPolicies& app_policies) {
   CheckAppPolicyResults::const_iterator it_results = results.begin();
 
+  typedef std::map<std::string, ActionsForApplicationPolicies>
+      ActionsForApplications;
+  ActionsForApplications actions_for_apps;
+
   for (; results.end() != it_results; ++it_results) {
+    const std::string app_id = it_results->first;
```
LuxoftAKutsan: @AKalinich-Luxoft ref

src/components/policy/policy_external/src/policy_manager_impl.cc
AKalinich-Luxoft: @LuxoftAKutsan this field was removed after decomposition and some updates. See https://github.com/smartdevicelink/sdl_core/pull/1779/commits/bfdcdd0f21dc83ffb8a607060d3392cad10406a4

src/components/policy/policy_external/src/policy_manager_impl.cc
AKalinich-Luxoft: @LuxoftAKutsan was removed after decomposition and moving this staff to separate function. See https://github.com/smartdevicelink/sdl_core/pull/1779/commits/bfdcdd0f21dc83ffb8a607060d3392cad10406a4

src/components/policy/policy_external/src/policy_manager_impl.cc
AKalinich-Luxoft: @LuxoftAKutsan function was separated on two parts, some variables were renamed. See https://github.com/smartdevicelink/sdl_core/pull/1779/commits/bfdcdd0f21dc83ffb8a607060d3392cad10406a4

src/components/policy/policy_external/src/policy_manager_impl.cc
AKalinich-Luxoft: @LuxoftAKutsan fixed in https://github.com/smartdevicelink/sdl_core/pull/1779/commits/bfdcdd0f21dc83ffb8a607060d3392cad10406a4

jacobkeeler:
This issue is also consistently reproducible with `PROPRIETARY` policy mode:

1. Add `Location-1` to the `functional_group` list in the `default` app policy on the PTU server
2. Connect test app
3. Perform PTU
4. Bring test app to FULL
5. Send GetVehicleData(gps:true)
6. Get DISALLOWED response

This PR does not fix this case

jacobkeeler:
I have verified that this does fix the issue for `EXTERNAL_PROPRIETARY`, however 👍

AKalinich-Luxoft:
@jacobkeeler can't reproduce this issue using your steps on commit https://github.com/smartdevicelink/sdl_core/pull/1779/commits/7e1d12a8bddd9180d3c43e41911c6620b1372739
Could you please provide content of your "app_policies" section in PTU file?

Also please note that PTU should contain app_id of testing app as mentioned here:
https://github.com/smartdevicelink/sdl_core/pull/1779#discussion_r141010459

jacobkeeler:
@AKalinich-Luxoft 
App policies in
1. Local PT:
```
        "app_policies": {
            "default": {
                "keep_context": false,
                "steal_focus": false,
                "priority": "NONE",
                "default_hmi": "NONE",
                "groups": ["Base-4", "RemoteControl"],
		"moduleType": [
			"RADIO",
			"CLIMATE"
		]
            },
            "device": {
                "keep_context": false,
                "steal_focus": false,
                "priority": "NONE",
                "default_hmi": "NONE",
                "groups": ["DataConsent-2"]
            },
            "pre_DataConsent": {
                "keep_context": false,
                "steal_focus": false,
                "priority": "NONE",
                "default_hmi": "NONE",
                "groups": ["BaseBeforeDataConsent"]
            }
```
2. Policy Server:
```
        "app_policies": {
            "default": {
                "keep_context": false,
                "steal_focus": false,
                "priority": "NONE",
                "default_hmi": "NONE",
                "groups": ["Base-4", "Location-1"]
            },
            "device": {
                "keep_context": false,
                "steal_focus": false,
                "priority": "NONE",
                "default_hmi": "NONE",
                "groups": ["DataConsent-2"]
            },
            "pre_DataConsent": {
                "keep_context": false,
                "steal_focus": false,
                "priority": "NONE",
                "default_hmi": "NONE",
                "groups": ["BaseBeforeDataConsent"]
            }
        }
```

jacobkeeler:
Alright, I think I understand now, the fact that the policy server is not sending the application policies in the first place is an error, and the PTU is invalid if it doesn't include the app IDs that were sent in the snapshot. In my case the server should send:
```
        "app_policies": {
            "default": {
                "keep_context": false,
                "steal_focus": false,
                "priority": "NONE",
                "default_hmi": "NONE",
                "groups": ["Base-4", "Location-1"]
            },
            "device": {
                "keep_context": false,
                "steal_focus": false,
                "priority": "NONE",
                "default_hmi": "NONE",
                "groups": ["DataConsent-2"]
            },
            "pre_DataConsent": {
                "keep_context": false,
                "steal_focus": false,
                "priority": "NONE",
                "default_hmi": "NONE",
                "groups": ["BaseBeforeDataConsent"]
            },
            "my_app_id": "default"
        }
```
My only issue with this is that the PTU is not considered by Core to be invalid if this field is omitted. I was using (an admittedly outdated version of) the [policy server](https://github.com/smartdevicelink/sdl_server/commits/master), which just sent a static json file as the PTU, and Core never considered that invalid.

LuxoftAKutsan:
@AnastasiyaBritanova @KhrystynaDubovyk  please clarify is there some SDL requirement for that use case : 
> the fact that the policy server is not sending the application policies in the first place is an error, and the PTU is invalid if it doesn't include the app IDs that were sent in the snapshot.
?

KhrystynaDubovyk:
> @AnastasiyaBritanova @KhrystynaDubovyk please clarify is there some SDL requirement for that use case :

the fact that the policy server is not sending the application policies in the first place is an error, and the PTU is invalid if it doesn't include the app IDs that were sent in the snapshot.
?

@LuxoftAKutsan, @jacobkeeler Currently we have no requirements for SDL to validate presence of previously sent with snapshot appID in PTU.  
Please, note : per our knowledge, the existing Ford policy backend already has this described functionality.  
If the same logic is required on SDL, should we process it as a new requirement?

jacobkeeler:
@KhrystynaDubovyk @AKalinich-Luxoft I think that an app with default policies should be updated in the PTU if those default policies are also updated. If there is a case where they shouldn't be, that seems like an error case to me. 

I would prefer that they be just updated rather than the PTU validation being strict on the matter, but the current case where that app is just left with a cached version of the default policies seems like a bug to me.

In any case, this PR does seem to improve this functionality, so I will likely still merge it for the release, but I would not consider the fix complete.

Issue #1782, opened 2017-09-26
Do not allow streaming state change after the app lost approval


t-yoshii:
**Description:**

In current implementation, OnVideoDataStreaming notification may be sent by app which already lost streaming approval. This can cause unexpected streaming stop.

Scenario:
When video app1 is switched to video app2, app1's OnVideoDataStreaming(false)/StopStream and app2's StartStream/OnVideoDataStreaming(true) are sent to HMI almost same timing.
If app1's OnVideoDataStreaming(false)/StopStream are sent after app2's StartStream/OnVideoDataStreaming(true), the video on HMI is stopped due to OnVideoDataStreaming(false).

I will create PR not to send OnVideoDataStreaming after app lost approval.

E-SAITO-TMC:
@joeygrover -san
A fix has been proposed for this issue. When will it be merged? For the next HU development of TOYOTA, it is necessary to improve the quality of SDL Core, so we believe it is necessary to modify this issue.

theresalech:
Hi @E-SAITO-TMC, contents for SDL releases are determined by the Steering Committee, and the Project Maintainer works on features and bug fixes raised as high priority by Steering Committee representatives. This particular issue has not yet been raised as high priority. However, if Toyota would like to see this fix in the next release of SDL Core (planned for October 28, 2020), we can make note of such.

Please confirm if this is priority for Toyota. If so, we will include this issue in our proposed release contents, which will be presented to the Steering Committee and voted upon/finalized in June. Thank you!

E-SAITO-TMC:
@theresalech -san Thank you for your comment. This issue is medium-priority for Toyota.

theresalech:
@E-SAITO-TMC, thank you. We have documented that this is medium priority for Toyota, and will include for Steering Committee review and vote when planning the contents for the Q4 Core release. This planning will take place in June.

iCollin:
Hello, this issue was fixed by https://github.com/smartdevicelink/sdl_core/pull/3154 and is tested by the first test in https://github.com/smartdevicelink/sdl_atf_test_scripts/pull/2298, which is passing against current develop. Closing as it is no longer reproducible.

PR #1783, opened 2017-09-26
Fix/do not allow streaming state change after lost approval

t-yoshii:
This PR is to fix issue #1782 

Unit tests are added.

### DIFF
--- a/src/components/application_manager/src/application_impl.cc
+++ b/src/components/application_manager/src/application_impl.cc
@@ -523,9 +523,12 @@ void ApplicationImpl::SuspendStreaming(
     video_streaming_suspended_ = true;
 
     if (video_streaming_approved()) {
-      MessageHelper::SendOnDataStreaming(service_type, false, application_manager_);
+      MessageHelper::SendOnDataStreaming(
+          service_type, false, application_manager_);
     } else {
-      LOG4CXX_INFO(logger_, "Do not suspend video streaming because already not approved");
+      LOG4CXX_INFO(
+          logger_,
+          "Do not suspend video streaming because already not approved");
     }
   } else if (ServiceType::kAudio == service_type) {
     audio_stream_suspend_timer_.Stop();
@@ -534,9 +537,12 @@ void ApplicationImpl::SuspendStreaming(
     audio_streaming_suspended_ = true;
 
     if (audio_streaming_approved()) {
-      MessageHelper::SendOnDataStreaming(service_type, false, application_manager_);
-    }  else {
-      LOG4CXX_INFO(logger_, "Do not suspend audio streaming because already not approved");
+      MessageHelper::SendOnDataStreaming(
+          service_type, false, application_manager_);
+    } else {
+      LOG4CXX_INFO(
+          logger_,
+          "Do not suspend audio streaming because already not approved");
     }
   }
 }
--- a/src/components/application_manager/src/application_impl.cc
+++ b/src/components/application_manager/src/application_impl.cc
@@ -517,33 +517,27 @@ void ApplicationImpl::SuspendStreaming(
   LOG4CXX_AUTO_TRACE(logger_);
 
   if (ServiceType::kMobileNav == service_type) {
+    if (!video_streaming_approved()) {
+      LOG4CXX_INFO(logger_, "Video streaming not approved");
+      return;
+    }
     video_stream_suspend_timer_.Stop();
     application_manager_.OnAppStreaming(app_id(), service_type, false);
     sync_primitives::AutoLock lock(video_streaming_suspended_lock_);
     video_streaming_suspended_ = true;
-
-    if (video_streaming_approved()) {
-      MessageHelper::SendOnDataStreaming(
-          service_type, false, application_manager_);
-    } else {
-      LOG4CXX_INFO(
-          logger_,
-          "Do not suspend video streaming because already not approved");
-    }
+    MessageHelper::SendOnDataStreaming(
+        service_type, false, application_manager_);
   } else if (ServiceType::kAudio == service_type) {
+    if (!audio_streaming_approved()) {
+      LOG4CXX_INFO(logger_, "Audio streaming not approved");
+      return;
+    }
     audio_stream_suspend_timer_.Stop();
     application_manager_.OnAppStreaming(app_id(), service_type, false);
     sync_primitives::AutoLock lock(audio_streaming_suspended_lock_);
     audio_streaming_suspended_ = true;
-
-    if (audio_streaming_approved()) {
-      MessageHelper::SendOnDataStreaming(
-          service_type, false, application_manager_);
-    } else {
-      LOG4CXX_INFO(
-          logger_,
-          "Do not suspend audio streaming because already not approved");
-    }
+    MessageHelper::SendOnDataStreaming(
+        service_type, false, application_manager_);
   }
 }
 
--- a/src/components/application_manager/src/application_impl.cc
+++ b/src/components/application_manager/src/application_impl.cc
@@ -521,13 +521,24 @@ void ApplicationImpl::SuspendStreaming(
     application_manager_.OnAppStreaming(app_id(), service_type, false);
     sync_primitives::AutoLock lock(video_streaming_suspended_lock_);
     video_streaming_suspended_ = true;
+
+    if (video_streaming_approved()) {
+      MessageHelper::SendOnDataStreaming(service_type, false, application_manager_);
+    } else {
+      LOG4CXX_INFO(logger_, "Do not suspend video streaming because already not approved");
+    }
   } else if (ServiceType::kAudio == service_type) {
     audio_stream_suspend_timer_.Stop();
     application_manager_.OnAppStreaming(app_id(), service_type, false);
     sync_primitives::AutoLock lock(audio_streaming_suspended_lock_);
     audio_streaming_suspended_ = true;
+
+    if (audio_streaming_approved()) {
+      MessageHelper::SendOnDataStreaming(service_type, false, application_manager_);
+    }  else {
+      LOG4CXX_INFO(logger_, "Do not suspend audio streaming because already not approved");
+    }
   }
-  MessageHelper::SendOnDataStreaming(service_type, false, application_manager_);
 }
 
 void ApplicationImpl::WakeUpStreaming(
--- a/src/components/application_manager/src/application_impl.cc
+++ b/src/components/application_manager/src/application_impl.cc
@@ -536,6 +536,10 @@ void ApplicationImpl::WakeUpStreaming(
   LOG4CXX_AUTO_TRACE(logger_);
 
   if (ServiceType::kMobileNav == service_type) {
+    if (!video_streaming_approved()) {
+      LOG4CXX_INFO(logger_, "Video streaming not approved yet");
+      return;
+    }
     sync_primitives::AutoLock lock(video_streaming_suspended_lock_);
     if (video_streaming_suspended_) {
       application_manager_.OnAppStreaming(app_id(), service_type, true);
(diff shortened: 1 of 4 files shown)


jacobkeeler:
@t-yoshii Please sign the [CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform) if you haven't already.

shoamano83:
Hi @jacobkeeler and @JackLivio , please consider increasing this priority as I am hitting the issue while implementing https://github.com/smartdevicelink/sdl_core/issues/2065. Thank you!

jacobkeeler:
@t-yoshii This crash appears to be a result of the `audio_stream_suspend_timer_` firing after the `ApplicationImplTest.Suspend_WakeUpAudioStreaming_StreamingApproved` is completed. The timer is set [here](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/application_manager/src/application_impl.cc#L548). I would consider this to be out of the scope of this PR, and this is likely the reason that this test was disabled in the first place.

src/components/application_manager/src/application_impl.cc
```diff
@@ -511,13 +511,24 @@ void ApplicationImpl::SuspendStreaming(
     application_manager_.OnAppStreaming(app_id(), service_type, false);
     sync_primitives::AutoLock lock(video_streaming_suspended_lock_);
     video_streaming_suspended_ = true;
+
+    if (video_streaming_approved()) {
```
jacobkeeler: I don't see any guarantee that another app is streaming when this app is disallowed, so it seems to me like this would just result in `OnVideoDataStreaming(false)` not being sent even when no app is streaming. This would result in the wrong state for the HMI. A better check would be to check if another app is streaming to determine whether to send this message or not.

src/components/application_manager/src/application_impl.cc
```diff
@@ -526,6 +537,10 @@ void ApplicationImpl::WakeUpStreaming(
   LOG4CXX_AUTO_TRACE(logger_);
 
   if (ServiceType::kMobileNav == service_type) {
+    if (!video_streaming_approved()) {
```
jacobkeeler: One potential solution, add an if statement which shuts down an existing stream if it is present when this method is called:

streamer_adapter.cc
```cpp
int32_t StreamerAdapter::get_current_app_id() {
  return current_application_;
}
```

media_manager_impl.cc
```cpp
int32_t MediaManagerImpl::GetCurrentStreamingApp(protocol_handler::ServiceType service_type) {
  return streamer_[service_type]->get_current_app_id();
}
```

application_manager_impl.cc
```cpp
ApplicationSharedPtr ApplicationManagerImpl::GetCurrentStreamingApp(protocol_handler::ServiceType service_type) {
  return application(media_manager_->GetCurrentStreamingApp(service_type));
}
```

ApplicationImpl::WakeUpStreaming
```cpp
ApplicationSharedPtr current_streaming_app = application_manager_.GetCurrentStreamingApp(service_type);
if (current_streaming_app->app_id() != 0 && current_streaming_app->app_id() != app_id()) {
  current_streaming_app->StopStreamingForce(service_type);
}
```

src/components/application_manager/src/application_impl.cc
t-yoshii: Thank you for your review.

OnVideoDataStreaming(false) will be sent before this app loses approval.

Application loses its approval at ApplicationImpl::StopNaviStreaming(). This function will be called from ApplicationImpl::StopStreaming or ApplicationImpl::StopStreamingForce. Both function call ApplicationImpl::SuspendStreaming() (this sends OnVideoDataStreaming(false) to HMI)
before calling ApplicationImpl::StopNaviStreaming().

So I think OnVideoDataStreaming(false) is already sent when this if statement become false.

src/components/application_manager/test/application_impl_test.cc
```diff
@@ -740,16 +741,37 @@ TEST_F(ApplicationImplTest, SuspendNaviStreaming) {
...
+  protocol_handler::ServiceType type = protocol_handler::ServiceType::kAudio;
+  EXPECT_CALL(mock_application_manager_, OnAppStreaming(app_id, type, false));
+  EXPECT_CALL(*MockMessageHelper::message_helper_mock(),
+              SendOnDataStreaming(type, false, _)).Times(0);
+  app_impl->SuspendStreaming(type);
+}
+
 // TODO {AKozoriz} : Fix tests with streaming (APPLINK-19289)
-TEST_F(ApplicationImplTest, DISABLED_Suspend_WakeUpAudioStreaming) {
+TEST_F(ApplicationImplTest, Suspend_WakeUpAudioStreaming_StreamingApproved) {
```
jacobkeeler: Please disable these tests, they fail for an unrelated reason as mentioned before:
1. `Suspend_WakeUpAudioStreaming_StreamingApproved`
2. `Suspend_WakeUpNaviStreaming_StreamingApproved`

The issue with these is that they start a timer in the `WakeUpStreaming` function that fires after the test is done, causing a crash.

Another option might be to modify this by adding another corresponding `SuspendStreaming` so that this timer is cancelled before the end of the test.

src/components/application_manager/src/application_impl.cc
jacobkeeler: Ah, I believe that I may have been confusing this with `video_streaming_allowed()`, which is modified outside of this class.

Is there any reason that this statement is not put above the rest of the statement, as is done in the `WakeUpStreaming` method?

```cpp
  if (video_streaming_approved()) {
    LOG4CXX_INFO(
        logger_,
        "Do not suspend video streaming because already not approved");
    return;
  }
  video_stream_suspend_timer_.Stop();
  application_manager_.OnAppStreaming(app_id(), service_type, false);
  sync_primitives::AutoLock lock(video_streaming_suspended_lock_);
  video_streaming_suspended_ = true;
  MessageHelper::SendOnDataStreaming(
      service_type, false, application_manager_);
```

jacobkeeler: My understanding was that you were saying that the issue was that `SuspendStreaming` was being called twice for the same service, which would mean that `video_streaming_suspended_` was already set to true if `video_streaming_approved()` is false. 

> So I think OnVideoDataStreaming(false) is already sent when this if statement become false.

If this double-call is not the cause of this issue, I can't see what could be, as these two functions are the only places where the `SendOnDataStreaming` method is used.

t-yoshii: Consider following scenario:
There are 2 video streaming apps.
1. App1 becomes FULL and starts navi and audio services and become approved.
2. App1 will send onVideoStreaming(true)
3. User switches App2 to FULL.
4. App1 becomes BACKGROUND and stops streaming
	4-1. App1 sends onVideoStreaming(false)
	4-2. App1 becomes unapproved
5. App2 starts streaming
	5-1. App2 becomes approved
	5-2. App2 sends onVideoStreaming(true)

If there is some potential timing issue that step 5-2 is executed before step 4-1, App2's streaming will be stopped right after streaming started. In this situation, App1 should not send onVideoStreaming(false) but should clear flags. This is my implementation.

In any case, this discussion is for abnormal scenerio. If you have a strong reason for early return is preferrable, I will change PR.

jacobkeeler: These flags are guaranteed to have been cleared if `video_streaming_approved` is false. As you have pointed out, `video_streaming_approved` is never set to false until after `SuspendStreaming` is called.

Your scenario just seems to bring us back to my first comment, `video_streaming_approved` tells us nothing about whether another app is streaming or not.

Using your scenario:
With these changes, if 5-2 happens before 4-1 the HMI will see
1. onVideoStreaming(true)
2. onVideoStreaming(false)

Which is the exact scenario this change is meant to avoid. This if statement does not prevent this, because `video_streaming_approved` is still true when `SuspendStreaming` is called for app 1, since 4-2 hasn't happened yet.

t-yoshii: Thanks for checking. I understand it.

I updated PR. Thanks!

jacobkeeler:
Tests are still failing: 
```
[ RUN      ] ApplicationImplTest.SuspendNaviStreaming_StreamingApproved
[       OK ] ApplicationImplTest.SuspendNaviStreaming_StreamingApproved (0 ms)
[ RUN      ] ApplicationImplTest.SuspendAudioStreaming_StreamingApproved
[       OK ] ApplicationImplTest.SuspendAudioStreaming_StreamingApproved (1 ms)
[ RUN      ] ApplicationImplTest.SuspendNaviStreaming_StreamingNotApproved
/home/jacob/sdl_core/src/components/application_manager/test/application_impl_test.cc:757: Failure
Actual function call count doesn't match EXPECT_CALL(mock_application_manager_, OnAppStreaming(app_id, type, false))...
         Expected: to be called once
           Actual: never called - unsatisfied and active
[  FAILED  ] ApplicationImplTest.SuspendNaviStreaming_StreamingNotApproved (0 ms)
[ RUN      ] ApplicationImplTest.SuspendAudioStreaming_StreamingNotApproved
/home/jacob/sdl_core/src/components/application_manager/test/application_impl_test.cc:766: Failure
Actual function call count doesn't match EXPECT_CALL(mock_application_manager_, OnAppStreaming(app_id, type, false))...
         Expected: to be called once
           Actual: never called - unsatisfied and active
[  FAILED  ] ApplicationImplTest.SuspendAudioStreaming_StreamingNotApproved (1 ms)
[ RUN      ] ApplicationImplTest.Suspend_WakeUpAudioStreaming_StreamingNotApproved
/home/jacob/sdl_core/src/components/application_manager/test/application_impl_test.cc:807: Failure
Actual function call count doesn't match EXPECT_CALL(mock_application_manager_, OnAppStreaming(app_id, type, false))...
         Expected: to be called once
           Actual: never called - unsatisfied and active
[  FAILED  ] ApplicationImplTest.Suspend_WakeUpAudioStreaming_StreamingNotApproved (1 ms)
[ RUN      ] ApplicationImplTest.Suspend_WakeUpNaviStreaming_StreamingNotApproved
/home/jacob/sdl_core/src/components/application_manager/test/application_impl_test.cc:823: Failure
Actual function call count doesn't match EXPECT_CALL(mock_application_manager_, OnAppStreaming(app_id, type, false))...
         Expected: to be called once
           Actual: never called - unsatisfied and active
[  FAILED  ] ApplicationImplTest.Suspend_WakeUpNaviStreaming_StreamingNotApproved (0 ms)
[ RUN      ] ApplicationImplTest.StopStreaming_StreamingApproved
[       OK ] ApplicationImplTest.StopStreaming_StreamingApproved (1 ms)
[ RUN      ] ApplicationImplTest.PushPopMobileMessage
[       OK ] ApplicationImplTest.PushPopMobileMessage (1 ms)
[----------] 43 tests from ApplicationImplTest (27 ms total)
```

jacobkeeler:
@t-yoshii I still do not see how this solution would fix the described issue, as I've stated earlier:
> Using your scenario:
> With these changes, if 5-2 happens before 4-1 the HMI will see
> 
> 1. onVideoStreaming(true)
> 2. onVideoStreaming(false)
> 
> Which is the exact scenario this change is meant to avoid. 

I have been unable to reproduce the base issue right now because it appears that I don't currently have the necessary setup, but it would be helpful if you have a test case which shows the exact sequence of events to reproduce this issue (including messages to and from mobile). Could you provide such a test scenario (possibly through an ATF test)?

src/components/application_manager/src/application_impl.cc
t-yoshii: I'm sorry I made you misunderstand.

I do not have a reproducing steps about the scenario which I explained (5-2 happens before 4-1).
When I created this PR, I wondered which to choose how to implement SuspendStreaming(). When app do not have approval, app should do "early return (* 1)" or "do not send onVideoDataStreaming (false) but clear other flags (*2)".

As my first decision, I thought that the latter (*2) would be better for the case where the problem described in the previous comment "potentialy" exist.
However, since you commented that above scenario will not occur, I changed it to (*1). 

Therefore, my last commit does not fix specific issue. I just changed the implementation style.

t-yoshii:
@jacobkeeler
I'm sorry I made you misunderstand.

I do not have a reproducing steps about the scenario which I explained (5-2 happens before 4-1).
When I created this PR, I wondered which to choose how to implement SuspendStreaming(). When app do not have approval, app should do "early return (* 1)" or "do not send onVideoDataStreaming (false) but clear other flags (*2)".

As my first decision, I thought that the latter (*2) would be better for the case where the problem described in the previous comment "potentialy" exist.
However, since you commented that above scenario will not occur, I changed it to (*1).

Therefore, my last commit does not fix specific issue. I just changed the implementation style.

iCollin:
Closing as https://github.com/smartdevicelink/sdl_core/issues/1782 is no longer reproducible.

PR #1785, opened 2017-09-26
Contribution documentation update

dev-gh:
Replaces outdated rules and add supported environment details

### DIFF
--- a/COMMITTERS.md
+++ b/COMMITTERS.md
@@ -59,7 +59,9 @@ Branch name should be:
 
 ##Pull request message
 * Describe the reason why you created pull request and what changes there are.
-* Add related GitHub issue id as link.
+* Add related Jira ticket as link.
+   ( EXAMPLE:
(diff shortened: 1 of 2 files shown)


COMMITTERS.md
```diff
@@ -87,7 +85,7 @@ Branch name should be:
 * Successfully passed review.
 * Rebase in case of existing conflicts.
 * Contributor can squash commits in case of adding same code in different commits.
```
jacobkeeler: This document as a whole looks like it is specifically for use by Luxoft developers, I have removed it in #1800. If it is still needed by Luxoft, it might be better if it were moved to another repository (maybe `LuxoftSDL` or `sdl_requirements`) as it is a bit confusing to keep it here considering CONTRIBUTING.md already exists.

In any case, since this is deleted in the new release, please remove any changes to this file in this PR.

dev-gh: @jacobkeeler 
Done with 096051f

PR #1787, opened 2017-09-27
Fix SendLocation response for not applicable resultCodes from HMI

AKalinich-Luxoft:
Currently SDL transfers from HMI to Mobile any result code which came in response. Some RPCs in MOBILE_API have own fixed list of valid result codes for this RPC. According to requirements if SDL receives some of result code which is not present in this fixed list, SDL should behave the same way as in case of "any unknown issue" and respond with `GENERIC_ERROR`.

In this fix was added response message validation check after applying schema and respond with `GENERIC_ERROR` if response message is invalid due to some reason. This fix affects all RPC responses.

### DIFF
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -1746,6 +1746,19 @@ bool ApplicationManagerImpl::ManageMobileCommand(
   }
 
   if (message_type == mobile_apis::messageType::response) {
+    if (!message->isValid()) {
+      LOG4CXX_ERROR(logger_, "Mobile message response is invalid");
+      smart_objects::SmartObjectSPtr response =
+          MessageHelper::CreateNegativeResponse(
+              connection_key,
+              static_cast<int32_t>(function_id),
+              correlation_id,
+              static_cast<int32_t>(mobile_apis::Result::GENERIC_ERROR));
+
+      SendMessageToMobile(response);
+      return false;
+    }
+
     if (command->Init()) {
       command->Run();
       command->CleanUp();
--- a/src/components/application_manager/src/commands/mobile/send_location_request.cc
+++ b/src/components/application_manager/src/commands/mobile/send_location_request.cc
@@ -146,7 +146,7 @@ void SendLocationRequest::on_event(const event_engine::Event& event) {
     const bool result = PrepareResultForMobileResponse(
         result_code, HmiInterfaces::HMI_INTERFACE_Navigation);
     SendResponse(result,
-                 MessageHelper::HMIToMobileResult(result_code),
+                 GetMobileResultCode(result_code),
                  response_info.empty() ? NULL : response_info.c_str(),
                  &(message[strings::params]));
     return;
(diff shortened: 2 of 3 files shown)


AKalinich-Luxoft:
@AByzhynar, @LuxoftAKutsan, @dev-gh please review

AByzhynar:
@AKalinich-Luxoft Please use already existed code from mobile_api_schema.cc(file generated automatically. Look in build dir)

AKalinich-Luxoft:
@AByzhynar done. Please review again

src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -1746,6 +1746,19 @@ bool ApplicationManagerImpl::ManageMobileCommand(
   }
 
   if (message_type == mobile_apis::messageType::response) {
+    if (!message->isValid()) {
+      LOG4CXX_ERROR(logger_, "RET Mobile message response is invalid");
```
AByzhynar: @AKalinich-Luxoft What is `RET` ?

AKalinich-Luxoft: @AByzhynar I suppose it means RETURN from function with some exceptional case and a reason of return. This RET is placed in all exceptional cases inside ApplicationManagerImpl::ManageMobileCommand

AKalinich-Luxoft: @AByzhynar removed in https://github.com/smartdevicelink/sdl_core/pull/1787/commits/59af9ea67c0acdeb85ffe39e65eafbd321bd01a3

src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -1746,6 +1746,19 @@ bool ApplicationManagerImpl::ManageMobileCommand(
   }
 
   if (message_type == mobile_apis::messageType::response) {
+    if (!message->isValid()) {
+      LOG4CXX_ERROR(logger_, "Mobile message response is invalid");
+      smart_objects::SmartObjectSPtr response =
+          MessageHelper::CreateNegativeResponse(
```
dev-gh: @AKalinich-Luxoft 
I understand PR descriptions as following: in case there is result code out of request allowed code list - return GENERIC_ERROR instead of HMI result code, but keep all other data. However, CreateNegativeResponse doesn't forward any data back in response.
Is my understanding wrong?

AKalinich-Luxoft: @dev-gh your understanding is correct, however I don't think we should keep all other data for response in this case

dev-gh: @AKalinich-Luxoft 
"I don't think" doesn't sound like a requirement.

AKalinich-Luxoft: @dev-gh requirement, which was linked for this defect APPLINK-16281:
In case an application sends a request and:
-> either unknown issue happened
-> either something went wrong,
SDL must:
respond with "GENERIC_ERROR, success:"false" to mobile application.

Issue #1740, opened 2017-08-29
SendLocation Updates


OlesiaV:
## Description:

_As a mobile application I want to be able to set the location on the head unit for plotting route when connected to the vehicle without being asked to reconfirm the location._

## Related Diagrams:

_Diagrams (for example sequence diagram) describing these changes_

## Links:

- **Requirement:** [#24](https://github.com/smartdevicelink/sdl_requirements/issues/24) SendLocation to set a location on the head unit
- **Parent task:** 

## Expected Delivery (Definition of Done)

- [ ] Requirements analysis and updates
- [ ] Source code updates
- [ ] Code comments
- [ ] UTs add/update
- [ ] Integration tests add/update
- [ ] **Manual tests (not required)**
- [ ] Add/update CI plans/jobs
- [ ] SDD updates
- [ ] SAD updates
- [ ] Guidelines update 1 ([sdl_core_guides](https://github.com/smartdevicelink/sdl_core_guides))
- [ ] Guidelines update 2 ([sdl_hmi_integration_guidelines](https://github.com/smartdevicelink/sdl_hmi_integration_guidelines))

PR #1799, opened 2017-10-03
Fix policy issues for disallowed parameters

AKalinich-Luxoft:
## EXTERNAL_PROPRIETARY
Added new values to `PermitResult` enum for more detailed result description of `CheckPermissions()` call. This values will be used in AM layer for generating info message string.

### PROPRIETARY
Added new values to `PermitResult` enum for more detailed result description of `CheckPermissions()` call. This values will be used in AM layer for generating info message string.
Also there was added missed logic for checking RPC params and returning correct `PermitResult` code and params lists to `CheckPermissions()` function.

### BOTH
Added generating of info message param depending on check permissions result:
- for disallowed rpc
- for all disallowed params in allowed rpc
- for partially disallowed params in allowed rpc

Fixed requests param reading from incoming message. There was a problem that `asBool()` function of smart object returns false for string/array/object types so this params will no be included to params check list and will not be cut off after that. This check was replaced to ignore only null/invalid smart object types.

Fixed affected unit tests logic.

`"GitHub issue link" ~ "https://github.com/SmartDeviceLink/sdl_core/issues/1740"`

### DIFF
--- a/src/components/application_manager/src/commands/command_request_impl.cc
+++ b/src/components/application_manager/src/commands/command_request_impl.cc
@@ -579,7 +579,10 @@ bool CommandRequestImpl::CheckAllowedParameters() {
   smart_objects::SmartMap::const_iterator iter_end = s_map.map_end();
 
   for (; iter != iter_end; ++iter) {
-    if (iter->second.asBool()) {
+    if (helpers::Compare<smart_objects::SmartType, helpers::NEQ, helpers::ALL>(
+            iter->second.getType(),
+            smart_objects::SmartType_Null,
+            smart_objects::SmartType_Invalid)) {
       LOG4CXX_DEBUG(logger_, "Request's param: " << iter->first);
       params.insert(iter->first);
     }
--- a/src/components/application_manager/src/commands/command_request_impl.cc
+++ b/src/components/application_manager/src/commands/command_request_impl.cc
@@ -104,6 +104,25 @@ const std::string CreateInfoForUnsupportedResult(
   }
 }
 
+std::string CreateInfoForPolicyPermitResult(const PermitResult value) {
+  switch (value) {
+    case PermitResult::kRpcDisallowed: {
+      return "RPC is disallowed by Policies";
+    }
+    case PermitResult::kRpcUserDisallowed: {
+      return "RPC is disallowed by the User";
+    }
+    case PermitResult::kRpcAllParamsDisallowed: {
+      return "Requested parameters are disallowed by Policies";
+    }
+    case PermitResult::kRpcAllParamsUserDisallowed: {
+      return "Requested parameters are disallowed by User";
+    }
+    default:
+      return std::string();
+  }
+}
+
 bool CheckResultCode(const ResponseInfo& first, const ResponseInfo& second) {
   if (first.is_ok && second.is_unsupported_resource &&
       second.interface_state == HmiInterfaces::STATE_NOT_AVAILABLE) {
@@ -582,7 +601,11 @@ bool CommandRequestImpl::CheckAllowedParameters() {
             check_result,
             correlation_id(),
             app->app_id());
-
+    std::string info =
+        CreateInfoForPolicyPermitResult(parameters_permissions_.permit_result);
+    if (!info.empty()) {
+      (*response)[strings::msg_params][strings::info] = info;
+    }
     application_manager_.SendMessageToMobile(response);
     return false;
   }
@@ -721,7 +744,11 @@ void CommandRequestImpl::AddDisallowedParametersToInfo(
   }
 
   if (!info.empty()) {
-    info += " disallowed by policies.";
+    const uint32_t params_count =
+        removed_parameters_permissions_.disallowed_params.size() +
+        removed_parameters_permissions_.undefined_params.size();
+    info += params_count > 1 ? " parameters are " : " parameter is ";
+    info += "disallowed by Policies";
 
     if (!response[strings::msg_params][strings::info].asString().empty()) {
       // If we already have info add info about disallowed params to it
--- a/src/components/application_manager/src/commands/mobile/send_location_request.cc
+++ b/src/components/application_manager/src/commands/mobile/send_location_request.cc
@@ -60,29 +60,25 @@ void SendLocationRequest::Run() {
     return;
   }
 
-  smart_objects::SmartObject& msg_params = (*message_)[strings::msg_params];
-  if (msg_params.keyExists(strings::delivery_mode)) {
-    const policy::RPCParams& allowed_params =
-        parameters_permissions().allowed_params;
-
-    if (helpers::in_range(allowed_params, strings::delivery_mode)) {
-      msg_params.erase(strings::delivery_mode);
-    }
-  }
-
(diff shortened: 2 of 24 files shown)


src/components/application_manager/src/commands/mobile/send_location_request.cc
```diff
@@ -60,29 +60,25 @@ void SendLocationRequest::Run() {
     return;
   }
 
-  smart_objects::SmartObject& msg_params = (*message_)[strings::msg_params];
-  if (msg_params.keyExists(strings::delivery_mode)) {
```
dev-gh: @AKalinich-Luxoft 
Removing this part you break the requirement https://adc.luxoft.com/jira/browse/APPLINK-21910

src/components/policy/policy_external/include/policy/policy_types.h
```diff
@@ -113,7 +113,13 @@ typedef std::vector<std::string> PermissionsList;
  */
 typedef std::vector<std::string> StringArray;
 
-enum PermitResult { kRpcAllowed = 0, kRpcDisallowed, kRpcUserDisallowed };
+enum PermitResult {
+  kRpcAllowed = 0,
```
dev-gh: @AKalinich-Luxoft 
I think better to rename new enum values 
kRpcParamsDisallowed --> kRpcAllParamsDisallowed
kRpcParamsUserDisallowed --> kRpcAllParamsUserDisallowed
because if only some parameters in request are disallowed, RPC can be processed and won't be disallowed

src/components/policy/policy_regular/include/policy/policy_types.h
```diff
@@ -113,7 +113,13 @@ typedef std::vector<std::string> PermissionsList;
  */
 typedef std::vector<std::string> StringArray;
 
-enum PermitResult { kRpcAllowed = 0, kRpcDisallowed, kRpcUserDisallowed };
+enum PermitResult {
+  kRpcAllowed = 0,
```
dev-gh: @AKalinich-Luxoft 
See comment for 'external' implementation

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -708,21 +731,27 @@ void CommandRequestImpl::AddDissalowedParameterToInfoString(
...
   for (; it != removed_parameters_permissions_.undefined_params.end(); ++it) {
     AddDissalowedParameterToInfoString(info, (*it));
+    ++params_added;
   }
 
   if (!info.empty()) {
-    info += " disallowed by policies.";
-
+    if (params_added > 1) {
+      info += " parameters are disallowed by Policies";
```
dev-gh: @AKalinich-Luxoft 
Won't be better something like
```
std::string article = removed_parameters_permissions_.undefined_params > 0 ||
                              removed_parameters_permissions_.disallowed_params > 0 ?
                              " are " : " is ";
info += "parameter(s) " + article + " disallowed by Policies"
```
Logic based on counter looks more error prone.

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -638,6 +642,25 @@ bool CommandRequestImpl::CheckHMICapabilities(
   return false;
 }
 
+std::string CommandRequestImpl::GetPermitResultInfoMessage() const {
```
dev-gh: @AKalinich-Luxoft 
I think it's better to convert it to free function and just pass result as parameter.

src/components/application_manager/src/commands/mobile/send_location_request.cc
AKalinich-Luxoft: @dev-gh not yet. If this param will be disallowed by policy, then it will be cut off by CommandRequestImpl::RemoveDisallowedParameters() call before exact command execution, so this check seems redundant on this step

src/components/policy/policy_external/include/policy/policy_types.h
AKalinich-Luxoft: @dev-gh good idea. Fixed in https://github.com/smartdevicelink/sdl_core/pull/1799/commits/80883f435a69895f8b0c1757f6458707c8fcfd12

src/components/policy/policy_regular/include/policy/policy_types.h
AKalinich-Luxoft: @dev-gh fixed in https://github.com/smartdevicelink/sdl_core/pull/1799/commits/92d67279a19c3e4f95011d018c98eb2d66b21629

src/components/application_manager/src/commands/command_request_impl.cc
AKalinich-Luxoft: @dev-gh done in https://github.com/smartdevicelink/sdl_core/pull/1799/commits/a693891743df516a77f3a65c451f24b5da9aafac

src/components/application_manager/src/commands/command_request_impl.cc
AKalinich-Luxoft: @dev-gh agree. Logic was updated in https://github.com/smartdevicelink/sdl_core/pull/1799/commits/a693891743df516a77f3a65c451f24b5da9aafac

src/components/application_manager/test/commands/command_request_impl_test.cc
```diff
@@ -393,8 +393,13 @@ TEST_F(CommandRequestImplTest,
   EXPECT_CALL(app_mngr_, CheckPolicyPermissions(_, _, _, _))
       .WillOnce(Return(mobile_apis::Result::INVALID_ENUM));
 
+  smart_objects::SmartObjectSPtr response =
+      new smart_objects::SmartObject(smart_objects::SmartType_Map);
```
AByzhynar: @AKalinich-Luxoft use `MakeShared` please

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -560,7 +579,8 @@ bool CommandRequestImpl::CheckAllowedParameters() {
   smart_objects::SmartMap::const_iterator iter_end = s_map.map_end();
 
   for (; iter != iter_end; ++iter) {
-    if (iter->second.asBool()) {
+    if (smart_objects::SmartType_Null != iter->second.getType() &&
+        smart_objects::SmartType_Invalid != iter->second.getType()) {
```
AByzhynar: @AKalinich-Luxoft May be we can use here `helpers::Compare` from utils?

AKalinich-Luxoft: @AByzhynar fixed in ac2f5e9fa37dab60ed9abd75421f70e3ae72d6fc

src/components/application_manager/test/commands/command_request_impl_test.cc
AKalinich-Luxoft: @AByzhynar fixed in 6cbc168f93754d569ef91cef7320931abdae4f19

AKalinich-Luxoft:
@AByzhynar there was added policy params checking for notifications. Please review commit 38e395694dc6ee90b820b024337e000501f83637

Issue #1802, opened 2017-10-04
media_manager_test fails with EXTENDED_MEDIA_MODE=ON


shoamano83:
### Step to reproduce:
1. Check out latest release/4.4.0 branch (commit 41d3029c1698e5b12a529a44fd2c0bae8f66202d)
2. Cherry-pick commit 1b7b9eef9897c3126b94163fd820ad691cb59825
3. Modify CMakeLists.txt to enable EXTENDED_MEDIA_MODE and BUILD_TESTS
4. Give a build ($ cmake ../sdl_core && make)
5. Run unit tests of media_manager_test ($ cd src/components/media_manager/test/ && ./media_manager_test)

### Actual Result:
Some unit tests fail with following error message:
```
[ RUN      ] MediaManagerImplTest.StartMicrophoneRecording_SourceFileIsWritable_ExpectTrue
/home/samano/projects/sdl_core/sdl_core/src/components/media_manager/test/media_manager_impl_test.cc:128: Failure
Actual function call count doesn't match EXPECT_CALL(mock_media_manager_settings_, recording_file_source())...
         Expected: to be called once
           Actual: never called - unsatisfied and active
/home/samano/projects/sdl_core/sdl_core/src/components/media_manager/test/media_manager_impl_test.cc:126: Failure
Actual function call count doesn't match EXPECT_CALL(mock_media_manager_settings_, app_resource_folder())...
         Expected: to be called once
           Actual: never called - unsatisfied and active
[  FAILED  ] MediaManagerImplTest.StartMicrophoneRecording_SourceFileIsWritable_ExpectTrue (14 ms)
[ RUN      ] MediaManagerImplTest.StartMicrophoneRecording_OutputFileNotExists_ExpectFalse
/home/samano/projects/sdl_core/sdl_core/src/components/media_manager/test/media_manager_impl_test.cc:128: Failure
Actual function call count doesn't match EXPECT_CALL(mock_media_manager_settings_, recording_file_source())...
         Expected: to be called once
           Actual: never called - unsatisfied and active
/home/samano/projects/sdl_core/sdl_core/src/components/media_manager/test/media_manager_impl_test.cc:126: Failure
Actual function call count doesn't match EXPECT_CALL(mock_media_manager_settings_, app_resource_folder())...
         Expected: to be called once
           Actual: never called - unsatisfied and active
[...]

PR #1803, opened 2017-10-04
fix: media_managet_test fails if EXTENDED_MEDIA_MODE is enabled

shoamano83:
Fixes https://github.com/smartdevicelink/sdl_core/issues/1802.


shoamano83:
@JackLivio @jacobkeeler Please kindly review.

PR #1828, opened 2017-10-25
Fix incorrect App ID caused by implicit conversion of smart pointer to integer

leekillough:
Fix silent bug caused by implicit conversion of smart pointer to integer. Garbage App ID based on pointer value. Lots of symptoms caused by this bug, which could be prevented by not letting smart pointers implicitly convert to boolean or integer.


jacobkeeler:
@leekillough Please sign the [CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform), this will allow us to accept your contribution.

PR #1841, opened 2017-11-03
 Fix disallowed info message generating for specific RPC

AKalinich-Luxoft:
There was found a problem that general info message in case of disallowed by policies params is not applicable for all RPCs. For example `SubscribeVehicleData/UnsubscribeVehicleData` has own specific format of info message in this case.
To fix that following changes were done:
- Added virtual function `AddSpecificInfoToResponse()` which could be overriden in child RPCs in case custom info message required
- Removed hardcoded checks of `function_id` from `command_request_impl` - all this checks will be done using polymorphism inside aproriate RPCs
- Fixed affected UT expectation. Invalid unit test was replaced with actual one

### DIFF
--- a/src/components/application_manager/include/application_manager/commands/command_helper.h
+++ b/src/components/application_manager/include/application_manager/commands/command_helper.h
@@ -1,0 +1,85 @@
+/*
+ Copyright (c) 2017, Ford Motor Company
+ All rights reserved.
+
+ Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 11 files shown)


AKalinich-Luxoft:
@AByzhynar, @LuxoftAKutsan, @dev-gh, @LitvinenkoIra please review

AByzhynar:
@AKalinich-Luxoft Fix conflicts please and check functionality and unit tests locally and on CI server.

PR #1842, opened 2017-11-03
Fix disallowed info message generating for specific RPC

AKalinich-Luxoft:
### NOTE
This is a duplicate of PR #1841 which send to another branch.

There was found a problem that general info message in case of disallowed by policies params is not applicable for all RPCs. For example `SubscribeVehicleData/UnsubscribeVehicleData` has own specific format of info message in this case.
To fix that following changes were done:
- Added virtual function `AddSpecificInfoToResponse()` which could be overriden in child RPCs in case custom info message required
- Removed hardcoded checks of `function_id` from `command_request_impl` - all this checks will be done using polymorphism inside aproriate RPCs
- Fixed affected UT expectation. Invalid unit test was replaced with actual one

### DIFF
--- a/src/components/application_manager/include/application_manager/commands/command_helper.h
+++ b/src/components/application_manager/include/application_manager/commands/command_helper.h
@@ -1,0 +1,85 @@
+/*
+ Copyright (c) 2017, Ford Motor Company
+ All rights reserved.
+
+ Redistribution and use in source and binary forms, with or without
+ modification, are permitted provided that the following conditions are met:
+
+ Redistributions of source code must retain the above copyright notice, this
+ list of conditions and the following disclaimer.
+
+ Redistributions in binary form must reproduce the above copyright notice,
+ this list of conditions and the following
+ disclaimer in the documentation and/or other materials provided with the
+ distribution.
+
+ Neither the name of the Ford Motor Company nor the names of its contributors
+ may be used to endorse or promote products derived from this software
+ without specific prior written permission.
+
+ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include "application_manager/commands/command_impl.h"
+#include "smart_objects/smart_object.h"
+
+namespace application_manager {
+namespace commands {
+
+namespace command_helper {
+/**
+ * @brief Adds disallowed parameters back to response with appropriate
+ * reasons
+ * @param permissions Structure with info about disallowed parameters
+ * @param response Response message, which should be extended with blocked
+ * parameters reasons
+ */
+void AddDisallowedParameters(const CommandParametersPermissions& permissions,
+                             smart_objects::SmartObject& response);
+
+/**
+ * @brief Adds disallowed parameters to response info
+ * @param permissions Structure with info about disallowed parameters
+ * @param response Response message, which info should be extended
+ */
+void AddDisallowedParametersToInfo(
+    const CommandParametersPermissions& permissions,
+    smart_objects::SmartObject& response);
+
+/**
+ * @brief Remove from current message parameters disallowed by policy table
+ * @param permissions Structure with info about parameters which should be
+ * disallowed
+ * @param removed_permissions Structure with info about parameters which
+ * actually was disallowd in current message
+ * @param msg_params Message params where disallowed params should be removed
+ */
+void RemoveDisallowedParameters(
+    const CommandParametersPermissions& permissions,
+    CommandParametersPermissions& removed_permissions,
+    smart_objects::SmartObject& msg_params);
+
+/**
+ * @brief Checks if any request param was marked as disallowed by policy
+ * @param permissions Structure with info about parameters which actually was
+ * disallowd in current message
+ * @return true if any param was marked as disallowed
+ */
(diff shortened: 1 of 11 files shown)


AKalinich-Luxoft:
@AByzhynar, @LuxoftAKutsan, @dev-gh, @LitvinenkoIra please review

src/components/application_manager/src/commands/mobile/subscribe_vehicle_data_request.cc
```diff
@@ -293,6 +293,20 @@ void SubscribeVehicleDataRequest::AddAlreadySubscribedVI(
   }
 }
 
+void SubscribeVehicleDataRequest::AddSpecificInfoToResponse(
+    smart_objects::SmartObject& response) {
+  LOG4CXX_AUTO_TRACE(logger_);
+  AddDisallowedParameters(response);
+  if (helpers::Compare<mobile_apis::Result::eType, helpers::EQ, helpers::ONE>(
```
dev-gh: @AKalinich-Luxoft 
As far as I see the only logic here is to decide whether info string needs to be cleared. So you can 
- move AddSpecificInfoToResponse, AddDisallowedParametersToInfo to private section of CommandRequestImpl
- create e.g. 'virtual bool DoCleanInfo(result_code)' in CommandRequestImpl class in private section and override it in commands subclasses
- call DoCleanInfo from AddSpecificInfoToResponse in CommandRequestImpl

Also I think it should be possible also to move to private Add/RemoveDisallowedParameters interfaces.

AByzhynar:
@AKalinich-Luxoft Fix conflicts please and check functionality and unit tests locally and on CI server.

src/components/application_manager/src/commands/mobile/subscribe_vehicle_data_request.cc
AKalinich-Luxoft: @dev-gh clearing info is not only logic in overriden functions. `AddDisallowedParameters()` and `AddDisallowedParametersToInfo()` also should be called in these overrides. These functions are used only here. So creating separate `DoCleanInfo()` virtual function is not a good idea. This is a reason why I added `AddSpecificInfoToResponse()` function.
As alternative I can move functions like Add/RemoveDisallowedParameters, AddDisallowedParametersToInfo for example to `MessageHelper` because this is not default behaviour for all classes inherited from `CommandRequestImpl`.

dev-gh: @AKalinich-Luxoft 
Moving code is even better, so move as much as possible out of CommandRequestImpl. I think its better to create another "command helper" space for such utils used by commands since MessageHelper is overwhelmed already with other stuff.

AKalinich-Luxoft: @dev-gh good idea. I added new helper for commands and moved part of functionality related to disallowed params to this helper. I believe that it is possible to move more functions from command base class, however it will be a lot of changes for PR with fix, so it will have a sense to do in a separate pull request.
Please review final result in 4652e3534f26ee37da3b56a5ebe8345c22266278

src/components/application_manager/include/application_manager/commands/command_helper.h
```diff
@@ -0,0 +1,92 @@
...
+#include <string>
+
+#include "application_manager/commands/command_impl.h"
+#include "smart_objects/smart_object.h"
+#include "utils/macro.h"
+
+namespace application_manager {
+namespace commands {
+
+namespace CommandHelper {
```
dev-gh: @AKalinich-Luxoft 
This looks like a class, so I would suggest to change is to 'command_helper'

src/components/application_manager/include/application_manager/commands/command_helper.h
```diff
@@ -0,0 +1,92 @@
...
+ INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef SRC_COMPONENTS_APPLICATION_MANAGER_INCLUDE_APPLICATION_MANAGER_COMMANDS_COMMAND_HELPER_H
+#define SRC_COMPONENTS_APPLICATION_MANAGER_INCLUDE_APPLICATION_MANAGER_COMMANDS_COMMAND_HELPER_H
+
+#include <algorithm>
```
dev-gh: @AKalinich-Luxoft 
I think this include is not needed

src/components/application_manager/include/application_manager/commands/command_helper.h
```diff
@@ -0,0 +1,92 @@
...
+ LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef SRC_COMPONENTS_APPLICATION_MANAGER_INCLUDE_APPLICATION_MANAGER_COMMANDS_COMMAND_HELPER_H
```
dev-gh: @AKalinich-Luxoft 
Suggest to use '#pragma once'

AKalinich-Luxoft: @dev-gh fixed in https://github.com/smartdevicelink/sdl_core/pull/1842/commits/a042c516f25c1ea93a6b1408e60989c24201efbc

src/components/application_manager/include/application_manager/commands/command_helper.h
AKalinich-Luxoft: @dev-gh agree. Moved to implementation where it is really used. See https://github.com/smartdevicelink/sdl_core/pull/1842/commits/a042c516f25c1ea93a6b1408e60989c24201efbc

src/components/application_manager/include/application_manager/commands/command_helper.h
AKalinich-Luxoft: @dev-gh yes, thank you. Initially that was a class, but later I decided to use just a namespace. Renamed in https://github.com/smartdevicelink/sdl_core/pull/1842/commits/a042c516f25c1ea93a6b1408e60989c24201efbc

src/components/application_manager/src/commands/command_helper.cc
```diff
@@ -0,0 +1,199 @@
...
+                           mobile_apis::VehicleDataResultCode::eType code)
+      : response_(response), code_(code) {}
+
+  bool operator()(const std::string& param) {
+    using namespace application_manager;
+    const VehicleData& vehicle_data = MessageHelper::vehicle_data();
+    VehicleData::const_iterator it = vehicle_data.find(param);
+    if (vehicle_data.end() != it) {
+      smart_objects::SmartObjectSPtr disallowed_param =
+          new smart_objects::SmartObject(smart_objects::SmartType_Map);
```
dev-gh: @AKalinich-Luxoft 
MakeShared would be better

AKalinich-Luxoft: @dev-gh fixed in https://github.com/smartdevicelink/sdl_core/pull/1842/commits/67e0f3076ebdf4fdf9b4d3802e8ca29732f2af11

PR #1847, opened 2017-11-08
Fix core crash in case when HMI sends invalid response

LitvinenkoIra:
Fix core crash in case when HMI sends invalid response with the "result" (or "error") of an integer type 
(or any other type, but **not an object**)

e.g. `HMI->SDL : { jsonrpc = "2.0", result = 0 } `

### DIFF
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -3115,6 +3115,22 @@ void ApplicationManagerImpl::Handle(const impl::MessageFromHmi message) {
     return;
   }
 
+  Json::Value value;
+  Json::Reader reader;
+  reader.parse(message->json_message(), value);
+  if (value.isMember(("result"))) {
+    if (!value["result"].isObject()) {
+      LOG4CXX_ERROR(logger_, "Invalid type of 'result' field.");
+      return;
+    }
+  }
+  if (value.isMember(("error"))) {
+    if (!value["error"].isObject()) {
+      LOG4CXX_ERROR(logger_, "Invalid type of 'error' field.");
+      return;
+    }
+  }
+
 #ifdef SDL_REMOTE_CONTROL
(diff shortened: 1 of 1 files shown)


src/components/functional_module/src/plugin_manager.cc
```diff
@@ -193,6 +193,12 @@ bool PluginManager::IsHMIMessageForPlugin(application_manager::MessagePtr msg) {
   Json::Value value;
   Json::Reader reader;
   reader.parse(msg->json_message(), value);
+
+  if (value.isMember(("result"))) {
```
AKalinich-Luxoft: @LitvinenkoIra according to [our specification](https://github.com/smartdevicelink/sdl_core/blob/master/src/components/interfaces/Json_HMI_message_specification.txt) "result" in HMI response is always object so I propose to move this check somewhere to upper level, for example in `ApplicationManagerImpl::Handle`

LitvinenkoIra: @AKalinich-Luxoft Done aaaf361da4c8fe81873c35a9b61ab8f247ae89ec

src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -3115,6 +3115,16 @@ void ApplicationManagerImpl::Handle(const impl::MessageFromHmi message) {
     return;
   }
 
+  Json::Value value;
+  Json::Reader reader;
+  reader.parse(message->json_message(), value);
+  if (value.isMember(("result"))) {
```
AByzhynar: @LitvinenkoIra May be it is worth to add the same check for erroneous response from hmi? error also must be an object according to specification

LitvinenkoIra: @AByzhynar Done d3d9c53035ce23e08e856bc71878bce9595404a5

PR #1852, opened 2017-11-09
Fix/alert maneuver wrong response

AByzhynar:
- Fixed generation of response to mobile app. 
Removed sending of empty info string to mobile app as afterwards SDL checks all responses according
to MobileApi smart schema and info message (if present) can not be empty. 
Therefore in previous implementation SDL (due to adding empty info string to response) converted successful responses from HMI to GENERIC_ERROR.

- Added function to MessageHelper which allows to convert SmartObject to std::string

### DIFF
--- a/src/components/application_manager/include/application_manager/message_helper.h
+++ b/src/components/application_manager/include/application_manager/message_helper.h
@@ -621,6 +621,9 @@ class MessageHelper {
 
   static bool PrintSmartObject(const smart_objects::SmartObject& object);
 
+  static std::string SmartObjectToString(
+      const smart_objects::SmartObject& object);
+
   template <typename From, typename To>
   static To ConvertEnumAPINoCheck(const From& input) {
     return static_cast<To>(input);
--- a/src/components/application_manager/include/application_manager/message_helper.h
+++ b/src/components/application_manager/include/application_manager/message_helper.h
@@ -621,6 +621,11 @@ class MessageHelper {
 
   static bool PrintSmartObject(const smart_objects::SmartObject& object);
 
+  /**
+   * @brief Converts common SmartObject to its string representation
+   * @param object SmartObject to convert
+   * @return string representation of given SmartObject
+   */
   static std::string SmartObjectToString(
       const smart_objects::SmartObject& object);
 
--- a/src/components/application_manager/src/commands/mobile/alert_maneuver_request.cc
+++ b/src/components/application_manager/src/commands/mobile/alert_maneuver_request.cc
@@ -173,14 +173,10 @@ void AlertManeuverRequest::on_event(const event_engine::Event& event) {
   std::string return_info;
   mobile_apis::Result::eType result_code;
   const bool result = PrepareResponseParameters(result_code, return_info);
-  bool must_be_empty_info = false;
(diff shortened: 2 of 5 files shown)


src/components/application_manager/include/application_manager/message_helper.h
```diff
@@ -621,6 +621,9 @@ class MessageHelper {
 
   static bool PrintSmartObject(const smart_objects::SmartObject& object);
 
+  static std::string SmartObjectToString(
```
dev-gh: @AByzhynar 
Where do you use it?

AByzhynar: @dev-gh I used it everywhere during debug. Every time during fixing of different bugs. Not only me. This function is needed by other devs also. After defect fixed I removed usage of this function as it is needed just for debug.

AByzhynar: @dev-gh It is needed everytime as it allows to print out messages from hmi and mobile. It helps to determine where is the problem.

dev-gh: @AByzhynar 
Perhaps 'other devs' are everyone except me. How do you think I should get that sacred  knowledge for w/o any description? So please create separate 'debug_helper' space in another file and put all tools like that there with full description. MessageHelper is overwhelmed already and needs to be cleaned up eventually.

AByzhynar: @dev-gh Declined as there are cases when we need to convert message to string and print it to log. Description will be added.

AByzhynar: @dev-gh Description added in c9b0c5c4e9bc6953544a5095a851c8e87ac0b867

dev-gh: @AByzhynar 
I didn't understand what is the problem to put it to another file and call it when its needed. Can you explain particular case?

AByzhynar: @dev-gh I don't want  to create another MessageHelper. I consider that all things related to messages should be located in one place. I have put it to MessageHelper as other similar converters are located here.

dev-gh: @AByzhynar 
By your explanation above it is debug tool and its not being used at the moment. I don't see any point to collect debug tools among number of other methods. I believe nobody but couple guys will ever know that it is there. So if you want other to use it you have to make clear its purpose. Same thing I did for application in iAP-over-Bt code and same thing did Andrey Kalinich for commands.

AByzhynar: @dev-gh I don't know what you did for application in iAP-over-Bt code and what did Andrey Kalinich for commands.

dev-gh: @AByzhynar 
Take a look

AByzhynar: @dev-gh You have made a lot of changes. Where I have to look exactly? Point to the commit please.

AByzhynar:
Rebuild required

AByzhynar:
Rebuild required

PR #1938, opened 2017-12-04
Add DEPRECATED macro, and fix resulting deprecation warnings

### LINKED ISSUE #1758: [SDL 0092] Mark public deprecated methods
Proposal: [Mark public deprecated methods](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0092-Deprecated-interfaces-markup.md)

>Public methods is part of SDL interface. New features may use some existing interfaces and expect some behavior. Sometimes some features use interfaces that are planned to be removed.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/278

jacobkeeler:
Fixes #1758 

### Risk

### Summary
- Added DEPRECATED macro
- Marked all deprecated methods with this new macro
- Fixed any warnings resulting from the use of this new macro

##### Enhancements
* Added DEPRECATED macro. A warning is generated whenever a method marked with this macro is used. 
    * NOTE: because of other warnings in the project without moving to the next c++ standard, the `-Wno-deprecated-declarations` flag is still used in building. This needs to be removed to see the new warnings generated by the DEPRECATED macro.

### CLA

### DIFF
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -1166,8 +1166,7 @@ void ApplicationManagerImpl::ReplaceHMIByMobileAppId(
   }
 }
 
-// DEPRECATED
-bool ApplicationManagerImpl::StartNaviService(
+DEPRECATED bool ApplicationManagerImpl::StartNaviService(
     uint32_t app_id, protocol_handler::ServiceType service_type) {
   using namespace protocol_handler;
   LOG4CXX_AUTO_TRACE(logger_);
@@ -1338,8 +1337,9 @@ void ApplicationManagerImpl::StopNaviService(
   app->StopStreaming(service_type);
 }
 
-// DEPRECATED
-bool ApplicationManagerImpl::OnServiceStartedCallback(
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+DEPRECATED bool ApplicationManagerImpl::OnServiceStartedCallback(
     const connection_handler::DeviceHandle& device_handle,
     const int32_t& session_key,
     const protocol_handler::ServiceType& type) {
@@ -1374,6 +1374,7 @@ bool ApplicationManagerImpl::OnServiceStartedCallback(
   }
   return false;
 }
+#pragma GCC diagnostic pop
 
 void ApplicationManagerImpl::OnServiceStartedCallback(
     const connection_handler::DeviceHandle& device_handle,
@@ -2283,7 +2284,7 @@ bool ApplicationManagerImpl::ConvertSOtoMessage(
   }
 
   if (message.getElement(jhs::S_PARAMS).keyExists(strings::binary_data)) {
-    application_manager::BinaryData binaryData(
+    const application_manager::BinaryData binaryData(
         message.getElement(jhs::S_PARAMS)
             .getElement(strings::binary_data)
             .asBinary());
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -1166,7 +1166,7 @@ void ApplicationManagerImpl::ReplaceHMIByMobileAppId(
   }
 }
 
-DEPRECATED bool ApplicationManagerImpl::StartNaviService(
+bool ApplicationManagerImpl::StartNaviService(
     uint32_t app_id, protocol_handler::ServiceType service_type) {
   using namespace protocol_handler;
   LOG4CXX_AUTO_TRACE(logger_);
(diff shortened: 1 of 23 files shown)


src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -1338,8 +1337,9 @@ void ApplicationManagerImpl::StopNaviService(
   app->StopStreaming(service_type);
 }
 
-// DEPRECATED
-bool ApplicationManagerImpl::OnServiceStartedCallback(
+#pragma GCC diagnostic push
```
LuxoftAKutsan: @jacobkeeler why this pragma required here ? Is this method is especial?

src/components/application_manager/src/message.cc
```diff
@@ -89,7 +89,7 @@ Message& Message::operator=(const Message& message) {
   set_data_size(message.data_size_);
   set_payload_size(message.payload_size_);
   if (message.binary_data_) {
-    set_binary_data(message.binary_data_);
+    set_binary_data((const BinaryData*)message.binary_data_);
```
LuxoftAKutsan: @jacobkeeler I would prefer to avoid the C style casts.

src/components/application_manager/src/message.cc
```diff
@@ -194,8 +194,7 @@ void Message::set_message_type(MessageType type) {
   type_ = type;
 }
 
-// DEPRECATED
-void Message::set_binary_data(BinaryData* data) {
+DEPRECATED void Message::set_binary_data(BinaryData* data) {
```
LuxoftAKutsan: @jacobkeeler isn't deprecation in header enough?

src/components/include/protocol_handler/session_observer.h
```diff
@@ -103,8 +103,19 @@ class SessionObserver {
...
+   * \brief Callback function used by ProtocolHandler
+   * when Mobile Application initiates session ending.
+   * \param connection_handle Connection identifier within which session exists
+   * \param sessionId Identifier of the session to be ended
+   * \param hashCode Hash used only in second version of SmartDeviceLink
+   * protocol.
+   * If not equal to hash assigned to session on start then operation fails.
+   * \param service_type Type of service
+   * \return uint32_t 0 if operation fails, session key otherwise
+   * \deprecated
```
LuxoftAKutsan: @jacobkeeler do we have some unified style of Doxygen documentation? 
I mean marking keywords with backslash `\brief` or with *at* symbol `@brief`?
For example, [here](https://github.com/smartdevicelink/sdl_core/blob/a8c3e2ac28d758b30a2b933918635ef33f2d4cd3/src/components/application_manager/include/application_manager/application.h#L161-L163) we use *at* symbol.
And I like marking with *at* symbol

src/components/protocol_handler/src/protocol_handler_impl.cc
```diff
@@ -1307,8 +1307,9 @@ class StartSessionHandler : public security_manager::SecurityManagerListener {
 }  // namespace
 #endif  // ENABLE_SECURITY
 
-// DEPRECATED
-RESULT_CODE ProtocolHandlerImpl::HandleControlMessageStartSession(
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
LuxoftAKutsan: @jacobkeeler please describe cases when you use ignoring -Wdeprecated-declarations diagnostic

src/components/protocol_handler/src/protocol_handler_impl.cc
```diff
@@ -1895,11 +1897,11 @@ RESULT_CODE ProtocolHandlerImpl::EncryptFrame(ProtocolFramePtr packet) {
         connection_key,
         security_manager::SecurityManager::ERROR_ENCRYPTION_FAILED,
         error_text);
+
+    uint32_t hash_id = packet->message_id();
```
LuxoftAKutsan: @jacobkeeler can be const?

src/components/application_manager/src/application_manager_impl.cc
jacobkeeler: This is a deprecated method which uses another deprecated method, the pragma prevent a warning being generated because of this.

src/components/protocol_handler/src/protocol_handler_impl.cc
jacobkeeler: This value is changed in the case of an error in `OnSessionEndedCallback`, it needs to be mutable

src/components/application_manager/src/application_manager_impl.cc
LuxoftAKutsan: @jacobkeeler  such knowledge should be shared somewhere.

src/components/include/protocol_handler/session_observer.h
jacobkeeler: I agree, and I suppose I could change it for this specific deprecated function. However, the rest of this file uses the `\` symbol instead of `@`. I was simply attempting to not break the consistency of this format within the file. It may be worth creating a PR changing all of the files to use this `@` symbol, but it is a minor enough detail that I don't think it should be a priority.

PR #1952, opened 2017-12-13
remove deadlock danger when removing Connections from a ConnectionMap

### LINKED ISSUE #1951: Avoid deadlock in Connection destructor
### Bug Report
When the TransportAdapterImpl removes a connection from the ConnectionMap, it will invoke the Connection destructor inside the "connections_lock_". This is undesirable, since it might cause a deadlock.

fronneburg:
Fixes #1951 

### Risk

### Summary
Created a new "RemoveConnection" method which safely removes a Connection object from the ConnectionMap.  Before calling "erase" on the connection object, the object's ref-count is incremented by assigning it to a local "ConnectionSPtr" variable. We then call "erase", NOT triggering the destructor, then release the "connections_lock_", and then let the local variable go out of scope, thus delaying the destructor until after the lock was released.

### CLA
- [X ] I have signed [the CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform)

### DIFF
--- a/src/components/transport_manager/src/transport_adapter/transport_adapter_impl.cc
+++ b/src/components/transport_manager/src/transport_adapter/transport_adapter_impl.cc
@@ -216,9 +216,7 @@ TransportAdapter::Error TransportAdapterImpl::Connect(
   const TransportAdapter::Error err =
       server_connection_factory_->CreateConnection(device_id, app_handle);
   if (TransportAdapter::OK != err) {
-    connections_lock_.AcquireForWriting();
-    connections_.erase(std::make_pair(device_id, app_handle));
-    connections_lock_.Release();
+    RemoveConnection(device_id, app_handle);
   }
   LOG4CXX_TRACE(logger_, "exit with error: " << err);
   return err;
@@ -562,14 +560,12 @@ void TransportAdapterImpl::DeviceDisconnected(
     listener->OnDisconnectDeviceDone(this, device_uid);
   }
 
-  connections_lock_.AcquireForWriting();
   for (ApplicationList::const_iterator i = app_list.begin();
        i != app_list.end();
        ++i) {
     ApplicationHandle app_handle = *i;
-    connections_.erase(std::make_pair(device_uid, app_handle));
+    RemoveConnection(device_uid, app_handle);
   }
-  connections_lock_.Release();
 
   RemoveDevice(device_uid);
   LOG4CXX_TRACE(logger_, "exit");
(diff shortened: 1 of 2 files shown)


src/components/transport_manager/src/transport_adapter/transport_adapter_impl.cc
```diff
@@ -756,25 +748,53 @@ void TransportAdapterImpl::RemoveFinalizedConnection(
     const DeviceUID& device_handle, const ApplicationHandle& app_handle) {
   const DeviceUID device_uid = device_handle;
   LOG4CXX_AUTO_TRACE(logger_);
-  sync_primitives::AutoWriteLock lock(connections_lock_);
+  connections_lock_.AcquireForWriting();
```
LuxoftAKutsan: @fronneburg why did you remove AutoLock in that case?

src/components/transport_manager/src/transport_adapter/transport_adapter_impl.cc
```diff
@@ -756,25 +748,53 @@ void TransportAdapterImpl::RemoveFinalizedConnection(
...
+  ConnectionSPtr connection = info.connection;
   connections_.erase(it_conn);
+  connections_lock_.Release();
+  LOG4CXX_TRACE(logger_, "RemoveFinalizedConnection Connections Lock Released");
 }
 
+void TransportAdapterImpl::RemoveConnection(
+    const DeviceUID& device_id, const ApplicationHandle& app_handle) {
+  ConnectionSPtr connection;
+  connections_lock_.AcquireForWriting();
```
LuxoftAKutsan: @fronneburg I would prefer using RAII here

src/components/transport_manager/src/transport_adapter/transport_adapter_impl.cc
fronneburg: @LuxoftAKutsan we need to make sure that the lock is released first, and THEN let "connection" go out of scope.

src/components/transport_manager/src/transport_adapter/transport_adapter_impl.cc
fronneburg: Same response here.

JackLivio:
@fronneburg Why is it important that the calling the connection destructor is delayed until after the lock is released?

fronneburg:
@JackLivio, we found a deadlock when the connection destructor (actually the "ThreadedConnection" Xevo implementation) tried to "join" a thread that was waiting on the "connections_lock_" in TransportAdapterImpl::ConnectionFinished.

iCollin:
Closed in favor of https://github.com/smartdevicelink/sdl_core/pull/3371 which is based on latest develop and changes the log type from TRACE to DEBUG.

Issue #1854, opened 2017-11-14
SDL Core dump after connect/disconnect USB cable


GetmanetsIrina:
## Occurrence: 
Rare 

## Preconditions:
1. SDL is built with PROPRIETARY flag.
2. SDL, HMI are started.

## Steps to reproduce:
1. Connect mobile via USB
2. Register app with HMI type "REMOTE_CONTROL"
3. Disconnect mobile via USB

## Expected result:
No SDL core dumps

## Actual result:
SDL Core Dump
```WARN [11 Jul 2017 11:55:47,424][0x7f70d3fff700][ApplicationManager] /home/astasiuk/GitHub/OPENSDL/sdl_core/src/components/application_manager/src/application_manager_impl.cc:1456 ApplicationManagerImpl::SendMessageToMobile: Function "OnHMIStatus" (#32768) not allowed by policyr```
```
gdb smartDeviceLinkCore core
GNU gdb (Ubuntu 7.7.1-0ubuntu5~14.04.2) 7.7.1
Copyright (C) 2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law. Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from smartDeviceLinkCore...done.

warning: core file may not match specified executable file.
[New LWP 484]
[New LWP 445]
[New LWP 448]
[New LWP 456]
[New LWP 447]
[New LWP 462]
[New LWP 449]
[New LWP 450]
[New LWP 478]
[New LWP 444]
[New LWP 452]
[New LWP 453]
[New LWP 461]
[New LWP 823]
[New LWP 839]
[New LWP 1356]
[New LWP 479]
[New LWP 483]
[New LWP 443]
[New LWP 451]
[New LWP 454]
[New LWP 455]
[New LWP 457]
[New LWP 458]
[New LWP 459]
[New LWP 460]
[New LWP 477]
[New LWP 480]
[New LWP 481]
[New LWP 482]
[...]

PR #1960, opened 2017-12-18
Fixes random SDL crash on USB disconnect

dev-gh:
Fixes #[1854](https://github.com/SmartDeviceLink/sdl_core/issues/1854)

### Risk

### Summary
Whenever USB device is unplugged there are several concurrent threads
that can remove USB connection instance: call from connection instance to
DisconnectDone() interface while aborting connection and call to
RemoveFinalizedConnections() while processing unexpected disconnect
event. Previously USB connection instance was marking itself as
"finalized" while calling AbortConnection() and than yield execution
because of cancelling the data transfer so the same connection was
being deleted from another thread during RemoveFinalizedConnection()
call which was causing consequent calls to instance members invalid.
To fix the issue order is changed so cancelling of transfer is being
done first and only after that connection is being marked as "finalized"
so it can be safely deleted either from RemoveFinalizedConnection() or
by calling DisconnectDone() in controller (TransportAdapterImpl).

### CLA

### DIFF
--- a/src/components/transport_manager/src/usb/libusb/usb_connection.cc
+++ b/src/components/transport_manager/src/usb/libusb/usb_connection.cc
@@ -81,6 +81,7 @@ UsbConnection::~UsbConnection() {
   LOG4CXX_TRACE(logger_, "enter with this" << this);
   Finalise();
   libusb_free_transfer(in_transfer_);
+  libusb_free_transfer(out_transfer_);
   delete[] in_buffer_;
   LOG4CXX_TRACE(logger_, "exit");
 }
@@ -292,9 +293,10 @@ void UsbConnection::Finalise() {
 
 void UsbConnection::AbortConnection() {
   LOG4CXX_TRACE(logger_, "enter");
+  Finalise();
   controller_->ConnectionAborted(
       device_uid_, app_handle_, CommunicationError());
-  Disconnect();
+  controller_->DisconnectDone(device_uid_, app_handle_);
   LOG4CXX_TRACE(logger_, "exit");
 }
 


src/components/transport_manager/src/usb/libusb/usb_connection.cc
```diff
@@ -81,6 +81,7 @@ UsbConnection::~UsbConnection() {
   LOG4CXX_TRACE(logger_, "enter with this" << this);
   Finalise();
   libusb_free_transfer(in_transfer_);
+  libusb_free_transfer(out_transfer_);
```
AKalinich-Luxoft: @dev-gh do we need to free out transfer here? Seems like it will be freed in `UsbConnection::OnOutTransfer()` callaback.

dev-gh: @AKalinich-Luxoft 
```
if (!current_out_message_.valid()) {
    libusb_free_transfer(transfer);
    out_transfer_ = NULL;
    waiting_out_transfer_cancel_ = false;
  }
```
If you mean this code than I think its being done only in specific case - when message is invalid.

AKalinich-Luxoft: @dev-gh in this specific case out_transfer will be a NULL. Is that OK to call `libusb_free_transfer` with NULL pointer in destructor?

dev-gh: @AKalinich-Luxoft 
Yep
```
/** \ingroup asyncio
 * Free a transfer structure. This should be called for all transfers
 * allocated with libusb_alloc_transfer().
 *
 * If the \ref libusb_transfer_flags::LIBUSB_TRANSFER_FREE_BUFFER
 * "LIBUSB_TRANSFER_FREE_BUFFER" flag is set and the transfer buffer is
 * non-NULL, this function will also free the transfer buffer using the
 * standard system memory allocator (e.g. free()).
 *

 * It is legal to call this function with a NULL transfer. In this case,
 * the function will simply return safely.
 *

 * It is not legal to free an active transfer (one which has been submitted
 * and has not yet completed).
 *
 * \param transfer the transfer to free
 */
void API_EXPORTED libusb_free_transfer(struct libusb_transfer *transfer)
```

src/components/transport_manager/src/usb/libusb/usb_connection.cc
```diff
@@ -292,9 +293,10 @@ void UsbConnection::Finalise() {
 
 void UsbConnection::AbortConnection() {
   LOG4CXX_TRACE(logger_, "enter");
+  Finalise();
```
yang1070: looks good to me

iCollin:
Closing as the [related issue](https://github.com/smartdevicelink/sdl_core/issues/1854) has been closed. Please tag me if this PR is still needed.

PR #1962, opened 2017-12-18
protect pt_ from concurrent access

fronneburg:
Fixes #[1961]

### Risk

### Summary
Protect access to the "pt_" member variable of the cache_manager class with a recursive lock

### CLA
- [X ] I have signed [the CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform)

### DIFF
--- a/src/components/policy/policy_external/src/cache_manager.cc
+++ b/src/components/policy/policy_external/src/cache_manager.cc
@@ -250,6 +250,7 @@ policy_table::MessageString FindLanguage(
 
 CacheManager::CacheManager()
     : CacheManagerInterface()
+    , cache_lock_(true)
     , pt_(new policy_table::Table)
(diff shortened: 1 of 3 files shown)


LuxoftAKutsan:
@fronneburg  do we have issue with policy table race conditions?
cache_manager is not thread safe because PolicyManager should guarantee it's thread safety. Additional mutexes would add performance issues and they are redundant I think

fronneburg:
In our testing we hit an issue with this in the LTS branch. We found a SIGSEGV crash where the tombstone clearly showed one thread in CacheManager::ApplyUpdate, while the other thread was in CacheManager::CheckPermissions. This PR is simply attempting to move our fix for that into the "develop" branch here.

PR #2030, opened 2018-02-27
Feature/add ut dtls encryption

ValeriiMalkov:
#This PR is ready for review.

Risk

This PR makes no API changes.

Summary

This PR provides unit test coverage for DTLS protocol
    - Correct initialization crypto manager with different protocols
    - Handshake without protocol verification
    - Handshake with CA verification only server side
    - Handshake with CA verification only client side
    - Handshake with CA verification both sides
    - Handshake client side fail with certificate not signed error
    - Handshake client side fail with certificate is expired error
    - Handshake fail with application name and application ID are not valid
    - Reset connection without verification
    - Reset connection with verification both sides
    - Test with correct protocol
    - Test with uncorrect protocol

### CLA


AByzhynar:
@ValeriiMalkov Check why build compilation is failed, please!

PR #2058, opened 2018-03-12
Fix disabling RC-functionality

LitvinenkoIra:
in case when RC functionality is disabled, all RC-requests from applications with REMOTE_CONTROL appHMIType should not be processed
https://github.com/smartdevicelink/sdl_requirements/blob/master/detailed_docs/RC/rc_enabling_disabling.md

### DIFF
--- a/src/components/application_manager/rpc_plugins/rc_rpc_plugin/include/rc_rpc_plugin/resource_allocation_manager.h
+++ b/src/components/application_manager/rpc_plugins/rc_rpc_plugin/include/rc_rpc_plugin/resource_allocation_manager.h
@@ -120,6 +120,10 @@ class ResourceAllocationManager {
    */
   virtual void SendOnRCStatusNotification() = 0;
 
+  virtual bool is_rc_enabled() const = 0;
+
+  virtual void set_rc_enabled(const bool value) = 0;
+
   virtual ~ResourceAllocationManager() {}
 };
 
--- a/src/components/application_manager/rpc_plugins/rc_rpc_plugin/src/resource_allocation_manager_impl.cc
+++ b/src/components/application_manager/rpc_plugins/rc_rpc_plugin/src/resource_allocation_manager_impl.cc
@@ -23,7 +23,8 @@ ResourceAllocationManagerImpl::ResourceAllocationManagerImpl(
     application_manager::rpc_service::RPCService& rpc_service)
     : current_access_mode_(hmi_apis::Common_RCAccessMode::AUTO_ALLOW)
     , app_mngr_(app_mngr)
-    , rpc_service_(rpc_service) {}
+    , rpc_service_(rpc_service)
+    , is_rc_enabled_(true) {}
 
 ResourceAllocationManagerImpl::~ResourceAllocationManagerImpl() {}
 
@@ -282,6 +283,14 @@ void ResourceAllocationManagerImpl::SendOnRCStatusNotification() {
     auto msg_to_hmi = CreateOnRCStatusNotificationToHmi(rc_app);
     rpc_service_.SendMessageToHMI(msg_to_hmi);
   }
+}
+
+bool ResourceAllocationManagerImpl::is_rc_enabled() const {
+  return is_rc_enabled_;
+}
+
+void ResourceAllocationManagerImpl::set_rc_enabled(const bool value) {
+  is_rc_enabled_ = value;
 }
 
 void ResourceAllocationManagerImpl::SetResourceFree(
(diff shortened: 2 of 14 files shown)


src/components/application_manager/rpc_plugins/rc_rpc_plugin/include/rc_rpc_plugin/resource_allocation_manager.h
```diff
@@ -120,6 +120,10 @@ class ResourceAllocationManager {
    */
   virtual void SendOnRCStatusNotification() = 0;
 
+  virtual bool is_rc_enabled() const = 0;
+
+  virtual void set_is_rc_enabled(const bool value) = 0;
```
AByzhynar: @LitvinenkoIra REbase to @BSolonenko branch or cherry-pick necessary commits. There are functions with proper naming

src/components/application_manager/rpc_plugins/rc_rpc_plugin/src/resource_allocation_manager_impl.cc
```diff
@@ -23,7 +23,8 @@ ResourceAllocationManagerImpl::ResourceAllocationManagerImpl(
     application_manager::rpc_service::RPCService& rpc_service)
     : current_access_mode_(hmi_apis::Common_RCAccessMode::AUTO_ALLOW)
     , app_mngr_(app_mngr)
-    , rpc_service_(rpc_service) {}
+    , rpc_service_(rpc_service)
+    , is_rc_enabled_(true) {}
```
AByzhynar: @LitvinenkoIra Why it is `true` by default? May be add value to ini file and read from there? To make it default-configurable without changing our SDL code.

src/components/application_manager/rpc_plugins/rc_rpc_plugin/include/rc_rpc_plugin/resource_allocation_manager.h
LitvinenkoIra: @AByzhynar you mean this functions? https://github.com/smartdevicelink/sdl_core/commit/900055a4b78449f61915980999ad210ba2c4b48e
is_rc_enabled - checks that RC functionality allowed (by user via HMI)
is_remote_control_supported - checks that app registered with AppHMIType::REMOTE_CONTROL

src/components/application_manager/rpc_plugins/rc_rpc_plugin/src/resource_allocation_manager_impl.cc
LitvinenkoIra: @AByzhynar as i understood from https://github.com/smartdevicelink/sdl_requirements/blob/master/detailed_docs/RC/rc_enabling_disabling.md
by default, RC functionality is enabled

AByzhynar: @LitvinenkoIra I agree. But to make it disabled by default - in case if needed : we have to change SDL code, recompile and so on. In case of .ini file value : just change 1 value. Nothing more.

src/components/application_manager/rpc_plugins/rc_rpc_plugin/include/rc_rpc_plugin/resource_allocation_manager.h
AByzhynar: @LitvinenkoIra Right. I meant those functions. But as these functions are different - rename properly

LuxoftAKutsan: @abyzhynar @ilitvinenko there is no need to add additional profile value for it. RC is enabled by default by requirements. In case if we won't to disable RC we always are able to send RPC. I don't see use case of modifying this default value

src/components/application_manager/rpc_plugins/rc_rpc_plugin/src/resource_allocation_manager_impl.cc
LuxoftAKutsan: @AByzhynar @ilitvinenko there is no need to add additional profile value for it. RC is enabled by default by requirements. In case if we won't to disable RC we always are able to send RPC. I don't see use case of modifying this default value

AByzhynar: @LuxoftAKutsan No problem.

src/components/application_manager/rpc_plugins/rc_rpc_plugin/include/rc_rpc_plugin/resource_allocation_manager.h
LitvinenkoIra: @AByzhynar renamed function 94a277ba57f8412d23e023946a2568290ecc86c2

AByzhynar:
@LitvinenkoIra Check coding style, please

Issue #2165, opened 2018-05-01
iOS app may receive another HMI level when SDL.ActivateApp is issued right after BC.OnAppRegistered notification


shoamano83:
### Bug Report
If HMI issues SDL.ActivateApp right after it receives BC.OnAppRegistered notification, the HMI Level of the app may change to another one, for example, the default HMI level defined in policy table.

This is a timing issue. It doesn't always reproduce, and it may not reproduce if the head unit has powerful CPU.

##### Reproduction Steps
1. Terminate iOS app on the phone (by swiping it out)
2. Launch the app from head unit by using iAP AppLaunch command (note: this is achieved by head unit's proprietary implementation)
3. HMI issues SDL.ActivateApp as soon as the app is registered.

##### Expected Behavior
The app launches on the phone, gets registered to Core, and should get HMI level FULL with SDL.ActivateApp.

##### Observed Behavior
The app is launched, registered and got HMI level FULL. However, it will receive another HMI level (for example, the level defined in the policy table which is NONE in our case.)

##### OS & Version Information
* OS/Version: iPhone with iOS 11.1.
* SDL Core Version: release 4.3
* Testing Against: proprietary iOS app and proprietary HMI

##### Test Case, Sample Code, and / or Example App
I think this is same as https://github.com/smartdevicelink/sdl_core/issues/805.

shoamano83:
We didn't see the issue with 4.1 Core after https://github.com/smartdevicelink/sdl_core/issues/805 is fixed, however we started seeing the issue again when we upgraded Core to 4.3.

Possible reason is that 4.3 includes additional implementation before resumption runs, which may cause the timing issue to happen more easily. For example,
- commit https://github.com/smartdevicelink/sdl_core/pull/1106/commits/088bc6dbd4a59b4bcea423accef081cb7beae889
- commit https://github.com/smartdevicelink/sdl_core/pull/1137/commits/0a6fa22c1c4337396db122fd1ff978b38673b4ba

shoamano83:
The root cause is that `ResumeCtrlImpl::StartResumption()` / `ResumeCtrlImpl::StartResumptionOnlyHMILevel()` are triggered *after* Core notifies HMI of OnAppRegistered.

This will cause two issues:

(1) If HMI issues SDL.ActivateApp really quickly, the default HMI level may be set after SDL.ActivateApp is processed.

[...]

PR #2166, opened 2018-05-01
fix: make sure to send OnAppRegistered notification after resumption

shoamano83:
Fixes https://github.com/smartdevicelink/sdl_core/issues/2165

### Risk

### Testing Plan
N/A

### Summary
This PR changes the order of resumption and sending BC.OnAppRegistered notification, so that BC.OnAppRegistered is always notified after resumption runs.

##### Breaking Changes
* None

##### Enhancements
* None

##### Bug Fixes
* If HMI issues SDL.ActivateApp right after BC.OnAppRegistered notification, the request may be processed before resumption runs. This caused HMI level to change again after ActivateApp is processed.

### Tasks Remaining:
- None.

### CLA


shoamano83:
We have applied this fix internally. It seems that we forgot to contribute to upstream, sorry about that.

Unit test is not available as this is a timing issue and cannot be reproduced reliably. What we did for verification is to add some `sleep()` to block the thread of `ResumeCtrlImpl::StartResumption()`, then issue SDL.ActivateApp from HMI as soon as possible.

src/components/application_manager/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -736,6 +734,9 @@ void RegisterAppInterfaceRequest::SendRegisterAppInterfaceResponseToMobile(
     resumer.StartResumptionOnlyHMILevel(application);
   }
 
+  SendOnAppRegisteredNotificationToHMI(
+      *(application.get()), resumption, need_restore_vr);
```
jacobkeeler: This results in the resumption commands (AddCommand, etc.) being sent before `OnAppRegistered` is sent to the HMI. This means that the HMI would need to add commands for an app that it doesn't know exists yet, which causes errors with the sample HMIs.

shoamano83:
Hi @jacobkeeler , thank you for pointing out the resumption issue.

Let me clarify:
- RPCs for resumption (e.g. AddCommand) have to be issued after HMI level for the app is configured
- HMI requires OnAppRegistered notification before receiving RPCs for resumption

If I understand the implementation correctly, then current sequence is something like:

1. Send RegisterAppInterface response to mobile
2. Notify HMI of BC.OnAppRegistered
3. Restore default HMI level from policy table
4. Send AddCommand, etc. to HMI for resumption
5. Start resumption timer
6. Once the timer is fired and the app hasn't been activated yet, then the app's state (including HMI level) is restored

This is still an idea but probably we can modify the sequence like below? I would like to seek for a solution which will minimize the impact and required change.

1. Send RegisterAppInterface response to mobile
2. Restore default HMI level from policy table
3. Start resumption timer
4. Notify HMI of BC.OnAppRegistered
5. Send AddCommand, etc. to HMI for resumption
6. Once the timer is fired and the app hasn't been activated yet, then the app's state (including HMI level) is restored

Thanks.

shoamano83:
Hi @jacobkeeler ,

I investigated the code some more and came up with another idea. The new approach is to delay SDL.ActivateApp getting processed until resumption is triggered.
I think the risk of this change is minimum since we don't change RAI sequence at all and it only affects SDL.ActivateApp. The only downside is that we add two public methods in `Application` class just for this feature, and the code size is increased.
I tested the change internally and confirmed that it fixes the original issue https://github.com/smartdevicelink/sdl_core/issues/2165 we observed.

Can I ask for another review please? I rebased the commits onto latest develop and I hope the changes are clear.

Thanks.

src/components/application_manager/include/application_manager/application.h
```diff
@@ -1224,6 +1224,24 @@ class Application : public virtual InitialApplicationData,
    */
   virtual const smart_objects::SmartObject& get_user_location() const = 0;
 
+  /*
```
jacobkeeler, line 1227:
```suggestion
  /**
```

src/components/application_manager/include/application_manager/application_impl.h
```diff
@@ -505,6 +505,24 @@ class ApplicationImpl : public virtual Application,
       const smart_objects::SmartObject& user_location) OVERRIDE;
 
   const smart_objects::SmartObject& get_user_location() const OVERRIDE;
+  
+  /*
```
jacobkeeler, line 509:
```suggestion
  /**
```

src/components/application_manager/test/application_impl_test.cc
```diff
@@ -879,6 +881,50 @@ TEST_F(ApplicationImplTest, SetDeferredResumptionHMILevelTest) {
...
+    virtual void exitThreadMain() {}
+
+   private:
+    std::shared_ptr<ApplicationImpl> app_;
+    std::shared_ptr<TestAsyncWaiter> waiter_;
+    unsigned int seconds_;
+  };
+
+  std::shared_ptr<TestAsyncWaiter> waiter =
+      std::make_shared<TestAsyncWaiter>();
```
jacobkeeler, line 911:
```suggestion
  std::shared_ptr<TestAsyncWaiter> waiter = std::make_shared<TestAsyncWaiter>();
```

src/components/application_manager/src/application_impl.cc
```diff
@@ -1123,4 +1124,30 @@ void ApplicationImpl::SwapMobileMessageQueue(
...
+
+  sync_primitives::AutoLock auto_lock(setup_in_progress_lock_);
+  bool previous_flag = setup_in_progress_;
+  setup_in_progress_ = in_progress;
+  LOG4CXX_DEBUG(logger_,
+                "setup_in_progress_ of app " << app_id_ << " updated to "
+                                             << setup_in_progress_);
+
+  // unblock WaitForSetupDone() when the flag changes from true to false
+  if (previous_flag == true && setup_in_progress_ == false) {
```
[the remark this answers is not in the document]
shoamano83: Thanks for your suggestion. I updated the code in commit ba0620857f0e15103acac523b9319704d8c1174c.

src/components/application_manager/src/application_impl.cc
```diff
@@ -1123,4 +1124,30 @@ void ApplicationImpl::SwapMobileMessageQueue(
...
+    setup_in_progress_cond_.Broadcast();
```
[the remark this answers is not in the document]
shoamano83: Hmm, I think we have no issue calling `Broadcast` with the lock held. In pthread library, it is valid and has no issue. (Actually some people prefer calling `pthread_cond_broadcast()` with the lock held.)
Reference: https://stackoverflow.com/questions/1105250/call-pthread-cond-broadcast-with-mutex-held-or-not

Regarding `AssertFreeAndMarkTaken`:

the function `wait()` acquires the lock before it returns. So if we release the lock after calling Broadcast, `wait()` just waits until the lock is released, and returns. Hence "assert free" is always true.

I do not have strong opinion on this, so if you still think we should unlock first then just let me know. I'll update the code.

iCollin:
Closed via https://github.com/smartdevicelink/sdl_core/pull/3559

Please feel free to reopen if the issue is reproduced.

PR #2204, opened 2018-05-22
Fixed: app does not activate when policies are disabled.

### LINKED ISSUE #842: App does not activate when policies are disabled
Using the web hmi, after activating an app, the alert shows "Activating App..." and nothing follows.

Same effect if built using extended policy or not.

BSolonenko:
Fixes #842

This PR is ready for review.

### Risk

### Summary
Removed POLICY_LIB_CHECK_VOID() from PolicyHandler::OnActivateApp 
as redundant, because the case with disabled policy library is already handled 
inside that function. Also, it works incorrectly for the case when the policy is 
disabled - this macro breaks the function at the beginning so applications could 
not be activated when policy library is disabled.

### CLA


BSolonenko:
Aalso, these changes were proposed in https://github.com/smartdevicelink/sdl_core/pull/951

BSolonenko:
Rebuild required

Issue #2143, opened 2018-04-17
Possible cyclic reference of shared pointers of `Application` class


shoamano83:
### Bug Report

There is a possible cyclic reference of shared pointers of `Application` class. Because of it, `ApplicationImpl` instance may not be released properly.

Looking at latest commit 10e17225d0a961f18493acd5ff8c1721e441cb14 of `develop` branch:

- An `Application` instance has `ApplicationState` in its member.
https://github.com/smartdevicelink/sdl_core/blob/10e17225d0a961f18493acd5ff8c1721e441cb14/src/components/application_manager/include/application_manager/application.h#L890
- `ApplicationState` holds a vector of shared pointers of `HmiState`.
https://github.com/smartdevicelink/sdl_core/blob/10e17225d0a961f18493acd5ff8c1721e441cb14/src/components/application_manager/include/application_manager/application_state.h#L152
- `HmiState` has a shared pointer of `Application` in its member.
https://github.com/smartdevicelink/sdl_core/blob/10e17225d0a961f18493acd5ff8c1721e441cb14/src/components/application_manager/include/application_manager/hmi_state.h#L203

##### Reproduction Steps

Here are the repro steps to confirm that `ApplicationImpl` instance may not be released.

1. Check out latest `develop` branch (commit 10e17225d0a961f18493acd5ff8c1721e441cb14 )
2. Apply attached patch to add extra log output [0001-Add-printf-in-ApplicationImpl-ctor-and-dtor.patch.txt](https://github.com/smartdevicelink/sdl_core/files/1918359/0001-Add-printf-in-ApplicationImpl-ctor-and-dtor.patch.txt)
3. Update CMakeLists.txt file to enable `BUILD_TESTS` option
4. Give a build
5. Under the build directory, go to src/components/application_manager/test/
6. Run `./application_manager_test --gtest_filter=ApplicationImplTest.AddHmiState_GetCurrentState`

##### Expected Behavior

During the test, following logs should be observed:
```
>>> ApplicationImpl class constructor
>>> ApplicationImpl class destructor
```

##### Observed Behavior

Only `>>> ApplicationImpl class constructor` is shown. Please refer to attached log.
[application_impl_unit_test_output.txt](https://github.com/smartdevicelink/sdl_core/files/1918360/application_impl_unit_test_output.txt)

##### OS & Version Information

* OS/Version: Ubuntu 16.04 (64-bit)
[...]

Issue #2192, opened 2018-05-18
Number of threads grows when connecting/disconnecting from a video streaming app repeatedly


shoamano83:
### Bug Report

I noticed that the number of threads used by SDL Core grows if I repeat a testing cycle of connecting a video streaming app and disconnecting it. It might cause a problem on a system where maximum number of threads are tightly limited, and SDL Core runs for a long time.

##### Reproduction Steps

1. Start HMI
2. Start SDL Core on debugger:
```
$ LD_LIBRARY_PATH=. gdb ./smartDeviceLinkCore
```
3. Launch a video streaming app on the phone. Let it connect to Core through TCP transport.
4. Make the app FULL level by selecting it on HMI
5. HMI shows dialogs asking to start video and audio streamings. *Don't answer to them.*
6. The app will be automatically unregistered. (It may be registered again shortly after, but it doesn't matter.)
7. Quit the app on the phone
8. Repeat steps 3.-7. for one more time
9. Suspend SDL Core by sending a SIGTRAP from another terminal (`$ kill -TRAP pid_of_sdl_core`)
10. On the debugger terminal, run `info threads` and check the number of threads

##### Expected Behavior

There should not be duplicate thread entries in the list.

##### Observed Behavior

Duplicate thread entries are shown in the list:
- "CloseNaviAppTim" threads
- "HMI levels time" threads
- "VideoStreamSusp" threads

```
  39   Thread 0x7fffa77ee700 (LWP 12184) "HMI levels time" pthread_cond_timedwait@@GLIBC_2.3.2 ()
    at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_timedwait.S:225
  43   Thread 0x7fffa5feb700 (LWP 12193) "VideoStreamSusp" pthread_cond_wait@@GLIBC_2.3.2 ()
    at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
  47   Thread 0x7fffa4fe9700 (LWP 12199) "HMI levels time" pthread_cond_timedwait@@GLIBC_2.3.2 ()
    at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_timedwait.S:225
  58   Thread 0x7fffa47e8700 (LWP 12228) "HMI levels time" pthread_cond_timedwait@@GLIBC_2.3.2 ()
    at ../sysdeps/unix/sysv/linux/x86_64/pthread_cond_timedwait.S:225
[...]

PR #2225, opened 2018-05-31
fix: remove shared pointer of Application from HmiState

shoamano83:
Fixes https://github.com/smartdevicelink/sdl_core/issues/2143

### Risk
This PR makes **minor** API changes:
- re-enable once deprecated public methods in HmiState class. (Since I've reverted the API change, the old methods don't need to be deprecated.)

### Testing Plan
- Run unit tests and make sure the change doesn't cause any failures.
- Run the sequence described in https://github.com/smartdevicelink/sdl_core/issues/2143 and make sure ApplicationImpl class' destructor is called

### Summary
This PR reverts the introduction of shared pointer of Application by commit https://github.com/smartdevicelink/sdl_core/commit/a6791129fb1cfcd5dad96e4987b1c2664aafb61e.
The code stores and uses app_id value as it did in Core 4.4.

##### Breaking Changes
* None.

##### Enhancements
* None.

##### Bug Fixes
* Fixes memory leak issue caused by cyclic reference of shared pointer

### Tasks Remaining:
- None.

### CLA


shoamano83:
Hi @JackLivio, @jacobkeeler, please kindly review.
Also, I think the issue is quite severe. Could you consider giving high priority on the issue?

Hi @dev-gh, @AByzhynar, it would be great if you could also review the change and make sure I don't break anything of transport switching feature. I ran unit tests and confirmed that the change doesn't make any failures, but I am not 100% sure.

Thanks in advance!

shoamano83:
By the way, the unit test of `protocol_handler` sometimes fails (it's not related to this PR). That issue is fixed by commit https://github.com/XevoInc/sdl_core/commit/a2dce7f6b5b681ea9012b12d0c72341acb2c1e4e, which is included in PR https://github.com/smartdevicelink/sdl_core/pull/2164.

AByzhynar:
@AKalinich-Luxoft please review it as it looks that this changes can possibly break the fix you provided in the past. Please check it carefully.

AKalinich-Luxoft:
@AByzhynar @shoamano83 
I did exact the same fix in Sync3.2v2 for resolving cyclic dependency for `ApplicationImpl` class. On Sync3.2v2 my fix works well and I tested it. I will propose you to do following updates in your pull request:
- In `HmiState` class please keep `hmi_app_id ` instead of `app_id` because the last one is changed between ignition cycles and after resumption process. So `app_id` of application could be not actual in some cases and you will not get a pointer to `ApplicationImpl` using it. If you will store `hmi_app_id`, than you will be able to get `ApplicationImpl` instance using `app_mngr_.application_by_hmi_app()`
- I will propose you to pass `hmi_app_id` instead of shared pointer to `HmiState` class as it is actually not used there and to be sure it will not be copied in `HmiState` classes

src/components/application_manager/include/application_manager/hmi_state.h
```diff
@@ -200,7 +198,7 @@ class HmiState {
   }
 
  protected:
-  utils::SharedPtr<Application> app_;
+  uint32_t app_id_;
```
AKalinich-Luxoft: @shoamano83 according to my comment above, propose to rename to `hmi_app_id_`

src/components/application_manager/src/hmi_state.cc
```diff
@@ -40,7 +40,7 @@ namespace application_manager {
 HmiState::HmiState(utils::SharedPtr<Application> app,
                    const ApplicationManager& app_mngr,
                    StateID state_id)
-    : app_(app)
+    : app_id_(app->app_id())
```
AKalinich-Luxoft: @shoamano83 according to my comment above, propose to pass `app->hmi_app_id()` value directly, instead of passing the shared pointer

src/components/application_manager/src/hmi_state.cc
```diff
@@ -49,53 +49,54 @@ HmiState::HmiState(utils::SharedPtr<Application> app,
...
+    , system_context_(mobile_apis::SystemContext::INVALID_ENUM) {}
 
 void HmiState::set_parent(HmiStatePtr parent) {
   DCHECK_OR_RETURN_VOID(parent);
   parent_ = parent;
 }
 
 bool HmiState::is_navi_app() const {
-  return app_->is_navi();
+  const ApplicationSharedPtr app = app_mngr_.application(app_id_);
```
AKalinich-Luxoft: @shoamano83 according to my comment above, propose to use `application_by_hmi_app()` instead

src/components/application_manager/include/application_manager/hmi_state.h
shoamano83: Updated with commit 14270c9b11d5ac824f1d615d8b2f36277d573b53.

src/components/application_manager/src/hmi_state.cc
shoamano83: Updated with commit 14270c9b11d5ac824f1d615d8b2f36277d573b53.

src/components/application_manager/src/hmi_state.cc
shoamano83: Updated with commit 14270c9b11d5ac824f1d615d8b2f36277d573b53.

shoamano83:
Great @AKalinich-Luxoft, thank you very much!

I have updated the code to reflect your comments. I also updated state_controller_test.cc to add mock function calls regarding hmi_app_id.

I ran the unit tests repeatedly for 10 times and confirmed that I didn't get any failure. (Note that I applied commit https://github.com/XevoInc/sdl_core/commit/a2dce7f6b5b681ea9012b12d0c72341acb2c1e4e to avoid the issues in protocol_handler.)

AByzhynar:
@shoamano83 Check unit tests please

shoamano83:
The gdb version `GNU gdb (Ubuntu 7.7.1-0ubuntu5~14.04.3) 7.7.1` suggested that Jenkins ran on Ubuntu 14.04. So I set up a virtual machine of Ubuntu 14.04 (amd64) and managed to reproduce the issue.

The reason of the crash was that `directory_name` member variable in `ApplicationImplTest` was destructed prior to `app_impl`, so that `ApplicationImpl`'s destructor would access to the freed memory through `application_manager_.get_settings().app_storage_folder()` mock call. This is an implementation issue of unit tests and is NOT an issue of `ApplicationImpl` implementation.

FYI, attached is an output of valgrind describing this issue. [valgrind_ubuntu14.04_getcurrentstate.txt](https://github.com/smartdevicelink/sdl_core/files/2087534/valgrind_ubuntu14.04_getcurrentstate.txt)


I fix this issue with commit https://github.com/smartdevicelink/sdl_core/pull/2225/commits/047021eeb657b4e9726b9356602fe4f356ff4efa. I am not sure why I don't get the same issue on Ubuntu 16.04. It might be dependent on gcc version (4.8.4 on Ubuntu 14.04, 5.4.0 on Ubuntu 16.04.)

shoamano83:
Then I hit the issue in protocol handler unit test which I described before:

http://opensdl-jenkins.luxoft.com:8080/job/OpenSDL_PR/1587/console
```
06:13:41 [ RUN      ] ProtocolHandlerImplTest.SendHeartBeat_Successful
06:13:41 /home/jenkins/workspace/OpenSDL_PR/src/components/protocol_handler/test/protocol_handler_tm_test.cc:1991: Failure
06:13:41 Actual function call count doesn't match EXPECT_CALL(transport_manager_mock, SendMessageToDevice(ExpectedMessage( FRAME_TYPE_CONTROL, FRAME_DATA_HEART_BEAT, PROTECTION_OFF, kControl)))...
06:13:41          Expected: to be called once
06:13:41            Actual: never called - unsatisfied and active
06:13:41 pure virtual method called
06:13:41 terminate called without an active exception
```
Again, this is NOT related to the PR. However, to make sure everything goes fine, I cherry-picked commit https://github.com/XevoInc/sdl_core/commit/a2dce7f6b5b681ea9012b12d0c72341acb2c1e4e. Now the unit tests go through without any failure.

@AKalinich-Luxoft, @AByzhynar, please kindly review again. Thanks.

shoamano83:
Now that https://github.com/smartdevicelink/sdl_core/pull/2252 has been merged to `develop`, I reverted the commit to fix protocol handler unit test as this is irrelevant to the original issue.

I also noticed a build failure after merging latest `develop` branch. I fixed it with commit https://github.com/smartdevicelink/sdl_core/pull/2225/commits/951356fdfe627892ad9c8ffc0f208b1f5636988e.

Again, the PR is ready for review.

shoamano83:
Hi @LuxoftAKutsan, sorry for confusion.

This PR addresses issue https://github.com/smartdevicelink/sdl_core/issues/2143.

After Core 4.4, `HmiState` class added a shared pointer of `Application`. Because of this change, a cyclic reference of the shared pointer has been introduced.
i.e. an `Application` instance holds a reference to `ApplicationState` instance, which holds a list of `HmiState` instances, each of which holds a reference of `Application`. So the reference counter of the shared pointer of `Application` will never hit zero.

The cyclic reference problem prevents any `ApplicationImpl` instance from being destructed, causing several issues such as memory leaks and resource leaks (e.g. https://github.com/smartdevicelink/sdl_core/issues/2192).

The fix for this issue is to revert the shared pointer. @AKalinich-Luxoft suggested to use `hmi_app_id` instead, so I used that id in the fix.

After the fix, jenkins reported a crash issue in the unit test. I found the root cause was because of improper implementation of the unit test. It has been fixed with https://github.com/smartdevicelink/sdl_core/pull/2225/commits/047021eeb657b4e9726b9356602fe4f356ff4efa.

I mentioned about https://github.com/smartdevicelink/sdl_core/pull/2252 just because I saw another jenkin's failure due to the protocol handler issue (see http://opensdl-jenkins.luxoft.com:8080/job/OpenSDL_PR/1587/console). To make sure that all unit tests pass, I had included commit https://github.com/smartdevicelink/sdl_core/pull/2225/commits/5a1cbef5fe74cb90466b2e186211f9c8242fdd89 which was irrelevant of the original issue. I have reverted it in the latest version.

src/components/application_manager/src/hmi_state.cc
```diff
@@ -305,7 +319,9 @@ std::ostream& operator<<(std::ostream& os, const HmiState::StateID src) {
 }
 
 std::ostream& operator<<(std::ostream& os, const HmiState& src) {
-  os << "HMIState(app id:" << src.app_->app_id() << ", state:" << src.state_id()
+  const ApplicationSharedPtr app =
+      src.app_mngr_.application_by_hmi_app(src.hmi_app_id_);
+  os << "HMIState(app id:" << app->app_id() << ", state:" << src.state_id()
```
AKalinich-Luxoft: @shoamano83 I would like to print out `hmi_app_id` as well as after your updates it will be widely used inside `HmiState`

shoamano83: Sure, updated with commit https://github.com/smartdevicelink/sdl_core/pull/2225/commits/2f9108b62aed4679768cd2fff1b04887990bc47a.

AByzhynar:
@shoamano83 Please rebase to latest develop branch

shoamano83:
I still see failures in `state_controller_test` unit test. The reason is that `RegularHmiState()` method wasn't invoked. I will continue looking into it.

shoamano83:
After investigation I found the test cases were not correct. `RegularHmiState()` method was never called during the tests, so we shouldn't add expectations for it.

The tests were passing because the expectations were NOT verified, because the mock instance wasn't disposed. So when I checked out `develop` branch and added `Mock::VerifyAndClearExpectations(simple_app_ptr_);` at the end of the tests, they failed.

I fixed the issue with commit f539805f12c3d2fede46915aea35f6ed0e517eea by simply removing the inappropriate expectations. Now I don't see any unit test failures.

Please kindly review again. Thanks.

PR #2232, opened 2018-06-05
Adjust cppcheck and check-style for Travis

APipko:
This PR is **ready** for review.

### Risk
This PR makes **no** API changes.

### Summary
Travis only performs cppcheck and check-style in src/ without build.

### CLA
- [x] I have signed [the CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform)

PR #2243, opened 2018-06-06
Feature/working travis build

conlain-k:

### Risk

### Summary
Travis now builds SDL core in debug and runs `make install` and `make test`, followed by the style checking scripts. The style checking has been updated to match PR #2232. This PR makes the same changes, but also adds in a working core build-install-test cycle.
As of the build of the last commit, this full cycle runs in 35 min 43 sec. Travis has a maximum job time of 50 minutes, so that should not be a problem as long as builds don't get significantly longer. 

### CLA

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,53 +13,22 @@ addons:
   apt:
     sources:
     - ubuntu-toolchain-r-test
-    packages:
-    - gcc-4.9
-    - g++-4.9
-    - libssl-dev
-    - libbluetooth3
-    - libbluetooth-dev
-    - libudev-dev
-    - libusb-1.0
-    - cmake
-    - html2text
-    - clang-format-3.6
-    - binutils
+
 before_install:
-  - sudo apt-get -qq update
-  - sudo apt-get -q -y install bluez-tools sqlite3 libsqlite3-dev automake1.11
-  - wget http://archive.ubuntu.com/ubuntu/pool/main/l/lcov/lcov_1.11-3_all.deb
-  - sudo dpkg -i lcov_1.11-3_all.deb
+  - sudo apt-get update
+  - sudo apt-get install libssl-dev libbluetooth3 libbluetooth-dev libudev-dev cmake html2text lcov git cmake automake1.11 build-essential libavahi-client-dev sqlite3 libsqlite3-dev libgtest-dev bluez-tools libpulse-dev libusb-1.0.0-dev cppcheck
+  - sudo apt-get install -f clang-format-3.6
+
 script:
-  - sudo ln -sf /usr/bin/gcov-4.9 /usr/bin/gcov
-  - bash -e tools/infrastructure/check_style.sh
-  - echo "Number of processing units available ${PROCESSING_UNITS_COUNT}"
-  - mkdir build && cd build && cmake ../ -DBUILD_TESTS=ON -DENABLE_GCOV=ON && make install -j${PROCESSING_UNITS_COUNT}
-  - sudo ldconfig
-  - export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:`pwd`/bin/lib ; make test
-  - bash ../tools/infrastructure/show_disabled.sh 
-  - bash -ex ../tools/infrastructure/collect_coverage.sh ./
+  - mkdir build 
+  - pushd build
+  - cmake ../ -DBUILD_TESTS=ON
+  - make install && sudo ldconfig && make test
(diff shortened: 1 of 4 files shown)


LuxoftAKutsan:
@conlain-k I like this PR! 
The only think that may go wrong is that build time is close to travis limitation (1h) 
If travis will be overloaded (not so rare case) we will out of 1h limitations and build will fail, and we cant control travis resources. 

We already had such situation on this project. 

I would propose to keep travis for lightweight checks:
 - stylecheck
 - cppcheck static code analysis

And perform build and unit tests on self hosted jenkins : 
 - http://opensdl-jenkins.luxoft.com:8080/job/OpenSDL_PR/

Jenkins is rather powerful for such purpose ( build and unit tests  time ~ 21 min) and we may control it's capacities.

BTW this changes are already done in this PR : https://github.com/smartdevicelink/sdl_core/pull/2232

conlain-k:
Hello @LuxoftAKutsan,

From the documentation I can find, free public [Travis jobs cap out at 50 minutes](https://docs.travis-ci.com/user/customizing-the-build#Build-Timeouts). The build currently takes around 40 minutes, so this cap is certainly a concern. From the PR summary, 

> Travis has a maximum job time of 50 minutes, so that should not be a problem as long as builds don't get significantly longer.

Currently it consistently succeeds within that 50 minute limit. Since Travis uses public servers, that timing isn't entirely reliable, but I believe that the overload case is an exception, rather than the norm. Travis has had excellent [uptime and reliability over the last 90 days](https://www.traviscistatus.com/); I think that the risk of a failed build, while non-trivial, will occur sporadically at worst. 

Additionally, we can accelerate the build by using [caching](https://docs.travis-ci.com/user/caching/#ccache-cache) or by multithreaded builds via `make -j N`. Both of these have their drawbacks, but both provide significant speedups in build times in the event that builds become too long. 

This PR currently just runs a `make` - `make test` cycle and style checking, no ATF tests, which take a lot more time. Jenkins will still provide that ATF functionality. 

Does the Jenkins build run on all pushes? Travis seems to build on either a per-commit or per-push basis, providing a more 'continuous' CI. 

Travis also lives on free hosted servers, matching the OSS ideology of SDL and indeed the rest of the SDL repos (sdl_android and sdl_ios both use Travis).  

>  The style checking has been updated to match PR #2232.

I did indeed lift the code-checking changes, just adding the build-testing on top.  

I see no problem with redundant build-testing. Travis is free and provides an open and public portal, what's not to like?

LuxoftAKutsan:
@conlain-k I agree with this concerns. 
But we already had  experience with travis on sdl_core project before. And there are some cases when build time close to limitations may be easily exceeded. 
On travis we can do make -j2 (this is free limitation), I will speedup build a little bit, but not in 2 times.  
Jenkins have all the same triggers (Push, PR, daily, nightly weekly, ... ). And may perform build much faster.

> I see no problem with redundant build-testing. Travis is free and provides an open and public portal, what's not to like?

So do I. But the thing is that I would rely like to make development and master branch protected (by travis) to prevent merge commits that breaks by style or contains other dummy errors.  
Right now, SDL unit tests are not stable enough to make them as trigger for allowing merging PR. There are some randomly failing UT's. And rebuilding on travis will take additional time, that may be annoying in branch should be merged as soon as possible.

That is why I propose first to make travis perform checking couple of fast checks (style, cppcheck), is stable and protect development branch. 
When we will stabilize our unit test passing and probaby improve build time of SDL ( it may be faster, BTW), we may insert in travis sdl building and unit tests run.

jacobkeeler:
@LuxoftAKutsan I don't think that we should make any checks necessary for merge, things can go wrong every now and then with CI no matter how we scope this. Making the check necessary just seems like a very unnecessary limitation considering how simple it is to check the result of the build before merging. In any case, it would be Livio's responsibility to make sure nothing is merged with a failing check without good reason.

If you don't want the build/unit tests to be a factor in whether the check passes, however, these could be moved to the `after-success` step of the job. Including the build somewhere in the job would be more useful in most cases to the average contributor to this project, as it is hosted independently from any group involved in the project.

conlain-k:
I have moved the style checks to the top and the `make test` to the `after-success` step so that the build will not mark as failed if one of the tests fail. This builds in around 40 minutes without any threading or caching.

LuxoftAKutsan:
@conlain-k well, I think I am OK with it.
BWT, this changes a rely desired, when we will be able to put it in develop?

PR #2286, opened 2018-06-20
avoid returning null hmiLevel

### LINKED ISSUE #2285: OnHMIStatus returns null hmiLevel if RPC has invalid/incomplete parameters
### Bug Report
SDL core will return incorrect JSON in the response for an OnHMIStatus RPC if the RPC doesn't have the correct parameters.
```
{
   "hmiLevel" : null
}
```

A message uses an internal C++ map. Since maps use default-construction for accessing non-existent entries, the use of the `[]` accessor creates an invalid `hmiLevel` which is then passed to the mobile device.

This means that doing something like `Map[strings::msg_params][strings::hmi_level]` will create an entry if none exists. This invalid entry translates to extra JSON that lacks required fields and has a `null` HMI level.

##### Reproduction Steps
Connect any app to SDL and make an OnHMIStatus RPC with a missing parameter (for example, the new `videoStreamingState` parameter). SDL core will try to validate the parameters and catch the missing one, causing it to return `INVALID_DATA` and report a failure. However, before forwarding this message to the mobile device it attempts to perform checks on the `hmiLevel` field of the message. This creates an invalid HMI level which is then passed on the mobile device as a `null` JSON entry.

##### Expected Behavior
Core should return `INVALID_DATA` and no hmi level or other information.

##### Observed Behavior
Core returns `INVALID_DATA` but also returns an incorrect hmi level parameter.

* OS/Version: Ubuntu 18.04
* SDL Core Version: develop
* Testing Against: iOS app, Generic & SDL HMIs

Unless the field is checked to exist first, `Map[strings::msg_params][strings::hmi_level]` will cause such a field to exist, even when not desired.

conlain-k:
Fixes #2285.

### Risk

### Testing Plan
run smoke tests, unit tests, verify correct behavior on both Android and iOS with poorly-formed OnHMIStatus calls.

### Summary
Add validity check to avoid default-constructed invalid hmi level.

##### Bug Fixes
* Stops core from inadvertently returning ```{
   "hmiLevel" : null
}``` JSON as an OnHMIStatus notification.

### CLA

### DIFF
--- a/src/components/application_manager/src/commands/mobile/on_hmi_status_notification.cc
+++ b/src/components/application_manager/src/commands/mobile/on_hmi_status_notification.cc
@@ -57,6 +57,16 @@ void OnHMIStatusNotification::Run() {
     return;
   }
 
+  // If the RPC was bad, give up now and don't try to check hmi level
+  if ((*message_)[strings::msg_params][strings::result_code] ==
+      mobile_apis::Result::INVALID_DATA) {
+    SendNotification();
+    // we failed due to invalid data, don't do anything else
+    return;
+  }
+
+  // NOTE c++ maps default-construct on the [] operator, so if there is no
+  // hmiLevel field this will create one that is invalid
   mobile_apis::HMILevel::eType hmi_level =
       static_cast<mobile_apis::HMILevel::eType>(
           (*message_)[strings::msg_params][strings::hmi_level].asInt());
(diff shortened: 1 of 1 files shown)


conlain-k:
It turns out that core sends an `OnHMIStatusNotification` in response to an `OnHMIStatusNotificationFromMobile` if the mobile makes an OnHMIStatus notification with a bad set of parameters. It shouldn't be responding to the notification anyways, but now the map checks if the hmi level is set in the message before sending it.

PR #2302, opened 2018-06-26
Create third-party cmake targets if none exists

### LINKED ISSUE #2301: CMake doesn't create Boost/libbson targets if already found
### Bug Report
CMake reports warnings and fails the `install-3rd_party` target if libbson or Boost is already installed on the system. 

##### Reproduction Steps
1. checkout develop on a system that has Boost or libbson already installed
2. run CMake

##### Expected Behavior
CMake will find libbson/Boost and provide some target that other CMake targets can use as a dependency (like the `install-3rd_party` target).

##### Observed Behavior
CMake cannot find any such target and reports warnings. Any uses of the expected target result in `make` failing.

Below is the CMake warning caused by this.

```CMake Warning (dev) at src/components/application_manager/CMakeLists.txt:477 (add_dependencies):
  Policy CMP0046 is not set: Error on non-existent dependency in
  add_dependencies.  Run "cmake --help-policy CMP0046" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.

  The dependency target "libbson" of target "ApplicationManager" does not
  exist.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

* OS/Version: Ubuntu 18.04
* SDL Core Version: current develop
* Testing Against: cmake version 3.10.2

Probable fix is to create empty targets for these if the library is found so that, for example, a later target that depends on Boost can safely list it as a dependency.

conlain-k:
Fixes #2301 

### Risk

### Testing Plan
Try different make targets to ensure they all build correctly with dependencies.
Shouldn't need to test core itself since it only affects configuration

### Summary
If Boost or libbson are found on a system, a default CMake target is created to match the one that would have been created in their absence. This allows other targets (Utils, etc.) to list Boost/libbson as dependencies for proper Makefile ordering. 
Should have no portability concerns.

##### Bug Fixes
* fixes documented bug

### CLA

### DIFF
--- a/src/3rd_party/CMakeLists.txt
+++ b/src/3rd_party/CMakeLists.txt
@@ -201,6 +201,7 @@ if (${BSON_LIB} MATCHES "BSON_LIB-NOTFOUND")
       BUILD_COMMAND make
       INSTALL_COMMAND ${BSON_INSTALL_COMMAND})
 else()
+  add_custom_target(libbson) # empty target, bson is already installed
   get_filename_component(BSON_LIBS_DIRECTORY ${BSON_LIB} DIRECTORY)
   get_filename_component(EMHASHMAP_LIBS_DIRECTORY ${EMHASHMAP_LIB} DIRECTORY)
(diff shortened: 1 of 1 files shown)


conlain-k:
changes are picked up between other commits and #2392, closing as duplicate

PR #2209, opened 2018-05-23
Feature/OnRCStatus notification

LitvinenkoIra:
### This implementation is based on [RPC design refactoring](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0168-rpc-design-refactoring.md), so this PR should be merged after https://github.com/smartdevicelink/sdl_core/pull/2250

Technical task : #1930 
This PR is **ready** for review.

This pull request contains the implementation of new OnRCStatus notification
(proposal: https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0106-remote-control-onRcStatus-notification.md)
For testing use the following test set:
https://github.com/smartdevicelink/sdl_atf_test_scripts/blob/feature/rc_OnRCStatus/test_sets/rc_OnRCStatus.txt

Pull request with ATF scripts: https://github.com/smartdevicelink/sdl_atf_test_scripts/pull/1911

For manual testing use:
HMI from : smartdevicelink/sdl_hmi#52
mobile application SPT from HockeyApp 20180213-Nightly-Android

### Risk
This PR makes **[major]** API changes.
### CLA
- [x] I have signed [the CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform)

PR #2306, opened 2018-06-27
Update OnRCStatus with a new allowed parameter

LitvinenkoIra:
### This implementation is based on OnRCStatus notification, so this PR should be merged after https://github.com/smartdevicelink/sdl_core/pull/2209

Technical task : #2244 

This pull request contains the implementation of Remote Control - Update OnRCStatus with a new allowed parameter
(proposal: https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0172-onRcStatus-allowed.md)

Test scripts: https://github.com/smartdevicelink/sdl_atf_test_scripts/pull/1952
For testing use the following test set:
https://github.com/smartdevicelink/sdl_atf_test_scripts/blob/feature/onRcStatus_allowed/test_sets/rc_OnRCStatus.txt

### DIFF
--- a/src/components/application_manager/rpc_plugins/rc_rpc_plugin/src/resource_allocation_manager_impl.cc
+++ b/src/components/application_manager/rpc_plugins/rc_rpc_plugin/src/resource_allocation_manager_impl.cc
@@ -326,10 +326,8 @@ void ResourceAllocationManagerImpl::SendOnRCStatusNotifications(
     auto rc_apps = RCRPCPlugin::GetRCApplications(app_mngr_);
     for (const auto& rc_app : rc_apps) {
       msg_to_mobile = CreateOnRCStatusNotificationToMobile(rc_app);
-      if (NotificationTrigger::RC_STATE_CHANGING == event) {
-        (*msg_to_mobile)[application_manager::strings::msg_params]
-                        [message_params::kAllowed] = is_rc_enabled();
-      }
+      (*msg_to_mobile)[application_manager::strings::msg_params]
+                      [message_params::kAllowed] = is_rc_enabled();
       rpc_service_.SendMessageToMobile(msg_to_mobile);
       msg_to_hmi = CreateOnRCStatusNotificationToHmi(rc_app);
       rpc_service_.SendMessageToHMI(msg_to_hmi);
--- a/src/components/application_manager/rpc_plugins/rc_rpc_plugin/src/resource_allocation_manager_impl.cc
+++ b/src/components/application_manager/rpc_plugins/rc_rpc_plugin/src/resource_allocation_manager_impl.cc
@@ -174,6 +174,8 @@ void ResourceAllocationManagerImpl::ProcessApplicationPolicyUpdate() {
       if (rc_extention) {
         rc_extention->UnsubscribeFromInteriorVehicleData(*module);
       }
+    }
+    if (!disallowed_modules.empty()) {
       SendOnRCStatusNotifications(
           NotificationTrigger::MODULE_ALLOCATION,
           utils::SharedPtr<application_manager::Application>());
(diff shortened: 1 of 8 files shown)


src/components/application_manager/rpc_plugins/rc_rpc_plugin/src/resource_allocation_manager_impl.cc
```diff
@@ -213,10 +219,133 @@ void ResourceAllocationManagerImpl::RemoveAppsSubscriptions(const Apps& apps) {
...
+    DCHECK(application);
+    msg_to_mobile = CreateOnRCStatusNotificationToMobile(application);
+    (*msg_to_mobile)[application_manager::strings::msg_params]
+                    [message_params::kAllowed] = is_rc_enabled();
+    rpc_service_.SendMessageToMobile(msg_to_mobile);
+  } else {
+    auto rc_apps = RCRPCPlugin::GetRCApplications(app_mngr_);
+    for (const auto& rc_app : rc_apps) {
+      msg_to_mobile = CreateOnRCStatusNotificationToMobile(rc_app);
+      if (NotificationTrigger::RC_STATE_CHANGING == event) {
```
JackLivio: I believe that the allowed parameter should be included for all reasons including `MODULE_ALLOCATION`

LitvinenkoIra: @JackLivio As allowed parameter is **not mandatory**, so conditions for it sending must be specified, please take a look at https://github.com/smartdevicelink/sdl_requirements/issues/84

JackLivio: The requirement you linked to states 
> SDL does not sends allowed parameter by module allocation/deallocation.

But that behavior is not defined anywhere in the actual proposal.

LitvinenkoIra: @JackLivio Yes, I agree with you.
@yang1070 What is your opinion regarding this case?

yang1070: @JackLivio The intention of the proposal is to let the app know the event change on HMI side that the user enables or disables the RC.  the newly added parameter is specially for this change event. The proposal itself does not clearly define whether the parameter shall be include or not in case of resource allocation/dis-allocation. I think that implies a not including because there is no change on it. 
However, always include the parameter is also OK for me, as that makes the SDL logic simpler.

JackLivio: @yang1070 Thank you for the response, I think I would prefer to keep the logic simpler. It was also causing some confusion on the proxy side when our team was conducting the review.

LuxoftAKutsan: @JackLivio shouldn't we make this parameter mandatory in that case?

LitvinenkoIra:
@JackLivio Fixed 
e1e966e

src/components/interfaces/MOBILE_API.xml
```diff
@@ -6345,6 +6345,13 @@
 
     <function name="OnRCStatus" functionID="OnRCStatusID" messagetype="notification">
       <description>Issued by SDL to notify the application about remote control status change on SDL</description>
+      <param name="allowed" type="Boolean" mandatory="false">
+        <description>
+            If "true" - RC is allowed; if "false" - RC is disallowed.
+            Not present in notification in case by module allocation/deallocation.
```
JackLivio: This description is incorrect. 

Please use description used in the evolution [proposal](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0172-onRcStatus-allowed.md)

> <description>If "true" - RC is allowed; if "false" - RC is disallowed.</description>

LitvinenkoIra: @JackLivio Fixed 
8cfa86d

Issue #1523, opened 2017-05-04
[SDL 0044] Use Boost library in SDL


theresalech:
Proposal: [Use Boost library in SDL](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0044-use-Boost-library.md)

> This proposal is to use Boost library v1.56.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/151

Steering Committee Decision:

> The Steering Committee has noted that there may be overlapping benefits of Boost and C++11 once C++11 has been implemented. However, as a date for C++11 implementation has not yet been determined, the benefits of using Boost library in the interim warrant this proposal's acceptance.

PR #2392, opened 2018-07-18
Feature/boost lock implementation

conlain-k:
Addresses #1523

### Risk

### Testing Plan
Test with unit tests and ATF to ensure no regressions occur

### Summary
Use Boost instead of pthread for lock implementations. 
Remove old boolean constructor of `Lock` wrapper in favor of a separate `RecursiveLock` class. Adds a new `BaseLock` that is an abstract superclass of both `Lock` and `RecursiveLock` so that they can be used polymorphically.
Change over condition variables to boost implementation as well since the two are related.

##### Enhancements
* Boost locks are less forgiving and more verbose on error, meaning any synchronization issues will be much more apparent.

### CLA


conlain-k:
Since this changes the boost 3rd party install by adding more components, it may require that boost versions previously installed by core be removed and reinstalled

PR #2503, opened 2018-08-09
SDL is crashed when HMI responds DialNumber(info value is not string)

### LINKED ISSUE #2432: SDL is crashed when HMI responds DialNumber(info value is not string)
### Bug Report
 SDL is crashed when HMI responds DialNumber(info value is not string)

### Precondition:
DialNumber rpc is allowed in policy table.
App is activated.

##### Reproduction Steps
1. MOB -> SDL:DialNumber()
2. SDL -> BC:DialNumber()
3. BC > SDL:DialNumber(SUCCESS, info = 1111)

##### Expected Behavior
SDL ->MOB:DialNumber(SUCCESS) without "info" parameter.

##### Observed Behavior
SDL is crashed.

* OS/Version:
* SDL Core Version:
* Testing Against:

ValeriiMalkov:
Fixes #2432 

### Risk

### Summary
If `info` param is present in the response and it is not a string, `.asCharArray()` would crash. Added a check that the `info` param is a string.

### CLA


src/components/application_manager/rpc_plugins/sdl_rpc_plugin/test/commands/mobile/dial_number_request_test.cc
```diff
@@ -180,6 +180,38 @@ TEST_F(DialNumberRequestTest, OnEvent_SUCCESS) {
...
+
+  return !is_info_present && is_result_code_correct;
+}
+
+TEST_F(DialNumberRequestTest, OnEvent_InfoIsNotString_InfoDropped) {
+  MessageSharedPtr command_msg(CreateMessage(smart_objects::SmartType_Map));
+  (*command_msg)[am::strings::params][am::strings::connection_key] =
+      kConnectionKey;
+  DialNumberRequestPtr command(CreateCommand<DialNumberRequest>(command_msg));
+  MockAppPtr mock_app(CreateMockApp());
```
ZhdanovP: @ValeriiMalkov I have noticed that mock_app has the same creation process for each test, it's seems better to move this creation to SetUp method

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/dial_number_request.cc
```diff
@@ -130,7 +130,9 @@ void DialNumberRequest::on_event(const event_engine::Event& event) {
 
   const bool is_success = mobile_apis::Result::SUCCESS == result_code;
   const bool is_info_valid =
-      message[strings::msg_params].keyExists(strings::info);
+      message[strings::msg_params].keyExists(strings::info) && 
+      smart_objects::SmartType_String ==
+          message[strings::msg_params][strings::info].getType();
```
ZhdanovP: @ValeriiMalkov It looks like a local fix. It checks type of info only for DialNumberRequest, if for instance the wrong type of info comes to another requester the crash can repeat. I think we need to validate info type before the specific response created.

LuxoftAKutsan: Agree we should check the type in info parameter while validating smart object

ValeriiMalkov: @LuxoftAKutsan @ZhdanovP Please see fix in second commit 9cd0665c0fd1efad238adbd8214ef572dbe5af1d

src/components/application_manager/src/rpc_handler_impl.cc
```diff
@@ -329,6 +329,11 @@ bool RPCHandlerImpl::ConvertMessageToSO(
 
       rpc::ValidationReport report("RPC");
 
+      if (output["msg_params"].keyExists(strings::info) &&
```
LuxoftAKutsan: @ValeriiMalkov  remove hardoded strings

ValeriiMalkov: @LuxoftAKutsan Fixed and forced updated

GetmanetsIrina:
Issue is not relevant on
SDL: https://github.com/smartdevicelink/sdl_core/tree/6.1.0
Script: smartdevicelink/sdl_atf_test_scripts#2008

PR #2207, opened 2018-05-23
Feature/New remote control modules and parameters

LitvinenkoIra:
Technical task : #1798 
This PR is **ready** for review.

This pull request contains the implementation of new remote control modules and parameters
(proposal: https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0099-new-remote-control-modules-and-parameters.md)
For testing use the following test set:
https://github.com/smartdevicelink/sdl_atf_test_scripts/blob/feature/rc_new_modules/test_sets/rc_AUDIO_LIGHT_HMI_SETTINGS.txt

Pull request with ATF scripts: https://github.com/smartdevicelink/sdl_atf_test_scripts/pull/1909

For manual testing use:
HMI from : smartdevicelink/sdl_hmi#52
mobile application SPT from HockeyApp 20180213-Nightly-Android

### Risk
This PR makes **[major]** API changes.
### CLA
- [x] I have signed [the CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform)

PR #2510, opened 2018-08-14
Review for #2207 New remote control modules

LuxoftAKutsan:
This pr is for review New remote control modules PR: #2207


LuxoftAKutsan:
Not actual

PR #2530, opened 2018-08-21
Send RAI response on last received response of resumption data

LuxoftAKutsan:


src/components/application_manager/include/application_manager/resumption/resume_ctrl.h
```diff
@@ -147,7 +147,9 @@ class ResumeCtrl {
    * @return true if it was saved, otherwise return false
    */
   virtual bool StartResumption(app_mngr::ApplicationSharedPtr application,
-                               const std::string& hash) = 0;
+                               const std::string& hash,
+                               std::function<void(mobile_apis::Result::eType,
+                                                  const std::string&)> callback) = 0;
   /**
```
mked-luxoft: i think it's better to use type aliases for function objects. it makes code more readable.

LuxoftAKutsan: @mked-luxoft fixed 313dbaca7a

PR #2542, opened 2018-08-27
Fix sdl core dump

ZhdanovP:
Fix sdl core dump.
After the commit e270c43 the plugins in application manager began to not boot up.


jacobkeeler:
Already resolved by #2537

PR #2543, opened 2018-08-28
Fix mem. seg. fault error in jsoncpp lib

### LINKED ISSUE #2442: CCB: Fix mem. seg. fault error in jsoncpp lib
### Bug Report
Fix mem. seg. fault error in jsoncpp lib 

##### Description:
The root cause of the issue is a stack overflow which is occurring
due to recursive call of deeply nested areas and objects. We have
readValue(..)  which calls readObject(..) or readArray(..),
which call readValue(...). The starting point is a call to parse(..).
In order to fix the issue a count var. is implemented, which
measure how deep is our recursion. When recursion threshold is
exceeded we return invalid messege.

* OS/Version: 
* SDL Core Version: 
* Testing Against:

ZhdanovP:
Fixes #2442 

### Risk

### Summary
The root cause of the issue is a stack overflow which is occurring
due to recursive call of deeply nested areas and objects. We have
readValue(..)  which calls readObject(..) or readArray(..),
which call readValue(...). The starting point is a call to parse(..).
In order to fix the issue a count var. is implemented, which
measure how deep is our recursion. When recursion threshold is
exceeded we return invalid messege.

### CLA

### DIFF
--- a/src/3rd_party-static/jsoncpp/include/json/reader.h
+++ b/src/3rd_party-static/jsoncpp/include/json/reader.h
@@ -135,9 +135,9 @@ namespace Json {
       bool readCppStyleComment();
       bool readString();
       void readNumber();
-      bool readValue();
-      bool readObject( Token &token );
-      bool readArray( Token &token );
+      bool readValue( int& depth_level );
+      bool readObject( Token &token, int& depth_level );
+      bool readArray( Token &token, int& depth_level );
       bool decodeNumber( Token &token );
       bool decodeString( Token &token );
       bool decodeString( Token &token, std::string &decoded );
--- a/src/3rd_party-static/jsoncpp/include/json/reader.h
+++ b/src/3rd_party-static/jsoncpp/include/json/reader.h
@@ -4,211 +4,218 @@
 // See file LICENSE for detail or copy at http://jsoncpp.sourceforge.net/LICENSE
 
 #ifndef CPPTL_JSON_READER_H_INCLUDED
-# define CPPTL_JSON_READER_H_INCLUDED
+#define CPPTL_JSON_READER_H_INCLUDED
 
 #if !defined(JSON_IS_AMALGAMATION)
-# include "features.h"
-# include "value.h"
-#endif // if !defined(JSON_IS_AMALGAMATION)
-# include <deque>
-# include <stack>
-# include <string>
-# include <iostream>
+#include "features.h"
+#include "value.h"
+#endif  // if !defined(JSON_IS_AMALGAMATION)
+#include <deque>
+#include <stack>
+#include <string>
+#include <iostream>
 
 namespace Json {
 
-   /** \brief Unserialize a <a HREF="http://www.json.org">JSON</a> document into a Value.
-    *
-    */
-   class JSON_API Reader
-   {
-   public:
-      typedef char Char;
-      typedef const Char *Location;
+/** \brief Unserialize a <a HREF="http://www.json.org">JSON</a> document into a
+ *Value.
+ *
+ */
+class JSON_API Reader {
+ public:
+  typedef char Char;
+  typedef const Char* Location;
 
-      /** \brief Constructs a Reader allowing all features
-       * for parsing.
-       */
-      Reader();
-
-      /** \brief Constructs a Reader allowing the specified feature set
-       * for parsing.
-       */
-      Reader( const Features &features );
-
-      /** \brief Read a Value from a <a HREF="http://www.json.org">JSON</a> document.
-       * \param document UTF-8 encoded string containing the document to read.
-       * \param root [out] Contains the root value of the document if it was
-       *             successfully parsed.
-       * \param collectComments \c true to collect comment and allow writing them back during
-       *                        serialization, \c false to discard comments.
-       *                        This parameter is ignored if Features::allowComments_
-       *                        is \c false.
-       * \return \c true if the document was successfully parsed, \c false if an error occurred.
-       */
-      bool parse( const std::string &document, 
-                  Value &root,
-                  bool collectComments = true );
-
-      /** \brief Read a Value from a <a HREF="http://www.json.org">JSON</a> document.
-       * \param beginDoc Pointer on the beginning of the UTF-8 encoded string of the document to read.
-       * \param endDoc Pointer on the end of the UTF-8 encoded string of the document to read. 
-       \               Must be >= beginDoc.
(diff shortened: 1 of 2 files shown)


jacobkeeler:
@ZhdanovP Please sign the CLA

src/3rd_party-static/jsoncpp/include/json/reader.h
```diff
@@ -135,9 +135,9 @@ namespace Json {
       bool readCppStyleComment();
       bool readString();
       void readNumber();
-      bool readValue();
-      bool readObject( Token &token );
-      bool readArray( Token &token );
+      bool readValue( int *const depth_level );
```
AKalinich-Luxoft: @ZhdanovP why are you using pointers instead of refs here?

src/3rd_party-static/jsoncpp/src/lib_json/json_reader.cpp
```diff
@@ -164,11 +166,12 @@ Reader::parse( const char *beginDoc, const char *endDoc,
 
 
 bool
-Reader::readValue()
+Reader::readValue(int *const depth_level )
 {
    Token token;
    skipCommentTokens( token );
    bool successful = true;
+   const int  max_allowed_depth_level = 50;
```
AKalinich-Luxoft: @ZhdanovP why 50? Is it known at which depth level seg fault occurs? I don't think that 50 is a maximum

src/3rd_party-static/jsoncpp/src/lib_json/json_reader.cpp
```diff
@@ -180,24 +183,43 @@ Reader::readValue()
    switch ( token.type_ )
    {
    case tokenObjectBegin:
-      successful = readObject( token );
+
+      if(*depth_level == max_allowed_depth_level){
```
KVGrygoriev: @ZhdanovP I would recommend you don't use strict conditions, more reliably >=

src/3rd_party-static/jsoncpp/include/json/reader.h
KVGrygoriev: @ZhdanovP What the reason use "int *const", why you don't use "int &"?
It's much easier for understanding, isn't?

src/3rd_party-static/jsoncpp/src/lib_json/json_reader.cpp
```diff
@@ -180,24 +183,43 @@ Reader::readValue()
...
+        successful = false;
+        break;
+      }
+
+      ++(*depth_level);
```
KVGrygoriev: @ZhdanovP redundant difficultness

src/3rd_party-static/jsoncpp/src/lib_json/json_reader.cpp
```diff
@@ -140,8 +141,9 @@ Reader::parse( const char *beginDoc, const char *endDoc,
    while ( !nodes_.empty() )
       nodes_.pop();
    nodes_.push( &root );
-   
-   bool successful = readValue();
+
+   int depth_level = 0;
```
KVGrygoriev: @ZhdanovP What the reason to have depth_level? As I see you don't use it.
Why depth_level variable sets up inside readValue and this value returns to called function, I believe that adjusting callable function from outside more clear way

src/3rd_party-static/jsoncpp/src/lib_json/json_reader.cpp
ZhdanovP: @AKalinich-Luxoft actually we don't have any requirements for the depth as I know. We use 50 in Sync3 this cause that I took 50

src/3rd_party-static/jsoncpp/include/json/reader.h
LuxoftAKutsan: @ZhdanovP depth_level is an output parameter isn't it? Please name it properly and create description for the function. I would also create interface without this parameter. In most cases I don't case what was deep of nested structures if I was not able to read value.

src/3rd_party-static/jsoncpp/src/lib_json/json_reader.cpp
```diff
@@ -119,11 +119,12 @@ Reader::parse( std::istream& sin,
    return parse( doc, root, collectComments );
 }
 
-bool 
-Reader::parse( const char *beginDoc, const char *endDoc, 
+bool
```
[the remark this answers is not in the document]
LuxoftAKutsan: @ZhdanovP remind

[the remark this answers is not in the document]
ZhdanovP: @LuxoftAKutsan please take a look 7f65a348a6829dc4677ee89a11be05db0db9f6d2

src/3rd_party-static/jsoncpp/src/lib_json/json_reader.cpp
ZhdanovP: @KVGrygoriev please take a look 00c2f229612723f7acf2ba89e90f6c815dc1699b

src/3rd_party-static/jsoncpp/src/lib_json/json_reader.cpp
ZhdanovP: @KVGrygoriev We have added `depth_level` for control the depth of recursion during parsing. 
I'am not sure that we need to move this parameter outside because it's internal logic of Reader::parse

ZhdanovP:
@AKalinich-Luxoft 

>  I would also create interface without this parameter. In most cases I don't case what was deep of nested structures if I was not able to read value.

I can add it if you insist but I actually we use this methods from the `Reader::parse `not separately in our case, so I think this is redundant now.

src/3rd_party-static/jsoncpp/include/json/reader.h
```diff
@@ -4,211 +4,218 @@
...
-# include "features.h"
-# include "value.h"
-#endif // if !defined(JSON_IS_AMALGAMATION)
-# include <deque>
-# include <stack>
-# include <string>
-# include <iostream>
+#include "features.h"
+#include "value.h"
+#endif  // if !defined(JSON_IS_AMALGAMATION)
```
ValeriiMalkov: @ZhdanovP What for this extra space ?

iCollin:
Closed in favor of #3155

PR #2548, opened 2018-08-28
Set initialize state for the Interger type

### LINKED ISSUE #2440: SDL does not set initialize state for the Interger type
### Bug Report
SDL does not set initialize state for the Interger type

##### Description:
The default constructor of Interger class assigns the min value
for the wrapped variable, hoowever it keeps the initialize state us
`Uninitialized`. So in case when latter user apply such operations as
`++`, `+=` the value will be changed but the state is not.

##### Expected Behavior
SDL does set initialize state for the Interger type

* OS/Version: 
* SDL Core Version: 
* Testing Against:

ZhdanovP:
Fixes #2440 

### Risk

### Summary
The default constructor of Interger class assigns the min value
for the wrapped variable, hoowever it keeps the initialize state us
`Uninitialized`. So in case when latter user apply such operations as
`++`, `+=` the value will be changed but the state is not.

The commit adds ability to change initialize state for the class
after mentioned operations. So user does not have to call additionaly
`initialize` method.

### CLA

### DIFF
--- a/src/components/rpc_base/include/rpc_base/rpc_base_inl.h
+++ b/src/components/rpc_base/include/rpc_base/rpc_base_inl.h
@@ -186,12 +186,14 @@ Integer<T, minval, maxval>& Integer<T, minval, maxval>::operator=(
 template <typename T, T minval, T maxval>
 Integer<T, minval, maxval>& Integer<T, minval, maxval>::operator++() {
   ++value_;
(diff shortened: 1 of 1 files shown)


AByzhynar:
@ZhdanovP Typo in branch name, commit and PR description. Check & correct

PR #2562, opened 2018-09-04
Avoid deadlock in delete buttons

LuxoftAKutsan:
Reduce the scope of ButtonSubscriptions to avoid Deadlock un function ApplicationManager::UnsubscribeFromButton

### DIFF
--- a/src/components/application_manager/src/resumption/resumption_data_processor.cc
+++ b/src/components/application_manager/src/resumption/resumption_data_processor.cc
@@ -595,18 +595,16 @@ void ResumptionDataProcessor::DeleteSubscriptions(const int32_t app_id) {
 void ResumptionDataProcessor::DeleteButtonsSubscriptions(
     ApplicationSharedPtr application) {
   LOG4CXX_AUTO_TRACE(logger_);
-  DataAccessor<ButtonSubscriptions> button_accessor =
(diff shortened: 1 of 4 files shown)


src/components/application_manager/src/resumption/resumption_data_processor.cc
```diff
@@ -595,18 +595,16 @@ void ResumptionDataProcessor::DeleteSubscriptions(const int32_t app_id) {
...
   LOG4CXX_AUTO_TRACE(logger_);
-  DataAccessor<ButtonSubscriptions> button_accessor =
-      application->SubscribedButtons();
-  ButtonSubscriptions button_subscriptions = button_accessor.GetData();
-  while (!button_subscriptions.empty()) {
-    auto btn = button_subscriptions.begin();
+  ButtonSubscriptions button_subscriptions =
+      application->SubscribedButtons().GetData();
+  for (auto btn : button_subscriptions) {
+    auto hmi_btn = static_cast<hmi_apis::Common_ButtonName::eType>(btn);
```
ZhdanovP: May be `hmi_btn` make const?

LuxoftAKutsan: Fixed

PR #2575, opened 2018-09-06
Fix/resumption error handling fixes

LuxoftAKutsan:
83416aa203 Add sending UNsubscribeWaypoiints in case of reverting resumption
aa33dc1bcc Send correct Reset global properties in case of reverting resumption
38ad972f4d Do not unsubscribe from way points in case of ignition off
47b4291f98 Send UnsubscribeVehicleData during reverting resumption

### DIFF
--- a/src/components/application_manager/src/resumption/resumption_data_processor.cc
+++ b/src/components/application_manager/src/resumption/resumption_data_processor.cc
@@ -537,7 +537,16 @@ void ResumptionDataProcessor::DeleteWayPointsSubscription(
   LOG4CXX_AUTO_TRACE(logger_);
 
   if (application_manager_.IsAppSubscribedForWayPoints(application)) {
+    LOG4CXX_DEBUG(logger_, "App is subscribed");
     application_manager_.UnsubscribeAppFromWayPoints(application);
+    auto subscribe_waypoints_msg = MessageHelper::CreateMessageForHMI(
+        hmi_apis::FunctionID::Navigation_UnsubscribeWayPoints,
+        application_manager_.GetNextHMICorrelationID());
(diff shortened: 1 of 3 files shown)


src/components/application_manager/src/resumption/resumption_data_processor.cc
```diff
@@ -481,18 +481,29 @@ void ResumptionDataProcessor::DeleteGlobalProperties(
...
+        application_manager_.GetNextHMICorrelationID());
+    (*msg)[strings::params][strings::function_id] =
+        hmi_apis::FunctionID::UI_SetGlobalProperties;
+    (*msg)[strings::msg_params] = *msg_params;
+    ProcessHMIRequest(msg, false);
   }
   if (result.HasTTSPropertiesReset()) {
-    smart_objects::SmartObjectSPtr tts_gl_props_reset_req =
+    smart_objects::SmartObjectSPtr msg_params =
         MessageHelper::CreateUIResetGlobalPropertiesRequest(result,
```
ZhdanovP: @LuxoftAKutsan Do we need the same`CreateUIResetGlobalPropertiesRequest`  for TTS and UI Properties? If yes may be do it once above the if's construction?

mked-luxoft: @LuxoftAKutsan  it doesn't seem to be changed in this PR, but i think that under `result.HasTTSPropertiesReset()` condition, `MessageHelper::CreateTTSResetGlobalPropertiesRequest` should be called.

LuxoftAKutsan: @ZhdanovP @mked-luxoft  fixed

PR #2576, opened 2018-09-06
HMI does not respond during timeout

ZhdanovP:
Send RESUME_FAILED when HMI doesn't respond


src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -774,11 +789,14 @@ void RegisterAppInterfaceRequest::SendRegisterAppInterfaceResponseToMobile(
   ApplicationSharedPtr application = application_manager_.application(key);
 
   response_params[strings::sync_msg_version][strings::major_version] =
-      major_version;  // From generated file interfaces/generated_msg_version.h
+      major_version;  // From generated file
```
LuxoftAKutsan: @ZhdanovP please try to keep such changes in a separate commit

PR #2621, opened 2018-09-19
SDL crash during cleanup.

### LINKED ISSUE #2428: SDL crashes in case 'AppStorageFolder' doesn't have read/write permission
### Bug Report
SDL crashes in case 'AppStorageFolder' doesn't have read/write permission

##### Reproduction Steps
1. In case when there is no read/write permission
for the value of AppStorageFolder in smartDeviceLink.ini
the SDL crashes in initialization because the
assertion for the Protocol Handler instance fails.
The reason for this failure is that Protocol Handler
is not initialized due to failure of Application
Manager Init method caused by lack ot r/w permission.

##### Expected Behavior
SDL log an error and stop working correctly
There is no any dump

* OS/Version: 
* SDL Core Version:
* Testing Against:

mkorniichuk:
Fixes #2428 #1829

### Risk

### Testing Plan
Manually tested

### Summary
If resumption_storage_ in ResumeCtrlImpl is failed to initialize,
SDL will try to stop components and exit with error code.
At this time application manager tries to unregister all applications and
call ResumeCtrlImpl::OnIgnitionOff() where resumption_storage_ is used. As a result, a crash will occur.

As an example, if AppStorageFolder have no read/write access, resumption_storage_
fails to initialize.

### CLA

### DIFF
--- a/src/components/application_manager/src/resumption/resume_ctrl_impl.cc
+++ b/src/components/application_manager/src/resumption/resume_ctrl_impl.cc
@@ -336,6 +336,10 @@ void ResumeCtrlImpl::OnSuspend() {
 void ResumeCtrlImpl::OnIgnitionOff() {
   LOG4CXX_AUTO_TRACE(logger_);
   if (!application_manager_.IsLowVoltage()) {
+    if (!resumption_storage_) {
+      LOG4CXX_ERROR(logger_, "resumption_storage_ is not initialized");
+      return;
+    }
     resumption_storage_->IncrementIgnOffCount();
     resumption_storage_->ResetGlobalIgnOnCount();
     FinalPersistData();
@@ -572,6 +576,10 @@ void ResumeCtrlImpl::SaveDataOnTimer() {
 
 void ResumeCtrlImpl::FinalPersistData() {
   LOG4CXX_AUTO_TRACE(logger_);
+  if (!resumption_storage_) {
+    LOG4CXX_ERROR(logger_, "resumption_storage_ is not initialized");
+    return;
+  }
   StopSavePersistentDataTimer();
   SaveAllApplications();
   resumption_storage_->Persist();
--- a/src/components/application_manager/src/resumption/resume_ctrl_impl.cc
+++ b/src/components/application_manager/src/resumption/resume_ctrl_impl.cc
@@ -335,11 +335,8 @@ void ResumeCtrlImpl::OnSuspend() {
 
 void ResumeCtrlImpl::OnIgnitionOff() {
   LOG4CXX_AUTO_TRACE(logger_);
+  DCHECK_OR_RETURN_VOID(resumption_storage_);
   if (!application_manager_.IsLowVoltage()) {
-    if (!resumption_storage_) {
-      LOG4CXX_ERROR(logger_, "resumption_storage_ is not initialized");
-      return;
-    }
     resumption_storage_->IncrementIgnOffCount();
     resumption_storage_->ResetGlobalIgnOnCount();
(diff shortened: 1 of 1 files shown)


src/components/application_manager/src/resumption/resume_ctrl_impl.cc
```diff
@@ -336,6 +336,10 @@ void ResumeCtrlImpl::OnSuspend() {
 void ResumeCtrlImpl::OnIgnitionOff() {
   LOG4CXX_AUTO_TRACE(logger_);
   if (!application_manager_.IsLowVoltage()) {
+    if (!resumption_storage_) {
```
LuxoftAKutsan: @mkorniichuk do we have case if resumption storage may be nit initialized? If no> I would put DCHECK here

LuxoftAKutsan: @mkorniichuk in case if FS is read only, it should be handled inside resumption storage, isn't it? Or even in Last state.

mkorniichuk: @LuxoftAKutsan DCHECK will stop application immediately with logged fatal error. But we still need to stop all components, don't we?
About the FS. It is handled in application manager Init(). So the Init() of resumption_storage is not even called. Then StopComponents takes place, where Stop method is called for application manager, and so the OnIgnitonOff method of ResumptionController is called as well. And since resumption_storage is not initialized, a crash will occur.

LuxoftAKutsan: DCHECK is assertion. Assertion will be removed from release build. During testing if we found such condition, we can just crash SDL with appropriate information.

src/components/application_manager/src/resumption/resume_ctrl_impl.cc
```diff
@@ -572,6 +576,10 @@ void ResumeCtrlImpl::SaveDataOnTimer() {
 
 void ResumeCtrlImpl::FinalPersistData() {
   LOG4CXX_AUTO_TRACE(logger_);
+  if (!resumption_storage_) {
+    LOG4CXX_ERROR(logger_, "resumption_storage_ is not initialized");
+    return;
+  }
```
jacobkeeler, line 582:
```suggestion
  DCHECK_OR_RETURN_VOID(resumption_storage_);
```
Agreed with @LuxoftAKutsan

src/components/application_manager/src/resumption/resume_ctrl_impl.cc
```diff
@@ -336,6 +336,10 @@ void ResumeCtrlImpl::OnSuspend() {
 void ResumeCtrlImpl::OnIgnitionOff() {
   LOG4CXX_AUTO_TRACE(logger_);
   if (!application_manager_.IsLowVoltage()) {
+    if (!resumption_storage_) {
+      LOG4CXX_ERROR(logger_, "resumption_storage_ is not initialized");
+      return;
+    }
```
jacobkeeler, line 342:
```suggestion
    DCHECK_OR_RETURN_VOID(resumption_storage_);
```
Agreed with @LuxoftAKutsan

jacobkeeler, line 342:
```suggestion  
  LOG4CXX_AUTO_TRACE(logger_);
  DCHECK_OR_RETURN_VOID(resumption_storage_);
  if (!application_manager_.IsLowVoltage()) {
```
Agreed with @LuxoftAKutsan

PR #2765, opened 2018-11-22
Refactoring cmake structure

BSolonenko:

### Summary
SDL may use modern cmake approach of creating targets. It will simplify
porting SDL on any platform. For each third party library used by SDL
created .cmake file, that will expose library as cmake_library.

This will allow to make clearer dependencies, avoid dependency gaps
(required for multithreading compilation) and replace dependency during
porting to other platform just with replacing .cmake third party library
file.

### CLA


src/3rd_party/CMakeLists.txt
```diff
@@ -241,7 +243,6 @@ endif()
 
 add_custom_target(install-3rd_party
   DEPENDS ${install-3rd_party_logger_var}
-  DEPENDS libbson
   DEPENDS Boost
```
LuxoftAKutsan: @BSolonenko  boost still does not use namespaces

BSolonenko: @LuxoftAKutsan this will be fixed when I will create PR with refactoring CMakeLists for 3rd-party libraries.

Issue #2215, opened 2018-05-24
[SDL 0158] Cloud App Transport Adapter


theresalech:
Proposal: [Cloud App Transport Adapter](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0158-cloud-app-transport-adapter.md)

> This proposal will detail a possible solution for allowing SDL-enabled cloud applications to be used in a vehicle. This proposal will cover the process of obtaining cloud application server endpoints, enabling which cloud apps should appear in the HMI app list, opening connections between Core and a cloud based app server, and how a cloud application can authenticate connected head units.

Review Issue: smartdevicelink/sdl_evolution#493

Revisions to the review issue: https://github.com/smartdevicelink/sdl_evolution/issues/666

Steering Committee Decision:
> The Steering Committee voted to accept this proposal with revisions. The revisions will include changing the `app_icon_url` policy table attribute to `icon_url` as mentioned in [this comment](https://github.com/smartdevicelink/sdl_evolution/issues/666#issuecomment-459826599), include the affecting platforms based on the protocol changes mentioned in [this comment](https://github.com/smartdevicelink/sdl_evolution/issues/666#issuecomment-460424274), and rename `cloudAppAuthToken` to `authToken` noted in [this comment](https://github.com/smartdevicelink/sdl_evolution/issues/666#issuecomment-460661104).  
Additionally, the author stated that OEMs have the ability to specify secure websockets or unsecured websockets whether you have WS or WSS specified in `cloud_transport_type` in the policy table.

> Revisions have been made 

Additional revisions accepted on 2019-03-15:

- Make CloudAppProperties parameter `appName` non-mandatory. 
- Make CloudAppProperties parameter `appName` an array.
- Rename CloudAppProperties parameter `appName` to `nicknames`

jordynmackool:
Accepted revisions as of 2019-03-06: https://github.com/smartdevicelink/sdl_evolution/issues/666#issuecomment-461205796

Additional revisions accepted on 2019-03-15: https://github.com/smartdevicelink/sdl_evolution/pull/689

- Make CloudAppProperties parameter `appName` non-mandatory. 
- Make CloudAppProperties parameter `appName` an array.
- Rename CloudAppProperties parameter `appName` to `nicknames`

jacobkeeler:
Closing with the merge of #2853

PR #2822, opened 2019-02-25
Add auth token to StartServiceACK payload

jacobkeeler:
Partially implements #2215 

### Risk

### Testing Plan
TBD

### Summary
Added `authToken` field to StartServiceACK payload when connecting to cloud applications

##### Enhancements
* Adds `authToken` to StartServiceACK

### CLA

### DIFF
--- a/src/components/application_manager/src/policies/policy_handler.cc
+++ b/src/components/application_manager/src/policies/policy_handler.cc
@@ -1771,6 +1771,17 @@ void PolicyHandler::OnCertificateUpdated(const std::string& certificate_data) {
 }
 #endif  // EXTERNAL_PROPRIETARY_MODE
 
+void PolicyHandler::OnAuthTokenUpdated(const std::string& policy_app_id,
+                                       const std::string& auth_token) {
+  LOG4CXX_AUTO_TRACE(logger_);
+  sync_primitives::AutoLock lock(listeners_lock_);
+  HandlersCollection::const_iterator it = listeners_.begin();
+  for (; it != listeners_.end(); ++it) {
+    PolicyHandlerObserver* observer = *it;
+    observer->OnAuthTokenUpdated(policy_app_id, auth_token);
+  }
+}
+
 void PolicyHandler::OnPTUFinished(const bool ptu_result) {
   LOG4CXX_AUTO_TRACE(logger_);
   sync_primitives::AutoLock lock(listeners_lock_);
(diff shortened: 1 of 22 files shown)


src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -1909,15 +1920,18 @@ void PolicyHandler::OnSetCloudAppProperties(
...
   policy_manager_->InitCloudApp(policy_app_id);
 
+  bool auth_token_update = false;
   if (msg_params.keyExists(strings::enabled)) {
-    policy_manager_->SetCloudAppEnabled(policy_app_id,
-                                        msg_params[strings::enabled].asBool());
-
+    bool enabled = msg_params[strings::enabled].asBool();
+    policy_manager_->SetCloudAppEnabled(policy_app_id, enabled);
+    auth_token_update = enabled;
```
JackLivio: Why should auth_token_update be set to enabled in this case?

src/components/protocol_handler/src/protocol_handler_impl.cc
```diff
@@ -410,10 +411,22 @@ void ProtocolHandlerImpl::SendStartSessionAck(
...
                                                     kDisabledSecondary);
       }
+
+      std::string policy_app_id =
+          connection_handler_.GetCloudAppID(connection_id);
+      if (*min_version >= min_cloud_app_version && !policy_app_id.empty()) {
+        sync_primitives::AutoLock lock(auth_token_map_lock_);
+        auto it = auth_token_map_.find(policy_app_id);
+        if (it != auth_token_map_.end()) {
+          char auth_token[256];
```
JackLivio: authToken can be a lot longer that 256 chars.

```
      <param name="authToken" type="String" maxlength="65535" mandatory="false">
            <description>Used to authenticate websocket connection on app activation</description>
        </param>
```

src/components/application_manager/src/policies/policy_handler.cc
jacobkeeler: Updates are sent for all enabled apps when the policy table is loaded or updated. This logic is to cover the case where an app is enabled after the policy table is initially loaded, meaning that the listener does not already have the auth token for this app.

src/components/protocol_handler/src/protocol_handler_impl.cc
jacobkeeler: Good catch

jacobkeeler: Resolved in latest comment

PR #2833, opened 2019-03-05
Handle hybrid app preference

jacobkeeler:
Partially implements #2215 

### Risk

### Testing Plan
TBD

### Summary
Add handling for the `hybrid_app_preference` property of cloud apps

##### Enhancements
* If both a cloud app and mobile app attempt to register with the same name, the hybrid app preference of the cloud app is used to determine which app should be registered
    * `CLOUD` - Mobile app is unregistered (or not allowed to register)
    * `MOBILE` - HMI app is unregistered (or not allowed to register)
    * `BOTH` - The first app that attempts to register is registered, the second is not allowed to register

### CLA

### DIFF
--- a/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/register_app_interface_request.cc
+++ b/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/register_app_interface_request.cc
@@ -281,17 +281,49 @@ void RegisterAppInterfaceRequest::Run() {
     return;
   }
 
-  mobile_apis::Result::eType coincidence_result = CheckCoincidence();
+  ApplicationSharedPtr duplicate_app;
+  mobile_apis::Result::eType coincidence_result =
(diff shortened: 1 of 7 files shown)


src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -922,7 +954,8 @@ void RegisterAppInterfaceRequest::SendOnAppRegisteredNotificationToHMI(
   DCHECK(rpc_service_.ManageHMICommand(notification));
 }
 
-mobile_apis::Result::eType RegisterAppInterfaceRequest::CheckCoincidence() {
+mobile_apis::Result::eType RegisterAppInterfaceRequest::CheckCoincidence(
+    ApplicationSharedPtr& out_app) {
```
ShobhitAd: Any reason why the function parameter is called `app` in the definition and `out_app` over here?

PR #2835, opened 2019-03-09
Allow for unknown enum values for App Service RPCs

jacobkeeler:

### Risk

### Testing Plan
Manual test of GetAppServiceData

### Summary
Add functionality to allow for unknown enum values for App Service RPCs

##### Bug Fixes
* Allow unknown enum values for App Service RPCs

### CLA

### DIFF
--- a/src/components/application_manager/src/rpc_service_impl.cc
+++ b/src/components/application_manager/src/rpc_service_impl.cc
@@ -388,7 +388,7 @@ void RPCServiceImpl::SendMessageToMobile(
 
   const bool is_result_code_exists =
       (*message)[strings::msg_params].keyExists(strings::result_code);
-  bool remove_unknown_parameters = true;
+  bool allow_unknown_parameters = false;
 
   if (!app) {
     LOG4CXX_ERROR(logger_, "No application associated with connection key");
@@ -431,7 +431,7 @@ void RPCServiceImpl::SendMessageToMobile(
     LOG4CXX_DEBUG(logger_,
                   "Allowing unknown parameters for response function "
                       << function_id);
-    remove_unknown_parameters = false;
+    allow_unknown_parameters = true;
   }
 
   if (rpc_passing &&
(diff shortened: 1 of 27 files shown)


src/components/smart_objects/include/smart_objects/smart_object.h
```diff
@@ -680,7 +680,8 @@ class SmartObject FINAL {
    */
```
ShobhitAd: Could add a description for the new `allow_unknown_enums` param to the function brief

src/components/smart_objects/include/smart_objects/smart_schema.h
```diff
@@ -68,10 +68,11 @@ class CSmartSchema FINAL {
    * @param MessageVersion to check mobile RPC version against RPC Spec History
    * @return ns_smart_objects::errors::eType
```
ShobhitAd: Could add a description for the new `allow_unknown_enums` param to the function brief

src/components/application_manager/src/rpc_service_impl.cc
```diff
@@ -631,15 +631,15 @@ bool RPCServiceImpl::ConvertSOtoMessage(
     case 0: {
       if (protocol_version == 1) {
         if (!formatters::CFormatterJsonSDLRPCv1::toString(
-                message, output_string, remove_unknown_parameters)) {
+                message, output_string, !allow_unknown_parameters)) {
```
ShobhitAd: This is completely optional, but as you mentioned we should avoid negative statements for boolean values. `remove_unknown_params` is still used for the `applySchema`, `unapplySchema` and `toString` functions.

jacobkeeler: We can't do this for any functions that already had the parameter before the release, since that would fundamentally change the meaning of the parameter, meaning it would be a breaking change.
There are a few other places where the parameter was added in this release, but I'm not too worried about those, especially since most of those are on the same layer as the functions which already had the parameter. In this sense, the naming is at least consistent within the RPC formatter/SmartObject schema layer.
I think we can hold off changing the other instances of this until the next major release, where we can make everything consistent.

PR #2837, opened 2019-03-11
Validating forwarded and App Service RPC requests

ShobhitAd:

### Risk

### Testing Plan
Manual testing with changing existing params and sending new params in RPC requests to core.

### Summary
The current implementation allows RPC passing requests and App Service RPC requests to be used without validation. The requests forwarded to core (after PassThrough requests to App Services) are also not validated.

The changes in this PR add a validate param which is used to validate existing parameters for App Service RPCs. For RPC Passing, the RPC request forwarded to core(when no app services can handle the request) is revalidated.

### CLA

### DIFF
--- a/src/components/application_manager/src/rpc_handler_impl.cc
+++ b/src/components/application_manager/src/rpc_handler_impl.cc
@@ -85,7 +85,7 @@ void RPCHandlerImpl::ProcessMessageFromMobile(
   }
 
   if (!ConvertMessageToSO(
-          *message, *so_from_mobile, remove_unknown_parameters)) {
+          *message, *so_from_mobile, remove_unknown_parameters, !rpc_passing)) {
     LOG4CXX_ERROR(logger_, "Cannot create smart object from message");
     return;
   }
@@ -106,7 +106,7 @@ void RPCHandlerImpl::ProcessMessageFromMobile(
                                           commands::Command::SOURCE_MOBILE,
                                           message_type)) {
       // Since PassThrough failed, refiltering the message
-      if (!ConvertMessageToSO(*message, *so_from_mobile, true)) {
+      if (!ConvertMessageToSO(*message, *so_from_mobile, true, true)) {
         LOG4CXX_ERROR(logger_, "Cannot create smart object from message");
         return;
       }
@@ -283,7 +283,8 @@ void RPCHandlerImpl::GetMessageVersion(
 bool RPCHandlerImpl::ConvertMessageToSO(
     const Message& message,
     ns_smart_device_link::ns_smart_objects::SmartObject& output,
-    const bool remove_unknown_parameters) {
+    const bool remove_unknown_parameters,
+    const bool validate_existing_params) {
   LOG4CXX_AUTO_TRACE(logger_);
   LOG4CXX_DEBUG(logger_,
                 "\t\t\tMessage to convert: protocol "
@@ -303,7 +304,7 @@ bool RPCHandlerImpl::ConvertMessageToSO(
               message.type(),
               message.correlation_id());
 
-      rpc::ValidationReport report("RPC");
+      smart_objects::SmartObject* so_ptr = (conversion_result) ? &output : NULL;
 
       // Attach RPC version to SmartObject if it does not exist yet.
       auto app_ptr = app_manager_.application(message.connection_key());
@@ -316,30 +317,22 @@ bool RPCHandlerImpl::ConvertMessageToSO(
         GetMessageVersion(output, msg_version);
       }
 
-      if (!conversion_result ||
-          !mobile_so_factory().attachSchema(
-              output, remove_unknown_parameters, msg_version) ||
-          ((output.validate(&report, msg_version) !=
-                smart_objects::errors::OK &&
-            remove_unknown_parameters))) {
-        LOG4CXX_WARN(logger_,
-                     "Failed to parse string to smart object with API version "
-                         << msg_version.toString() << " : "
-                         << message.json_message());
+      if (validate_existing_params) {
+        if (!ValidateRPCSO(so_ptr,
+                           message.connection_key(),
+                           message.correlation_id(),
+                           message.function_id(),
+                           msg_version,
+                           remove_unknown_parameters)) {
+          LOG4CXX_WARN(
+              logger_,
+              "Failed to parse string to smart object with API version "
+                  << msg_version.toString() << " : " << message.json_message());
 
-        std::shared_ptr<smart_objects::SmartObject> response(
-            MessageHelper::CreateNegativeResponse(
-                message.connection_key(),
-                message.function_id(),
-                message.correlation_id(),
-                mobile_apis::Result::INVALID_DATA));
-
-        (*response)[strings::msg_params][strings::info] =
-            rpc::PrettyFormat(report);
-        app_manager_.GetRPCService().ManageMobileCommand(
-            response, commands::Command::SOURCE_SDL);
-        return false;
+          return false;
+        }
       }
+
       LOG4CXX_DEBUG(logger_,
                     "Convertion result for sdl object is true function_id "
                         << output[jhs::S_PARAMS][jhs::S_FUNCTION_ID].asInt());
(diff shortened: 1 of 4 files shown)


src/components/application_manager/src/rpc_passing_handler.cc
```diff
@@ -228,6 +233,21 @@ void RPCPassingHandler::ForwardRequestToCore(uint32_t correlation_id) {
       std::make_shared<smart_objects::SmartObject>(message);
   rpc_request_queue.erase(correlation_id);
   rpc_request_queue_lock_.Release();
+
+  // Revalidate rpc message before forwarding to core
```
jacobkeeler: Eh, wouldn't say `Revalidate`, it's technically being validated for the first time, just `Validate`.

src/components/application_manager/src/rpc_handler_impl.cc
```diff
@@ -283,7 +283,8 @@ void RPCHandlerImpl::GetMessageVersion(
 bool RPCHandlerImpl::ConvertMessageToSO(
     const Message& message,
     ns_smart_device_link::ns_smart_objects::SmartObject& output,
-    const bool remove_unknown_parameters) {
+    const bool remove_unknown_parameters,
+    const bool validate_existing_params) {
```
jacobkeeler: I think that `validate` would be descriptive enough alone

src/components/application_manager/src/rpc_handler_impl.cc
```diff
@@ -449,6 +445,37 @@ bool RPCHandlerImpl::ConvertMessageToSO(
...
+  if (!message ||
+      !mobile_so_factory().attachSchema(
+          *message, remove_unknown_params, msg_version) ||
+      ((message->validate(&report, msg_version) !=
+        smart_objects::errors::OK))) {
+    LOG4CXX_WARN(logger_,
+                 "Failed to parse string to smart object with API version "
+                     << msg_version.toString() << " : " << message->asString());
+
+    std::shared_ptr<smart_objects::SmartObject> response(
```
jacobkeeler: I think this might be better outside of this function, as it doesn't match the naming of the function. I will note that this would require the `report` object to be an added as an output parameter, though it would remove the need for other message-related parameters.

src/components/application_manager/src/rpc_passing_handler.cc
```diff
@@ -228,6 +233,20 @@ void RPCPassingHandler::ForwardRequestToCore(uint32_t correlation_id) {
...
+      message[strings::params][strings::connection_key].asUInt();
+  int32_t function_id = message[strings::params][strings::function_id].asInt();
+
+  auto app_ptr = app_manager_.application(connection_key);
+  utils::SemanticVersion msg_version(0, 0, 0);
+  if (app_ptr) {
+    msg_version = app_ptr->msg_version();
+  }
+
+  app_manager_.GetRPCHandler().ValidateRPCSO(
```
jacobkeeler: You need to check the result of this function.

src/components/application_manager/src/rpc_handler_impl.cc
```diff
@@ -449,6 +445,37 @@ bool RPCHandlerImpl::ConvertMessageToSO(
   return true;
 }
 
+bool RPCHandlerImpl::ValidateRPCSO(smart_objects::SmartObject* message,
```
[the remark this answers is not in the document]
ShobhitAd: Renamed to `ValidateRpcSO`

src/components/application_manager/src/rpc_passing_handler.cc
ShobhitAd: Changed to Validate

src/components/application_manager/src/rpc_handler_impl.cc
ShobhitAd: Eh, don't want to confuse it with the validate function(in terms of readability). Could rename it to `validate_params` or `validate_rpc`

src/components/application_manager/src/rpc_handler_impl.cc
```diff
@@ -316,30 +317,22 @@ bool RPCHandlerImpl::ConvertMessageToSO(
...
-                message.function_id(),
-                message.correlation_id(),
-                mobile_apis::Result::INVALID_DATA));
-
-        (*response)[strings::msg_params][strings::info] =
-            rpc::PrettyFormat(report);
-        app_manager_.GetRPCService().ManageMobileCommand(
-            response, commands::Command::SOURCE_SDL);
-        return false;
+      if (validate_existing_params) {
```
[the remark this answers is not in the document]
ShobhitAd: Made fix

src/components/application_manager/src/rpc_handler_impl.cc
```diff
@@ -388,16 +381,19 @@ bool RPCHandlerImpl::ConvertMessageToSO(
...
-        LOG4CXX_ERROR(logger_,
-                      "Incorrect parameter from HMI"
-                          << rpc::PrettyFormat(report));
-
-        output.erase(strings::msg_params);
-        output[strings::params][hmi_response::code] =
-            hmi_apis::Common_Result::INVALID_DATA;
-        output[strings::msg_params][strings::info] = rpc::PrettyFormat(report);
-        return false;
+      if (validate_existing_params) {
```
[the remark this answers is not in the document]
ShobhitAd: Made fix

src/components/application_manager/src/rpc_handler_impl.cc
```diff
@@ -449,6 +445,37 @@ bool RPCHandlerImpl::ConvertMessageToSO(
...
+                                   uint32_t connection_key,
+                                   uint32_t correlation_id,
+                                   int32_t function_id,
+                                   utils::SemanticVersion& msg_version,
+                                   bool remove_unknown_params) {
+  rpc::ValidationReport report("RPC");
+  if (!message ||
+      !mobile_so_factory().attachSchema(
+          *message, remove_unknown_params, msg_version) ||
+      ((message->validate(&report, msg_version) !=
```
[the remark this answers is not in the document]
ShobhitAd: Remove extra parentheses

src/components/application_manager/src/rpc_handler_impl.cc
ShobhitAd: This would make sense if we wanted to have a different response(or no response) every time we used this function. But since its a generic error related to validating the RPC smart object shouldn't we just let it remain in this function?

src/components/application_manager/src/rpc_passing_handler.cc
ShobhitAd: I guess I would simply return in the case the validation fails since the function already handles sending the `INVALID_DATA` response?

src/components/application_manager/src/rpc_handler_impl.cc
jacobkeeler: Ah, got you. I think `validate_params` would be fine

src/components/application_manager/src/rpc_handler_impl.cc
jacobkeeler: The way that the function is worded, it implies that it performs the validation and returns the result, not that it sends a response within the actual function. Unfortunately, I can't think of a better name for the function which would encompass this behavior, but if it was renamed properly I'd be fine with it. I could also see a case where you might just want to not to send a response back (as is done with HMI notifications at the moment).

ShobhitAd: Moved the response message outside of the function

src/components/application_manager/src/rpc_passing_handler.cc
ShobhitAd: Made the change ```if (!app_manager_.GetRPCHandler().ValidateRpcSO(
          &message, msg_version, report, true)) {
    std::shared_ptr<smart_objects::SmartObject> response(
        MessageHelper::CreateNegativeResponse(
            connection_key,
            function_id,
            correlation_id,
            mobile_apis::Result::INVALID_DATA));

    (*response)[strings::msg_params][strings::info] = rpc::PrettyFormat(report);
    app_manager_.GetRPCService().ManageMobileCommand(
        response, commands::Command::SOURCE_SDL);
    return;
  }```

PR #2854, opened 2019-03-19
Coverity 5.1 Fixes

JackLivio:

### Risk

### Testing Plan
- Submit build to coverity.
- Run unit tests.

### Summary
- Check return values
- Pass struct by reference
- Cast value to match return type

### CLA

### DIFF
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -2942,9 +2942,11 @@ void ApplicationManagerImpl::UnregisterApplication(
     auto it_app = applications_.begin();
     while (applications_.end() != it_app) {
       if (app_id == (*it_app)->app_id()) {
-        connection_handler().GetDeviceID((*it_app)->mac_address(), &handle);
-        app_to_remove = *it_app;
-        applications_.erase(it_app++);
+        if (connection_handler().GetDeviceID((*it_app)->mac_address(),
+                                             &handle)) {
+          app_to_remove = *it_app;
+          applications_.erase(it_app++);
+        }
       } else {
         ++it_app;
       }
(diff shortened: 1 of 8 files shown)


src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -2942,9 +2942,11 @@ void ApplicationManagerImpl::UnregisterApplication(
     auto it_app = applications_.begin();
     while (applications_.end() != it_app) {
       if (app_id == (*it_app)->app_id()) {
-        connection_handler().GetDeviceID((*it_app)->mac_address(), &handle);
-        app_to_remove = *it_app;
-        applications_.erase(it_app++);
+        if (connection_handler().GetDeviceID((*it_app)->mac_address(),
```
jacobkeeler: Are you sure that the app still shouldn't be removed if GetDeviceID fails? Also, shouldn't there be a `break` at the end of this statement? It seems we shouldn't reach this case multiple times.

On that note, couldn't the `GetDeviceID` be called outside of this loop, with `app_to_remove->mac_address()` as the argument?

src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -2970,15 +2968,20 @@ void ApplicationManagerImpl::UnregisterApplication(
...
-    ApplicationSharedPtr app = FindApp(applications(), finder);
-    if (!app) {
-      LOG4CXX_DEBUG(
-          logger_, "There is no more SDL4 apps with device handle: " << handle);
 
-      RemoveAppsWaitingForRegistration(handle);
-      RefreshCloudAppInformation();
-      SendUpdateAppList();
+    if (connection_handler().GetDeviceID(app_to_remove->mac_address(),
+                                         &handle)) {
```
jacobkeeler: I'm a bit worried that there might be a case where GetDeviceID would fail for a cloud application, in which case the cloud device would not be properly recreated with `RefreshCloudAppInformation` maybe have an `else if (app_to_remove->is_cloud_app())` to cover this?

PR #2880, opened 2019-04-08
Error Response for AppService.GetAppServiceData with unknown service type/service id

ShobhitAd:

### Risk

### Testing Plan
Create ATF test with HMI sending GetAppServiceData with unknown service type/service id

### Summary
At the moment, if the HMI sends a `AppService.GetAppServiceData` request with an unknown service type/service id core does not respond and the request times out. This fix adds an error response for this case

### CLA

### DIFF
--- a/src/components/application_manager/src/commands/request_from_hmi.cc
+++ b/src/components/application_manager/src/commands/request_from_hmi.cc
@@ -173,16 +173,22 @@ void RequestFromHMI::SendProviderRequest(
   LOG4CXX_AUTO_TRACE(logger_);
   bool hmi_destination = false;
   ApplicationSharedPtr app;
+  std::string error_msg = "No app service provider available";
+
   if ((*msg)[strings::msg_params].keyExists(strings::service_type)) {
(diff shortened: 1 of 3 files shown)


src/components/application_manager/src/commands/request_from_hmi.cc
```diff
@@ -196,6 +202,11 @@ void RequestFromHMI::SendProviderRequest(
 
   if (!app) {
     LOG4CXX_DEBUG(logger_, "Invalid App Provider pointer");
+    SendErrorResponse(correlation_id(),
+                      static_cast<hmi_apis::FunctionID::eType>(function_id()),
+                      hmi_apis::Common_Result::DATA_NOT_AVAILABLE,
```
jacobkeeler: Should send `INVALID_ID` if serviceId was provided.

jacobkeeler: Also, we should include this logic on the HMI version of this function as well.

PR #2900, opened 2019-04-25
Building with option BUILD_CLOUD_APP_SUPPORT=OFF

### LINKED ISSUE #2895: Build With Option `BUILD_CLOUD_APP_SUPPORT`=OFF Fails
### Bug Report
Build With Option `BUILD_CLOUD_APP_SUPPORT`=OFF Fails

##### Reproduction Steps
1. Set CMAKE Flag `BUILD_CLOUD_APP_SUPPORT`=OFF
2. Build Core
##### Expected Behavior
Core compiles with no errors.

##### Observed Behavior
Build fails on undefined references to dependencies because there are missing ifdefs to only compile certain cloud app classes when the build option is ON.

* OS/Version: Ubuntu 18
* SDL Core Version: 5.1

ShobhitAd:
Fixes #2895 

### Risk

### Testing Plan
Building with option `BUILD_CLOUD_APP_SUPPORT=OFF` and checking for errors

### Summary
Excluded cloud app transport files when `BUILD_CLOUD_APP_SUPPORT=OFF` and added default returns for cloud app functions in other files

### CLA

### DIFF
--- a/src/components/application_manager/src/application_impl.cc
+++ b/src/components/application_manager/src/application_impl.cc
@@ -1185,6 +1185,9 @@ const std::string& ApplicationImpl::cloud_app_certificate() const {
 }
 
 bool ApplicationImpl::is_cloud_app() const {
+#if !defined(CLOUD_APP_WEBSOCKET_TRANSPORT_SUPPORT)
+  return false;
+#endif  // CLOUD_APP_WEBSOCKET_TRANSPORT_SUPPORT
   return !endpoint_.empty();
 }
 
--- a/src/components/application_manager/src/application_impl.cc
+++ b/src/components/application_manager/src/application_impl.cc
@@ -1187,8 +1187,9 @@ const std::string& ApplicationImpl::cloud_app_certificate() const {
 bool ApplicationImpl::is_cloud_app() const {
 #if !defined(CLOUD_APP_WEBSOCKET_TRANSPORT_SUPPORT)
   return false;
-#endif  // CLOUD_APP_WEBSOCKET_TRANSPORT_SUPPORT
+#else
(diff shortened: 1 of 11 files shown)


src/components/application_manager/src/application_impl.cc
```diff
@@ -1185,6 +1185,9 @@ const std::string& ApplicationImpl::cloud_app_certificate() const {
 }
 
 bool ApplicationImpl::is_cloud_app() const {
+#if !defined(CLOUD_APP_WEBSOCKET_TRANSPORT_SUPPORT)
+  return false;
+#endif  // CLOUD_APP_WEBSOCKET_TRANSPORT_SUPPORT
```
jacobkeeler: I feel that `else` macros should be included, otherwise we will probably get "dead code" warnings from Coverity scans

jacobkeeler:
Should probably include guards in the application_manager as well (functions such as `RefreshCloudAppInformation`).

src/components/application_manager/src/application_impl.cc
LuxoftAKutsan: @jacobkeeler should we have is_cloud_app app function if CLOUD_APP_WEBSOCKET_TRANSPORT_SUPPORT is turned off? I would say that definition of  this function can wrapped in ifdef

jacobkeeler: I'd be inclined to agree with you, but this would require quite a significant amount of `ifdef`s to be added to the project (somewhere around 20). Plus it seems to me that this would be a major API change, since this function would no longer be available when this feature is disabled. I think that just having this function return `false` when cloud apps aren't enabled is fine.

PR #2912, opened 2019-05-10
RPC Passing Handler unit tests

ShobhitAd:

### Risk

### CLA

### DIFF
--- a/src/components/application_manager/test/rpc_passing_handler_test.cc
+++ b/src/components/application_manager/test/rpc_passing_handler_test.cc
@@ -1,0 +1,348 @@
+/*
+ * Copyright (c) 2019, Ford Motor Company
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ * Redistributions of source code must retain the above copyright notice, this
+ * list of conditions and the following disclaimer.
+ *
+ * Redistributions in binary form must reproduce the above copyright notice,
+ * this list of conditions and the following
+ * disclaimer in the documentation and/or other materials provided with the
+ * distribution.
+ *
+ * Neither the name of the Ford Motor Company nor the names of its contributors
+ * may be used to endorse or promote products derived from this software
+ * without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <gmock/gmock.h>
+#include "gtest/gtest.h"
+
+#include "application_manager/rpc_passing_handler.h"
+
+// #include "application_manager/commands/command.h"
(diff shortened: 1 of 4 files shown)


src/components/application_manager/test/rpc_passing_handler_test.cc
```diff
@@ -0,0 +1,447 @@
...
+    EXPECT_CALL(mock_app_manager_, GetRPCHandler());
+    EXPECT_CALL(mock_rpc_handler_,
+                ValidateRpcSO(request_params.message, _, _, false))
+        .WillOnce(Return(true));
+    EXPECT_CALL(mock_app_manager_, GetRPCService());
+    EXPECT_CALL(mock_rpc_service_,
+                ManageMobileCommand(Pointee(request_params.message),
+                                    am::commands::Command::SOURCE_MOBILE));
+  }
+  sleep(1);
```
jacobkeeler: Look at `TestASyncWaiter` for asynchronous expectations

src/components/application_manager/test/rpc_passing_handler_test.cc
```diff
@@ -0,0 +1,447 @@
...
+  }
+
+  void SendRequestToASP(const PassthroughParams& params, uint32_t timeout = 0) {
+    if (timeout == 0) {
+      // If timeout argument is not sepcified
+      timeout = kRPCPassthroughTimeout;
+    }
+
+    {
+      InSequence dummy;
```
jacobkeeler: Is it possible to avoid using `InSequence`? I would assume this means that the functions need to be called in order, which I imagine would lead to fragile tests.

src/components/application_manager/test/rpc_passing_handler_test.cc
```diff
@@ -0,0 +1,447 @@
...
+      // Call RPCPassThrough with request smart object
+
+      // Will call PopulateRPCRequestQueue
+      EXPECT_CALL(mock_app_service_manager_, GetActiveServices())
+          .WillOnce(Return(app_services_));
+      EXPECT_CALL(mock_app_manager_,
+                  IncreaseForwardedRequestTimeout(params.connection_key,
+                                                  params.correlation_id))
+          .Times(app_services_.size());
```
jacobkeeler: Hmmm, I suppose I didn't realize that this was how this worked before. This is incorrect, this function should be called each time that the RPC is passed to another service, not several times when preparing passthrough. I believe that this means that the timeout doesn't actually work now, because it doesn't just add to the timeout each time this is called, it resets the timeout to `default_timeout + rpc_passthrough_timeout`, meaning that this is the total time that the app has before a GENERIC_ERROR response.

jacobkeeler: Actually never mind, this is called before this message is added to the `RequestController`, so it won't time out because of this. That said, this call will have absolutely no effect, so it should be removed.

src/components/application_manager/test/rpc_passing_handler_test.cc
ShobhitAd: Resolved in https://github.com/smartdevicelink/sdl_core/pull/2912/commits/d194dfd9e425d0539a1c4a487eaf5dd389cd623b

src/components/application_manager/test/rpc_passing_handler_test.cc
ShobhitAd: Resolved in https://github.com/smartdevicelink/sdl_core/pull/2912/commits/d194dfd9e425d0539a1c4a487eaf5dd389cd623b

src/components/application_manager/test/rpc_passing_handler_test.cc
ShobhitAd: Since this fix will require changes to the `rpc_passing_handler.cc` and the test file it should probably be in its own PR.

jacobkeeler: Should still remove this expectation, since it doesn't have a return value anyway.

src/components/application_manager/test/rpc_passing_handler_test.cc
```diff
@@ -0,0 +1,429 @@
...
+  // Will cycle to core (no other app services in list)
+  EXPECT_CALL(mock_app_manager_, GetRPCHandler());
+  EXPECT_CALL(mock_rpc_handler_,
+              ValidateRpcSO(request_params.message, _, _, false))
+      .WillOnce(Return(true));
+  EXPECT_CALL(mock_app_manager_, GetRPCService());
+  EXPECT_CALL(mock_rpc_service_,
+              ManageMobileCommand(Pointee(request_params.message),
+                                  am::commands::Command::SOURCE_MOBILE))
+      .WillOnce(DoAll(NotifyTestAsyncWaiter(waiter), Return(true)));
```
jacobkeeler: Technically, these `EXPECT_CALL`s should be above `SendRequestToASL`

ShobhitAd: I guess but these calls aren't directly triggered during the execution of SendRequestToASP(or the RPCPassthrough function) but rather after the timeout which could be considered a separate event triggered by the WaitFor function

ShobhitAd: Resolved in https://github.com/smartdevicelink/sdl_core/pull/2912/commits/7d556a33dc13e3411e22c3e07eaad330c45a1e8b

src/components/application_manager/test/rpc_passing_handler_test.cc
ShobhitAd: Resolved in https://github.com/smartdevicelink/sdl_core/pull/2912/commits/7d556a33dc13e3411e22c3e07eaad330c45a1e8b

PR #2948, opened 2019-06-14
Add CloseApplication RPC

jacobkeeler:
Implements #1931 

### Risk

### Testing Plan
Create base ATF tests for new RPC

### Summary
Add new `CloseApplication` RPC to Mobile and HMI interfaces.

##### Enhancements
* Add new `CloseApplication` RPC to Mobile and HMI interfaces.

### CLA

### DIFF
--- a/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/include/sdl_rpc_plugin/commands/hmi/basic_communication_close_application_request.h
+++ b/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/include/sdl_rpc_plugin/commands/hmi/basic_communication_close_application_request.h
@@ -1,0 +1,80 @@
+/*
+ Copyright (c) 2019, Ford Motor Company, Livio
+ All rights reserved.
+
+ Redistribution and use in source and binary forms, with or without
+ modification, are permitted provided that the following conditions are met:
+
+ Redistributions of source code must retain the above copyright notice, this
+ list of conditions and the following disclaimer.
+
+ Redistributions in binary form must reproduce the above copyright notice,
+ this list of conditions and the following
+ disclaimer in the documentation and/or other materials provided with the
+ distribution.
+
+ Neither the name of the the copyright holders nor the names of their
+ contributors may be used to endorse or promote products derived from this
+ software without specific prior written permission.
+
+ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef SRC_COMPONENTS_APPLICATION_MANAGER_RPC_PLUGINS_SDL_RPC_PLUGIN_INCLUDE_SDL_RPC_PLUGIN_COMMANDS_HMI_BASIC_COMMUNICATION_CLOSE_APPLICATION_REQUEST_H_
+#define SRC_COMPONENTS_APPLICATION_MANAGER_RPC_PLUGINS_SDL_RPC_PLUGIN_INCLUDE_SDL_RPC_PLUGIN_COMMANDS_HMI_BASIC_COMMUNICATION_CLOSE_APPLICATION_REQUEST_H_
+
+#include "application_manager/commands/request_to_hmi.h"
+#include "sdl_rpc_plugin/sdl_rpc_plugin.h"
+
+namespace sdl_rpc_plugin {
+namespace app_mngr = application_manager;
+
+namespace commands {
+
+/**
+ * @brief BasicCommunicationCloseApplicationRequest command class
+ **/
+class BasicCommunicationCloseApplicationRequest
+    : public app_mngr::commands::RequestToHMI {
+ public:
+  /**
+   * @brief BasicCommunicationCloseApplicationRequest class constructor
+   *
+   * @param message Incoming SmartObject message
+   **/
+  BasicCommunicationCloseApplicationRequest(
+      const app_mngr::commands::MessageSharedPtr& message,
+      app_mngr::ApplicationManager& application_manager,
+      app_mngr::rpc_service::RPCService& rpc_service,
+      app_mngr::HMICapabilities& hmi_capabilities,
+      policy::PolicyHandlerInterface& policy_handle);
+
+  /**
+   * @brief BasicCommunicationCloseApplicationRequest class destructor
+   **/
+  virtual ~BasicCommunicationCloseApplicationRequest();
+
+  /**
+   * @brief Execute command
+   **/
+  virtual void Run();
+
+ private:
+  DISALLOW_COPY_AND_ASSIGN(BasicCommunicationCloseApplicationRequest);
+};
+
+}  // namespace commands
+
+}  // namespace sdl_rpc_plugin
+
+#endif  // SRC_COMPONENTS_APPLICATION_MANAGER_RPC_PLUGINS_SDL_RPC_PLUGIN_INCLUDE_SDL_RPC_PLUGIN_COMMANDS_HMI_BASIC_COMMUNICATION_CLOSE_APPLICATION_REQUEST_H_
--- a/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/include/sdl_rpc_plugin/commands/hmi/basic_communication_close_application_request.h
+++ b/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/include/sdl_rpc_plugin/commands/hmi/basic_communication_close_application_request.h
@@ -62,12 +62,12 @@ class BasicCommunicationCloseApplicationRequest
   /**
    * @brief BasicCommunicationCloseApplicationRequest class destructor
    **/
-  virtual ~BasicCommunicationCloseApplicationRequest();
+  ~BasicCommunicationCloseApplicationRequest() OVERRIDE;
 
   /**
    * @brief Execute command
    **/
-  virtual void Run();
+  void Run() OVERRIDE;
 
  private:
   DISALLOW_COPY_AND_ASSIGN(BasicCommunicationCloseApplicationRequest);
--- a/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/close_application_request.cc
+++ b/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/close_application_request.cc
@@ -1,0 +1,118 @@
+/*
+ Copyright (c) 2019, Ford Motor Company, Livio
+ All rights reserved.
+
+ Redistribution and use in source and binary forms, with or without
+ modification, are permitted provided that the following conditions are met:
+
+ Redistributions of source code must retain the above copyright notice, this
+ list of conditions and the following disclaimer.
+
+ Redistributions in binary form must reproduce the above copyright notice,
+ this list of conditions and the following
+ disclaimer in the documentation and/or other materials provided with the
+ distribution.
+
+ Neither the name of the the copyright holders nor the names of their
+ contributors may be used to endorse or promote products derived from this
+ software without specific prior written permission.
+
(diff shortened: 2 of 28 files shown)


src/components/application_manager/rpc_plugins/sdl_rpc_plugin/include/sdl_rpc_plugin/commands/hmi/basic_communication_close_application_request.h
```diff
@@ -0,0 +1,80 @@
...
+
+  /**
+   * @brief BasicCommunicationCloseApplicationRequest class destructor
+   **/
+  virtual ~BasicCommunicationCloseApplicationRequest();
+
+  /**
+   * @brief Execute command
+   **/
+  virtual void Run();
```
LuxoftAKutsan: @jacobkeeler I would use OVERRIDE macro and avoid virtual keyword

LuxoftAKutsan:
@jacobkeeler This PR contains additional Request\Resoponse to HMI.
There was clarification with proposal author, that for Mobile Close RPC can be used existing ActivateApp to None .
https://github.com/smartdevicelink/sdl_requirements/issues/111 
@kshala-ford  could you please provide any information about it?

kshala-ford:
@LuxoftAKutsan does this PR follow option 2 of smartdevicelink/sdl_requirements#111 ?

jacobkeeler:
@kshala-ford It would appear so.
@LuxoftAKutsan My only reservation with using `BC.ActivateApp` for this purpose is that it seems like an improper use of the RPC. The app is not being activated, it is being deactivated. I would think that the `level` parameter of `BC.ActivateApp` is really only meant to be filled with LIMITED or BACKGROUND in specific cases, NONE doesn't make much sense in the context of this message since the description states that the app should be brought to the foreground. It would technically work as it is implemented in the `sdl_hmi` currently, but the `generic_hmi` doesn't have this sort of logic because it isn't really described in the [HMI integration guidelines](https://smartdevicelink.com/en/docs/hmi/master/basiccommunication/activateapp/).

LuxoftAKutsan:
@kshala-ford  this PR follow option 2 from https://github.com/smartdevicelink/sdl_requirements/issues/111. 

@jacobkeeler From my perspective Adding `CloseApplication` RPC to HMI API will add confusion. 
We will have  :
 - BC.ActiveteApp (with possible HMI level NONE and background, and as I remember SDL use int in some cases) 
 - CloseApplicaiton

Now we have 2 options : 

- Add explicit description that `NONE`, `BACKGROUND` should not be used in ActivateApp and add new `CloseApplication` RPC. Look throw existing sequenses and replace sending `ActivateApp[NONE]`  to `CloseApplication` (I suppose that one of the use cases is revoke applicaiton). 

- Avoid adding additional RPC `CloseApplication`, and use  `ActivateApp[NONE]`  instead of it. Add apropriate implementation to generic_hmi.

Also it is not clear for me difference between sdl_hmi and generic_hmi. Why do we need both in opensource?

jacobkeeler:
@LuxoftAKutsan I believe that the documentation would need to be updated in either case, because the HMI documentation doesn't describe the HMI's behavior for each HMI level. As stated, it doesn't seem like `ActivateApp(NONE)` is a valid case as it is currently described:

```
<description>Request from SDL to HMI to bring specified application to front on UI e.g make it HMI status 'FULL'.</description>
```

In addition, it seems to me that `HMILevel` is generally used as a Mobile<->Core concept rather than a Core<->HMI concept. `BC.ActivateApp` is the only message which includes this enum in the entire HMI API, so it appears to me that it might have been added after the RPC was initially implemented to fix a specific issue.

The `generic_hmi` is generally preferred as a reference implementation for SDL, but it is missing several components at the moment (such as Navigation). Being a reference implementation, it is not meant to imitate a full infotainment system, so components such as VehicleInfo will likely never be a part of the `generic_hmi`. The `sdl_hmi` is still available specifically as an emulator for a full infotainment system, including the components which cannot be included in the scope of the `generic_hmi`.

src/components/application_manager/src/state_controller_impl.cc
```diff
@@ -159,7 +164,7 @@ void StateControllerImpl::SetRegularState(
     const mobile_apis::AudioStreamingState::eType audio_state,
     const mobile_apis::VideoStreamingState::eType video_state,
     const mobile_apis::SystemContext::eType system_context,
-    const bool send_activate_app) {
+    const bool send_state_to_hmi) {
```
LuxoftAKutsan: @jacobkeeler confusing naming send_state_to_hmi. 
This boolean parameter controls if request to HMI should be used for put application in the new state. 
`send_state_to_hmi` sounds like you will notify HMI about changing state. I would suggest to rename it to "request_hmi"

src/components/application_manager/src/state_controller_impl.cc
```diff
@@ -791,31 +797,33 @@ void StateControllerImpl::OnApplicationRegistered(
   OnStateChanged(app, initial_state, new_state);
 }
 
-int64_t StateControllerImpl::SendBCActivateApp(
+int64_t StateControllerImpl::SendStateToHMI(
```
LuxoftAKutsan: @jacobkeeler I would rename it to something like RequestStateChangeToHMI

src/components/application_manager/src/state_controller_impl.cc
```diff
@@ -791,31 +797,33 @@ void StateControllerImpl::OnApplicationRegistered(
...
-          app,
-          app_mngr_.connection_handler().get_session_observer(),
-          app_mngr_.GetPolicyHandler(),
-          level,
-          send_policy_priority,
-          app_mngr_);
-  if (!bc_activate_app_request) {
-    LOG4CXX_ERROR(logger_, "Unable to create BC.ActivateAppRequest");
+  smart_objects::SmartObjectSPtr request = NULL;
+  if (level == hmi_apis::Common_HMILevel::NONE) {
```
LuxoftAKutsan: @jacobkeeler good practice to use the variable to the right of comparison

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/close_application_request.cc
```diff
@@ -0,0 +1,118 @@
...
+  auto msg_params = message[strings::msg_params];
+
+  hmi_apis::Common_Result::eType hmi_result =
+      static_cast<hmi_apis::Common_Result::eType>(
+          message[strings::params][hmi_response::code].asInt());
+
+  mobile_apis::Result::eType result =
+      MessageHelper::HMIToMobileResult(hmi_result);
+  bool success = PrepareResultForMobileResponse(
+      hmi_result, HmiInterfaces::HMI_INTERFACE_AppService);
```
JackLivio: > HMI_INTERFACE_AppService

Shouldn't this be BasicCommunication?

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/close_application_request.cc
```diff
@@ -0,0 +1,118 @@
...
+  if (success) {
```
JackLivio: Should part of the success be to check if the HMI also sent an OnAppDeactivated? Before setting the app state to HMI_LEVEL none, should the state be checked to see if it is in limited or background?

src/components/application_manager/src/message_helper/message_helper.cc
```diff
@@ -1789,12 +1789,16 @@ void MessageHelper::SendOnAppUnregNotificationToHMI(
 
 smart_objects::SmartObjectSPtr MessageHelper::GetBCActivateAppRequestToHMI(
     ApplicationConstSharedPtr app,
-    const protocol_handler::SessionObserver& session_observer,
     const policy::PolicyHandlerInterface& policy_handler,
     hmi_apis::Common_HMILevel::eType level,
     bool send_policy_priority,
     ApplicationManager& app_mngr) {
   DCHECK_OR_RETURN(app, smart_objects::SmartObjectSPtr());
+  if (hmi_apis::Common_HMILevel::NONE == level) {
```
JackLivio: Isnt the app in NONE before it is activated? Can you please explain this if condition and early return?

src/components/application_manager/src/state_controller_impl.cc
```diff
@@ -791,31 +797,33 @@ void StateControllerImpl::OnApplicationRegistered(
...
-    LOG4CXX_ERROR(logger_, "Unable to send BC.ActivateAppRequest");
+  if (!app_mngr_.GetRPCService().ManageHMICommand(request)) {
+    LOG4CXX_ERROR(logger_, "Unable to send request");
     return -1;
   }
-  const int64_t corr_id =
-      (*bc_activate_app_request)[strings::params][strings::correlation_id]
-          .asInt();
-  return corr_id;
+  const uint32_t corr_id =
```
JackLivio: Why the change to uint32 and then recast to int64?

src/components/interfaces/HMI_API.xml
```diff
@@ -3981,6 +3981,14 @@
     </function>
     <function name="ActivateApp" messagetype="response">
     </function>
+    <function name="CloseApplication" messagetype="request">
+      <description>Request from the application to exit the foreground.</description>
+      <param name="appID" type="Integer" mandatory="true">
```
JackLivio: Is this the hmi app id or the policy app id?

src/components/interfaces/MOBILE_API.xml
```diff
@@ -2645,6 +2645,7 @@
         <element name="GetAppServiceDataID" value="53" hexvalue="35" since="5.1" />
         <element name="GetFileID" value="54" hexvalue="36" since="5.1" />
         <element name="PerformAppServiceInteractionID" value="55" hexvalue="37" since="5.1" />
+        <element name="CloseApplicationID" value="56" hexvalue="38" since="5.2" />
```
JackLivio: This function id is going to conflict with other functions ids. [](uhttps://github.com/smartdevicelink/rpc_spec/pull/171/filesrl)

src/components/interfaces/HMI_API.xml
jacobkeeler: HMI app ID. It's an Integer, not a String

src/components/application_manager/src/state_controller_impl.cc
jacobkeeler: It seems this was originally chosen for the return value of `SendBCActivateApp` to allow for an error response of `-1`, correlation IDs use uint32_t. I could remove these casts, but it was being done implicitly before anyway.

src/components/application_manager/src/message_helper/message_helper.cc
jacobkeeler: This is the requested level, not the current level. This condition is to prevent `BC.ActivateApp(NONE)`, since this would be used to deactivate the application instead. `BC.CloseApplication` is used instead.

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/close_application_request.cc
jacobkeeler: I think OnExitApplication would be more appropriate, but I don't think it's necessary for the HMI to send this message, the response should be enough.

src/components/interfaces/MOBILE_API.xml
JackLivio:
```suggestion
        <element name="CloseApplicationID" value="58" hexvalue="3A" since="6.0" />
```

src/components/interfaces/MOBILE_API.xml
```diff
@@ -7311,6 +7312,28 @@
         </param>
     </function>
 
+    <function name="CloseApplication" functionID="CloseApplicationID" messagetype="request" since="5.2">
```
JackLivio:
```suggestion
    <function name="CloseApplication" functionID="CloseApplicationID" messagetype="request" since="6.0">
```

src/components/policy/policy_external/include/policy/policy_table/enums.h
```diff
@@ -471,6 +471,11 @@ enum FunctionID {
    */
   PerformAppServiceInteractionID = 55,
 
+  /**
+   * @brief CloseApplicationID.
+   */
+  CloseApplicationID = 56,
```
JackLivio:
```suggestion
  CloseApplicationID = 58,
```

src/components/policy/policy_regular/include/policy/policy_table/enums.h
```diff
@@ -456,6 +456,11 @@ enum FunctionID {
    */
   PerformAppServiceInteractionID = 55,
 
+  /**
+   * @brief CloseApplicationID.
+   */
+  CloseApplicationID = 56,
```
JackLivio:
```suggestion
  CloseApplicationID = 58,
```

src/components/interfaces/MOBILE_API.xml
```diff
@@ -7311,6 +7312,28 @@
...
+        <description>Request from the application to exit the foreground and enter HMI_NONE.</description>
+    </function>
+
+    <function name="CloseApplication" functionID="CloseApplicationID" messagetype="response" since="5.2">
```
jacobkeeler:
```suggestion
    <function name="CloseApplication" functionID="CloseApplicationID" messagetype="response" since="6.0">
```

src/components/application_manager/src/state_controller_impl.cc
jacobkeeler: @LuxoftAKutsan fixed with https://github.com/smartdevicelink/sdl_core/pull/2948/commits/8080f10bd7b7f5bbe074cbdafdc19fba7ebe8118

src/components/application_manager/src/state_controller_impl.cc
jacobkeeler: @LuxoftAKutsan fixed with https://github.com/smartdevicelink/sdl_core/pull/2948/commits/8080f10bd7b7f5bbe074cbdafdc19fba7ebe8118

src/components/application_manager/src/state_controller_impl.cc
jacobkeeler: @LuxoftAKutsan fixed with https://github.com/smartdevicelink/sdl_core/pull/2948/commits/8080f10bd7b7f5bbe074cbdafdc19fba7ebe8118

PR #2927, opened 2019-05-23
Add Parameter Media Service Data mediaImage

JackLivio:
Fixes #2897 

This PR is **ready** for review.

### Risk
This PR makes **[no / minor / major]** API changes.

### Testing Plan
Publish media app service and send mediaImage parameter in OnAppServiceData.

### Summary
Changes are small and only required the parameters to be added to the mobile and hmi api. HMI app consumers will now be able to receive the mediaImage parameter.

### CLA
- [x] I have signed [the CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform)

PR #2953, opened 2019-06-19
Update MOBILE API to version 6.0.0

jacobkeeler:
Revision to #2927

### Risk
This PR makes **major** API changes.

### Summary
Update MOBILE API to version 6.0.0

##### Enhancements
* Update MOBILE API to version 6.0.0

### CLA

### DIFF
--- a/src/components/interfaces/MOBILE_API.xml
+++ b/src/components/interfaces/MOBILE_API.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" standalone="no"?>
 <?xml-stylesheet type="text/xml" href="protocol2html.xsl"?>
 
-<interface name="SmartDeviceLink RAPI" version="5.2.0" minVersion="1.0" date="2019-03-19">
+<interface name="SmartDeviceLink RAPI" version="6.0.0" minVersion="1.0" date="2019-03-19">
     <enum name="Result" internal_scope="base" since="1.0">
         <element name="SUCCESS">
             <description>The request succeeded</description>
@@ -3830,7 +3830,7 @@
                 Audiobook: The total number of sections in the playback queue (e.g. the number of chapters in the book)
             </description>
         </param>
-        <param name="mediaImage" type="Image" mandatory="false" since="5.2">
+        <param name="mediaImage" type="Image" mandatory="false" since="6.0">
           <description>
               Music: The album art of the current track
               Podcast: The podcast or chapter artwork of the current podcast episode


LuxoftAKutsan:
@JackLivio @jacobkeeler 
After merging this PR

All new and modified parameters in the following proposal will have `since` value `6.0` :

 - Adding GPS Shift support
 - Pixel density and Scale
 - Support running the same app from multiple devices at the same time
 - Remote Control - Radio and Climate Parameter Update
 - Service Status Update to HMI
 - RPC message protection
 - Power to the Box
 - Widget support
 - Allow Multiple Modules per Module Type
 - SDL-passenger-mode

I have a question related to backward compatibility.

Because of major version change, we can avoid deprecating old versions of modified parameters, and remove them.

It is a significant point mostly for the following proposals :
 - Support running the same app from multiple devices at the same time
 - Remote Control - Radio and Climate Parameter Update
 - Widget support

This proposal contains a modification of existing params.
Including proposal in minor release would require keeping backward compatibility and marking old parameter versions as deprecated.
Including proposal in major version allow us to simplify implementation and remove deprecated parameters.

AByzhynar:
@jacobkeeler Don't we need to change related parts in sdl_core c++ code like: https://github.com/smartdevicelink/sdl_core/blob/develop/src/components/utils/src/semantic_version.cc

JackLivio:
@AByzhynar No those do not need to be changed. They are used in defining specific behavior for core/mobile/hmi communication depending on the negotiated rpc verision. There might need to be a rpc_version_6 introduced if we want to be able to support backwards compatibility for pre 6.0 mobile apps while creating breaking changes for future versions.

JackLivio:
@LuxoftAKutsan As far as I know we want to keep them deprecated in 6.0 so those params have a period of deprecation and align with the accepted proposals. If you strongly believe they should be removed with no deprecation, there would need to be revisions made to those proposals.

LuxoftAKutsan:
@JackLivio Do we have any clear `check list` of reasons to remove a parameter, but not deprecate it? 
From my perspective : 
 - The parameter is not widely used by existing mobile apps.
 - All signficant mobile apps that used this parameter confirmed that mobile application will be updated to support new version
 - Overhead for supporting old parameter is to big ( BTW this point is not realy specific)

LuxoftAKutsan:
@JackLivio @jacobkeeler  And I have additional question regarding deprecated methods in C++ interfaces. Can the be deleted with an increasing major release version?

JackLivio:
@LuxoftAKutsan We have a different stance on Core changes. We are ok with just removing public methods in the Core code without deprecating in this major version change release.

PR #2979, opened 2019-07-31
Fix/remove vd from db if empty in pt

LuxoftAKutsan:
Fix of the floating bug: 

Sometimes after an ignition cycle policy table was not updated with deleted vehicle data in PTU


atiwari9:
Approved

PR #2980, opened 2019-07-31
Feature/widget support

mked-luxoft:
Implements #2918 

### Risk
This PR makes **[minor]** API changes.

### Testing Plan
* Unit tests were updated/extended
* ATF scripts 

### Summary
- Updated interfaces
- Added new RPCs: CreateWindow/DeleteWindow
- Updated OnSystemCapabiltitiesUpdated notification. 
- Updated existing RPCs with WindowID processing 
- Updated Show RPC with new parameter `templateConfiguration`
- Updated OnHMIStatus notification
- Updated State Controller as every app window can have its own state
- Updated/added all related unit tests

### Other parts of delivery:
 - [ATF test scripts](https://github.com/smartdevicelink/sdl_atf_test_scripts/pull/2218)
 - [SDL HMI](https://github.com/smartdevicelink/sdl_hmi/pull/213)
 - [SDL HMI Integration Guidelines](https://github.com/smartdevicelink/sdl_hmi_integration_guidelines/pull/169)

### CLA

### DIFF
--- a/src/components/application_manager/include/application_manager/application.h
+++ b/src/components/application_manager/include/application_manager/application.h
@@ -564,6 +564,20 @@ class Application : public virtual InitialApplicationData,
    */
   virtual void set_is_application_data_changed(bool state_application_data) = 0;
 
+  /**
+   * @brief Checks if application data is allowed to be resumed
+   * @return TRUE if data of application is allowed to be resumed, otherwise
+   * return FALSE
+   */
+  virtual bool is_app_data_resumption_allowed() const = 0;
+
+  /**
+   * @brief Sets resumption allowance for application data
+   * @param allowed - true if app data allowed to be resumed,
+   * false value means that app data is disallowed for resumption
+   */
+  virtual void set_app_data_resumption_allowance(const bool allowed) = 0;
+
   virtual void CloseActiveMessage() = 0;
   virtual bool IsFullscreen() const = 0;
   virtual void ChangeSupportingAppHMIType() = 0;
--- a/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/hmi/on_bc_system_capability_updated_notification_from_hmi.cc
+++ b/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/hmi/on_bc_system_capability_updated_notification_from_hmi.cc
@@ -87,7 +87,7 @@ OnBCSystemCapabilityUpdatedNotificationFromHMI::
   (*message_)[strings::params][strings::connection_key] =
       (*message_)[strings::msg_params][strings::app_id];
   (*message_)[strings::msg_params].erase(strings::app_id);
-  if (app->is_resuming()) {
+  if (app->is_resuming() && app->is_app_data_resumption_allowed()) {
     LOG4CXX_DEBUG(logger_, "Application is resuming");
     app->display_capabilities_builder().UpdateDisplayCapabilities(
         display_capabilities);
--- a/src/components/application_manager/src/application_impl.cc
+++ b/src/components/application_manager/src/application_impl.cc
@@ -113,6 +113,7 @@ ApplicationImpl::ApplicationImpl(
     , video_streaming_suspended_(true)
     , audio_streaming_suspended_(true)
     , is_app_allowed_(true)
+    , is_app_data_resumption_allowed_(false)
     , has_been_activated_(false)
     , tts_properties_in_none_(false)
     , tts_properties_in_full_(false)
@@ -718,6 +719,14 @@ bool ApplicationImpl::set_app_icon_path(const std::string& path) {
 
 void ApplicationImpl::set_app_allowed(const bool allowed) {
   is_app_allowed_ = allowed;
+}
+
+bool ApplicationImpl::is_app_data_resumption_allowed() const {
+  return is_app_data_resumption_allowed_;
+}
+
+void ApplicationImpl::set_app_data_resumption_allowance(const bool allowed) {
+  is_app_data_resumption_allowed_ = allowed;
 }
 
 void ApplicationImpl::set_secondary_device(
--- a/src/components/application_manager/src/resumption/resume_ctrl_impl.cc
+++ b/src/components/application_manager/src/resumption/resume_ctrl_impl.cc
@@ -217,7 +217,8 @@ bool ResumeCtrlImpl::RestoreAppHMIState(ApplicationSharedPtr application) {
 
       const bool app_hmi_state_is_set =
           SetAppHMIState(application, saved_hmi_level, true);
-      if (app_hmi_state_is_set) {
+      if (app_hmi_state_is_set &&
+          application->is_app_data_resumption_allowed()) {
         RestoreAppWidgets(application, saved_app);
       }
     } else {
@@ -681,9 +682,14 @@ bool ResumeCtrlImpl::CheckApplicationHash(ApplicationSharedPtr application,
                 "app_id : " << application->app_id() << " hash : " << hash);
   smart_objects::SmartObject saved_app;
   const std::string& device_mac = application->mac_address();
-  bool result = resumption_storage_->GetSavedApplication(
+  const bool get_app_result = resumption_storage_->GetSavedApplication(
       application->policy_app_id(), device_mac, saved_app);
-  return result ? saved_app[strings::hash_id].asString() == hash : false;
+  const bool check_result =
+      get_app_result ? saved_app[strings::hash_id].asString() == hash : false;
+  if (check_result) {
+    application->set_app_data_resumption_allowance(true);
+  }
+  return check_result;
 }
 
 void ResumeCtrlImpl::SaveDataOnTimer() {
--- a/src/components/application_manager/include/application_manager/application_impl.h
+++ b/src/components/application_manager/include/application_manager/application_impl.h
@@ -283,6 +283,9 @@ class ApplicationImpl : public virtual Application,
 
   virtual bool is_application_data_changed() const;
 
+  bool is_app_data_resumption_allowed() const OVERRIDE;
+  void set_app_data_resumption_allowance(const bool allowed) OVERRIDE;
+
   virtual void set_is_application_data_changed(bool state_application_data);
 
   /**
@@ -568,6 +571,7 @@ class ApplicationImpl : public virtual Application,
   sync_primitives::Lock streaming_stop_lock_;
 
   bool is_app_allowed_;
+  bool is_app_data_resumption_allowed_;
   bool has_been_activated_;
   bool tts_properties_in_none_;
   bool tts_properties_in_full_;
--- a/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/on_system_capability_updated_notification.cc
+++ b/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/on_system_capability_updated_notification.cc
@@ -208,7 +208,7 @@ void OnSystemCapabilityUpdatedNotification::Run() {
     if (mobile_apis::SystemCapabilityType::DISPLAYS == system_capability_type) {
       LOG4CXX_DEBUG(logger_, "Using common display capabilities");
       auto capabilities = hmi_capabilities_.system_display_capabilities();
-      if (app->is_resuming()) {
+      if (app->is_resuming() && app->is_app_data_resumption_allowed()) {
         LOG4CXX_DEBUG(logger_,
                       "Application "
                           << app->app_id()
--- a/src/components/application_manager/test/include/application_manager/mock_application.h
+++ b/src/components/application_manager/test/include/application_manager/mock_application.h
@@ -58,6 +58,8 @@ class MockApplication : public ::application_manager::Application {
   MOCK_CONST_METHOD0(is_application_data_changed, bool());
   MOCK_METHOD1(set_is_application_data_changed,
                void(bool state_application_data));
+  MOCK_CONST_METHOD0(is_app_data_resumption_allowed, bool());
+  MOCK_METHOD1(set_app_data_resumption_allowance, void(bool allowed));
   MOCK_METHOD0(CloseActiveMessage, void());
   MOCK_CONST_METHOD0(IsFullscreen, bool());
   MOCK_METHOD0(ChangeSupportingAppHMIType, void());


mked-luxoft:
@kshala-ford the description was fixed. Please approve this PR or feel free to suggest changes if necessary. The PR to rpc_spec can be found here https://github.com/smartdevicelink/rpc_spec/pull/185.

src/components/application_manager/include/application_manager/application.h
```diff
@@ -188,9 +196,16 @@ class DynamicApplicationData {
   virtual const smart_objects::SmartObject* keyboard_props() const = 0;
   virtual const smart_objects::SmartObject* menu_title() const = 0;
   virtual const smart_objects::SmartObject* menu_icon() const = 0;
-  virtual const smart_objects::SmartObject* day_color_scheme() const = 0;
-  virtual const smart_objects::SmartObject* night_color_scheme() const = 0;
-  virtual const std::string& display_layout() const = 0;
+  virtual smart_objects::SmartObject day_color_scheme() const = 0;
+  virtual smart_objects::SmartObject night_color_scheme() const = 0;
+  virtual std::string display_layout() const = 0;
```
JackLivio: Why change these from pointers and references?

mked-luxoft: `window_templates_` map was introduced in this feature, and we need to perform search for default window to access color schemes. Handling negative search cases is simpler with constructing objects. Given that null objects are safer than null ptrs and these smart objects are relatively small, memory overhead would not be so great.

AByzhynar: @JackLivio to add to the comment above - As well as there is no need to track that all dynamically allocated memory will be freed in the constructor as it was done before.

AStasiuk:
@theresalech, this PR is already reviewed and approved by Ford (by @kshala-ford) and ready for Livio review.

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/include/sdl_rpc_plugin/commands/hmi/on_bc_system_capability_updated_notification_from_hmi.h
```diff
@@ -0,0 +1,97 @@
...
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef SRC_COMPONENTS_APPLICATION_MANAGER_RPC_PLUGINS_SDL_RPC_PLUGIN_INCLUDE_SDL_RPC_PLUGIN_COMMANDS_HMI_ON_BC_SYSTEM_CAPABILITY_UPDATED_NOTIFICATION_FROM_HMI_H_
+#define SRC_COMPONENTS_APPLICATION_MANAGER_RPC_PLUGINS_SDL_RPC_PLUGIN_INCLUDE_SDL_RPC_PLUGIN_COMMANDS_HMI_ON_BC_SYSTEM_CAPABILITY_UPDATED_NOTIFICATION_FROM_HMI_H_
+
+#include "application_manager/commands/notification_from_hmi.h"
+
+namespace sdl_rpc_plugin {
```
JackLivio: I dont believe adding this RPC was mentioned in the proposal. Could you please describe the use cases for this HMI RPC?

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/on_button_event_notification.cc
```diff
@@ -102,8 +102,10 @@ void OnButtonEventNotification::Run() {
       return;
     }
 
-    if ((mobile_api::HMILevel::HMI_FULL != app->hmi_level()) &&
-        (mobile_api::HMILevel::HMI_LIMITED != app->hmi_level())) {
+    const mobile_apis::HMILevel::eType app_hmi_level =
```
JackLivio: Can buttons be in a widget outside of the default window? If yes, I believe this RPC needs to take into account the hmi level of the widget windows.

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/on_button_press_notification.cc
```diff
@@ -104,8 +104,10 @@ void OnButtonPressNotification::Run() {
     }
 
     // Send ButtonPress notification only in HMI_FULL or HMI_LIMITED mode
-    if ((mobile_api::HMILevel::HMI_FULL != app->hmi_level()) &&
-        (mobile_api::HMILevel::HMI_LIMITED != app->hmi_level())) {
+    const mobile_apis::HMILevel::eType app_hmi_level =
```
JackLivio: Same comment as OnButtonEvent

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/include/sdl_rpc_plugin/commands/hmi/on_bc_system_capability_updated_notification_from_hmi.h
AByzhynar: @JackLivio Use cases are described here: 
- https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0216-widget-support.md#automatic-subscription-to-display-and-window-capabilities

- It is not new RPC . It was discussed with @kshala-ford . As we already have OnSystemCapabilityUpdated notification - it was decided to update it in order to be sent from HMI to SDL.

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/set_display_layout_request.cc
```diff
@@ -154,7 +152,15 @@ void SetDisplayLayoutRequest::on_event(const event_engine::Event& event) {
                               hmi_response::templates_available);
           }
         }
+
+        // In case of successful response warn user that this RPC is deprecated
+        result_code = hmi_apis::Common_Result::WARNINGS;
```
JackLivio: This warning message should only be relayed to apps using RPC version that this rpc is deprecated for. ie apps >= 6.0 && apps < 7.0

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/on_button_event_notification.cc
dboltovskyi: @JackLivio Thank you for this finding. Defenetly button may belong to main or widget window. So appropriate window's HMI level needs to be taken into account.
We're preparing an update for core and scripts.

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/show_request.cc
```diff
@@ -94,6 +99,125 @@ void ShowRequest::HandleMetadata(const char* field_id,
...
+                                                         << " for window ID: "
+                                                         << current_window_id_);
+      layout_change_required_ = true;
+      return true;
+    }
+    LOG4CXX_DEBUG(logger_, "Show Request: No Layout Change");
+    return false;
+  };
+
+  const auto set_day_color_scheme = [&app, this]() -> bool {
```
JackLivio: I think set_day_color_scheme and set_night_color_scheme need to take into account if layout_change_required is true. If there is a color change requested when the template changes, and a color was previously set, it doesnt look like dcs_change_required_/ncs_change_required_ will ever be set to true.

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/show_request.cc
```diff
@@ -94,6 +99,125 @@ void ShowRequest::HandleMetadata(const char* field_id,
...
+    if (!template_config_.keyExists(strings::day_color_scheme)) {
+      return false;
+    }
+    if (app.day_color_scheme(current_window_id_).getType() !=
+            smart_objects::SmartType_Null &&
+        template_config_[strings::day_color_scheme] !=
+            app.day_color_scheme(current_window_id_)) {
+      // Color scheme param exists and has been previously set,
+      // hence do not allow color change
+      LOG4CXX_DEBUG(logger_, "Day Color Scheme change is rejected");
```
JackLivio: Also this information should probably be passed back to the app requesting the color change that was rejected

JackLivio: Nevermind i see the error response below

src/components/application_manager/src/application_data_impl.cc
```diff
@@ -316,18 +308,55 @@ const smart_objects::SmartObject* DynamicApplicationDataImpl::menu_icon()
   return menu_icon_;
 }
 
-const smart_objects::SmartObject* DynamicApplicationDataImpl::day_color_scheme()
+smart_objects::SmartObject DynamicApplicationDataImpl::day_color_scheme()
```
JackLivio: Isnt the color scheme addressable to different window ids? Should this method take the window id as a parameter to return the proper color scheme?

JackLivio: Never mind, I see the overloaded day_color_scheme method below

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/show_request.cc
mked-luxoft: As of this implementation,`Show` request duplicates logic of `SetDisplayLayout` request. As stated in [Color Scheme](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0147-template-color-scheme.md) proposal, 
`SDL Core should track the number of attempted SetDisplayLayout requests with the current template and REJECT any beyond the first with the reason "Using SetDisplayLayout to change the color scheme may only be done once.`

JackLivio: The key wording here is:

> with the current template

The developer should be able to change the color scheme after the initial color scheme any time the template layout is changed.

JackLivio: Here is the existing behavior in set display layout request. I believe this behavior should be transferred to the show requests logic handling of color schemes.

https://github.com/smartdevicelink/sdl_core/blob/master/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/set_display_layout_request.cc#L78-L82

JackLivio: Also referencing the comment in your code,

> // Template switched, hence allow any color change

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/show_request.cc
```diff
@@ -94,6 +99,125 @@ void ShowRequest::HandleMetadata(const char* field_id,
...
+  LOG4CXX_AUTO_TRACE(logger_);
+  if (helpers::Compare<mobile_apis::Result::eType, helpers::EQ, helpers::ONE>(
+          result,
+          mobile_apis::Result::SUCCESS,
+          mobile_apis::Result::WARNINGS)) {
+    if (layout_change_required_) {
+      const std::string new_layout =
+          template_config_[strings::template_layout].asString();
+      LOG4CXX_DEBUG(logger_, "New layout : " << new_layout << " is applied");
+      app.set_window_layout(current_window_id_, new_layout);
```
AByzhynar: @mked-luxoft Looks that you can just add something like 
```cpp
if (template_config_.keyExists(strings::day_color_scheme)) {
  app.set_day_color_scheme(current_window_id_,
	                                        template_config_[strings::day_color_scheme]);
 }

if (template_config_.keyExists(strings::night_color_scheme)) {
  app.set_night_color_scheme(current_window_id_,
	                                           template_config_[strings::night_color_scheme]);
 }
 return;
```
to solve the problem described by @JackLivio

src/components/application_manager/src/application_data_impl.cc
```diff
@@ -555,6 +664,33 @@ bool DynamicApplicationDataImpl::IsSubMenuNameAlreadyExist(
   return false;
 }
 
+DataAccessor<WindowParamsMap>
+DynamicApplicationDataImpl::window_optional_params_map() const {
+  return DataAccessor<WindowParamsMap>(window_params_map_,
+                                       window_params_map_lock_ptr_);
+}
+
+void DynamicApplicationDataImpl::AddWindowInfo(
```
JackLivio: I think "AddWindowInfo" implies that the info is appended to existing info. Because this method is overwriting window info for a specific window id, I think the method should be renamed "SetWindowInfo"

src/components/application_manager/src/application_impl.cc
```diff
@@ -1112,13 +1155,20 @@ void ApplicationImpl::set_audio_streaming_state(
                  " for non-media application to different from NOT_AUDIBLE");
     return;
   }
-  CurrentHmiState()->set_audio_streaming_state(state);
+
+  // According to proposal SDL-0216 audio and video streaming states should
```
JackLivio: The proposal only mentions audio streaming state. It does not mention video.

src/components/application_manager/src/display_capabilities_builder.cc
```diff
@@ -0,0 +1,128 @@
...
+  LOG4CXX_AUTO_TRACE(logger_);
+}
+
+void DisplayCapabilitiesBuilder::InitBuilder(
+    DisplayCapabilitiesBuilder::ResumeCallback resume_callback,
+    const smart_objects::SmartObject& windows_info) {
+  LOG4CXX_AUTO_TRACE(logger_);
+  sync_primitives::AutoLock lock(display_capabilities_lock_);
+  resume_callback_ = resume_callback;
+  window_ids_to_resume_.insert(kMainWindowID);
```
JackLivio: What is the difference between the terminology im seeing between "main windows" and "default window"?

src/components/application_manager/src/resumption/resume_ctrl_impl.cc
```diff
@@ -219,6 +230,89 @@ bool ResumeCtrlImpl::RestoreAppHMIState(ApplicationSharedPtr application) {
...
+  const auto& msg_params = (*request->second)[strings::msg_params];
+  const auto hmi_app_id = msg_params[strings::app_id].asInt();
+  auto application = application_manager_.application_by_hmi_app(hmi_app_id);
+  if (!application) {
+    LOG4CXX_ERROR(logger_,
+                  "Application is not registered by hmi id: " << hmi_app_id);
+    requests_msg_.erase(request);
+    return;
+  }
+  const WindowID window_id = msg_params[strings::window_id].asInt();
```
JackLivio: I see a lot of mixing this parameter as an Int and as a UInt.

src/components/application_manager/src/rpc_handler_impl.cc
```diff
@@ -159,10 +159,6 @@ void RPCHandlerImpl::ProcessMessageFromHMI(
   smart_objects::SmartObjectSPtr smart_object =
       std::make_shared<smart_objects::SmartObject>();
   bool allow_unknown_parameters = false;
-  if (!smart_object) {
```
JackLivio: Reason for this change?

src/components/application_manager/include/application_manager/application.h
JackLivio: 👍

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/include/sdl_rpc_plugin/commands/hmi/on_bc_system_capability_updated_notification_from_hmi.h
JackLivio: Even though BC.OnSystemCapabilityUpdated existed in the HMI API, defining a new direction (HMI->Core) is the same as creating a new RPC. I dont mind this part as much as I feel that creating this RPC is unnecessary. All of the display capabilities are already made known to core via UI.GetCapabilities from the HMI, or from the hmi_capabilities.json preloaded file. By adding this RPC there is now a third way to define display capabilities?

mked-luxoft: We need a way to get window capabilities after a certain window is created on HMI, by HMI sending `OnSystemCapabilitiesUpdated` notification. `UI.GetCapabilities` contains old display capabilities which are kept for backward compatibility and are used for RegisterAppInterfaceResponse.displayCapabilities.

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/on_button_event_notification.cc
mked-luxoft: @JackLivio fixed in 9862c6687b716c086c446c99ff0df04b5402bd2c

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/on_button_press_notification.cc
mked-luxoft: fixed in 9862c6687b716c086c446c99ff0df04b5402bd2c

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/set_display_layout_request.cc
mked-luxoft: it was discussed with @kshala-ford that this message needs to be present regardless app version

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/show_request.cc
mked-luxoft: @JackLivio fixed in ebb5dcc5575fe835d3b64522cd8015a00af55cc9

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/show_request.cc
mked-luxoft: applied in ebb5dcc5575fe835d3b64522cd8015a00af55cc9

src/components/application_manager/src/rpc_handler_impl.cc
mked-luxoft: Since `smart_object` is a shared ptr which is allocated with `std::make_shared`, i don't think that null pointer check makes sense here.

src/components/application_manager/src/application_data_impl.cc
mked-luxoft: fixed in 656568d5f8c8ff6c75ac80395d810fefe6c3a817

src/components/application_manager/src/application_impl.cc
mked-luxoft: seems that only the comment is wrong. changed in 1b37aa48ddb2672281a97d083bbe0c48453bada8

src/components/application_manager/src/display_capabilities_builder.cc
mked-luxoft: changed to `kDefaultWindow` in bf45a8d85b661b7faeacf726898b403d9978a3aa

src/components/application_manager/src/resumption/resume_ctrl_impl.cc
mked-luxoft: right, at the beginning it was unsigned type which was changed to signed one during internal review. changed all uses to signed in 4a8ee34c2fb59dd4cd8789df38f6642c95e19bb9

mked-luxoft:
@JackLivio your comments have been addressed, please take a look. Also, please note, that there are several fixup commits at the moment that will need to be squashed after your approval but before the merge.

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/include/sdl_rpc_plugin/commands/hmi/on_bc_system_capability_updated_notification_from_hmi.h
JackLivio: @mked-luxoft @kshala-ford 

Ok thank you for the clarification. The naming being used here is so confusing and I am just noticing the difference.

The struct used by OnSystemCapabilityUpdated is called DisplayCapability but the parameter is called displayCapabilities. While there is the older struct DisplayCapabilities that is used by GetCapabilities, and the parameter is also called displayCapabilities. Good thing one of these is being phased out.

Also, if DisplayCapabilities is deprecated in favor of window capabilities, then the DisplayCapabilitiesStruct itself should also be marked deprecated for version 6.0, not just in the parameter definition of the RAI. See "DisplayType" in the mobile_api for a reference.

mked-luxoft: @JackLivio fixed in e7420930e99dffc09102a070cbd753b260d5d4be

JackLivio:
I believe that there should be a WidgetSupport functional group added to the sdl_preloaded_pt.json.

JackLivio:
In testing we found that an app was able to send create window with associatedServiceType: MEDIA repeatedly.

```

   "msg_params" : {
      "associatedServiceType" : "MEDIA",
      "duplicateUpdatesFromWindowID" : 0,
      "type" : 1,
      "windowID" : 101,
      "windowName" : "Widget101"
   }


   "msg_params" : {
      "associatedServiceType" : "MEDIA",
      "duplicateUpdatesFromWindowID" : 0,
      "type" : 1,
      "windowID" : 102,
      "windowName" : "Widget102"
   }
```

According to the proposal:

> It is only allowed to have one window per service type. This means that a media app can only have a single MEDIA widget.

JackLivio:
In testing we found that we were able to create more widgets than the max number of widget windows defined in the system capabilities.

I believe this value should be enforced either in sdl core or by the hmi.

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/hmi/on_bc_system_capability_updated_notification_from_hmi.cc
```diff
@@ -0,0 +1,131 @@
...
+
+  LOG4CXX_DEBUG(logger_, "Updating display capabilities for app " << app_id);
+  app->set_display_capabilities(display_capabilities);
+
+  // Remove app_id from notification to mobile
+  (*message_)[strings::params][strings::connection_key] =
+      (*message_)[strings::msg_params][strings::app_id];
+  (*message_)[strings::msg_params].erase(strings::app_id);
+  if (app->is_resuming()) {
+    LOG4CXX_DEBUG(logger_, "Application is resuming");
```
JackLivio: After registering an app with the sdl hmi, this line is printing, but the app is not resumed into full. Also the RPC OnSystemCapability the hmi is sending is never being sent to mobile because of this caching logic and the failed resumption.

JackLivio: I think I am experiencing and issue with resumption. app->is_resuming() is true, but resumption is not taking place successfully.

JackLivio: PS. This resumption issue does not happen every time.

JackLivio:
Seeing an issue with GetSystemCapability.

- I register an app and get the systemCapabilityUpdate for the main window. ✅ 
- I call createWindow for a widget with window ID 101 and get the systemCapabilityUpdate for the widget window. ✅ 
- I then call getSystemCapability type:DISPLAYS and the response only includes the displayCapabilities for the widget window. It does not include the display capabilities for the main window ❌ .

The getSystemCapabilitiesResponse: 
[GetSystemCapabilitiesResponse.txt](https://github.com/smartdevicelink/sdl_hmi/files/3498455/GetSystemCapabilitiesResponse.txt)

dboltovskyi:
@JackLivio Please notice the issues described in comments bellow are corrected in https://github.com/smartdevicelink/sdl_core/pull/2980/commits/ee4e1c7782ffc095bddd1f19dfc475a854fb3ca8
```
1) https://github.com/smartdevicelink/sdl_core/pull/2980#issuecomment-520884024
2) https://github.com/smartdevicelink/sdl_core/pull/2980#issuecomment-520889265
3) https://github.com/smartdevicelink/sdl_core/pull/2980#issuecomment-520967303
```

mked-luxoft:
> I believe that there should be a WidgetSupport functional group added to the sdl_preloaded_pt.json.

@JackLivio added in 62cf9d425a8e8414383115d505585da93694da98

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -828,6 +828,10 @@ void RegisterAppInterfaceRequest::SendRegisterAppInterfaceResponseToMobile(
                                             application->mac_address());
   }
 
+  if (resumption) {
```
JackLivio: I think there should be an additional condition to check that result_code != RESUME_FAILED. Alternative approach would be to set resumption to false in the event of RESUME_FAILED result code set above.

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/set_display_layout_request.cc
JackLivio: @mked-luxoft Sorry but I don't agree with this response. Sending a warning result code could break an apps implementation that negotiates an RPC spec version where SetDisplayLayout is completely valid. According to the versioning in the API, this RPC is only deprecated for apps using rpc spec version 6.x, therefore only 6.x apps should be receiving the warning.

kshala-ford: Hello @JackLivio agree. To be clear here: For version < 6.0 the result code should be `SUCCESS`. For version >= 6.0 it should be `WARNINGS`. 

There should not be a condition like `apps < 7.0` as this is not specified.

AByzhynar: @JackLivio Thank you. Fixed in f8085a5316b7b0deef654c69489a63b33d444a7f

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/register_app_interface_request.cc
AByzhynar: @JackLivio Thank you. Fixed in d38383d186ecab67a35a7acd90aeaf8a94832b93

AByzhynar: @JackLivio FYI. PR was rebased to the lates develop as new features were merged. All physical and logical issues were fixed. Unit tests fixed/updated. ATF scripts were rebased as well.

JackLivio:
Thank you for the updates. For future reference could you please try to avoid force pushing? Its hard to verify the new updates when there are all of these extra commits showing since the last comments.

AByzhynar:
> Thank you for the updates. For future reference could you please try to avoid force pushing? Its hard to verify the new updates when there are all of these extra commits showing since the last comments.

@JackLivio Sorry for that. We will not rebase anymore until review is finished. Thank you.

JackLivio:
@mked-luxoft @AByzhynar When testing the new changes related to successful resumption, if we connect an app and force the app to send the wrong hash id, it is causing core to crash on this error.

```
terminate called after throwing an instance of 'std::bad_function_call'
  what():  bad_function_call
./start.sh: line 2:  8041 Aborted                 (core dumped) LD_LIBRARY_PATH=. ./smartDeviceLinkCore

```

We tested this repeatedly and it occurred every time. We tested this on a separate branch of core to see if it was an inherent issue in the project and the other branch did not see such crashes.

AByzhynar:
> @mked-luxoft @AByzhynar When testing the new changes related to successful/unsuccessful resumption, if we connect an app and force the app to send the wrong hash id, it is causing core to crash on this error.
> 
> ```
> terminate called after throwing an instance of 'std::bad_function_call'
>   what():  bad_function_call
> ./start.sh: line 2:  8041 Aborted                 (core dumped) LD_LIBRARY_PATH=. ./smartDeviceLinkCore
> ```
> 
> We tested this repeatedly and it occurred every time. We tested this on a separate branch of core to see if it was an inherent issue in the project and the other branch did not see such crashes.
> 
> I am still seeing this log which is concerning since resumption was failed.
> 
> > DEBUG [14:35:39,947][Commands] Application is resuming
> 
> ^ log is located in on_bc_system_capability_updated_notification_from_hmi.
> 
> The app object is set to resuming true in ResumeCtrlImpl::SetAppHMIState.
> 
> > application->set_is_resuming(true);

@JackLivio fixed in 0e485c61c13b900983fd0a58af32c38e0505b505
Added detailed comment to the code. 
Also app has to be in resuming state even if hashes don't match as we have 2 part of resumption: 
- App data (resumed only if hashes match) 
- App HMI Level which is resumed regardless of hash
So here: 
   The app object is set to resuming true in ResumeCtrlImpl::SetAppHMIState.
    application->set_is_resuming(true); // means that SDL tries to resume app HMI level

JackLivio:
@AByzhynar Thank you for the clarification. 

What category do widget windows fall under? We send a wrong hash id on a resumption but the HMI still receives all of the create windows for widgets created in the previous session. My initial assumption is that widgets would fall under the App Data resumption category and require a correct hash id.

AByzhynar:
> @AByzhynar Thank you for the clarification.
> 
> What category do widget windows fall under? We send a wrong hash id on a resumption but the HMI still receives all of the create windows for widgets created in the previous session. My initial assumption is that widgets would fall under the App Data resumption category and require a correct hash id.

@JackLivio You are absolutely right. Widgets is app data. So you have found another issue. I will analyze and fix it . Thank you.

AByzhynar:
> > @AByzhynar Thank you for the clarification.
> > What category do widget windows fall under? We send a wrong hash id on a resumption but the HMI still receives all of the create windows for widgets created in the previous session. My initial assumption is that widgets would fall under the App Data resumption category and require a correct hash id.
> 
> @JackLivio You are absolutely right. Widgets is app data. So you have found another issue. I will analyze and fix it . Thank you.

@JackLivio Issue is fixed in 55c67ea33906799127c45ddfad165571b2d98103

PR #2982, opened 2019-08-02
Avoid converting to Json non object values

LuxoftAKutsan:
Json::write adds additional `"` to string type as json value.
But string parameters should not contain additional `"`


atiwari9:
Approved

PR #2985, opened 2019-08-06
Fix messages conversion to json

mkorniichuk:

### Summary
The conversion of messages to JSON was commented in external policy flow due to some reason.
So the data persists in the policy table, but not in JSON view of the same table.
Current PR brings back messages to the JSON representation of PT.

### CLA


LuxoftAKutsan:
This minor issue, because HMI  is able to get ConsumerFriendlyMessages ftom SDL without using GetPolicyConfigurationData RPC. 
And Intorodusing of this fix will increase Policy snapshot size. 
But for GetPolicyConfigurationData will be created after merging https://github.com/smartdevicelink/sdl_core/pull/2977 .
Better solution should be descussed to avoid extension of Policy snapshot for external proprietary flow.

PR #3005, opened 2019-09-02
Add check for already stored vehicle data item params

### LINKED ISSUE #3003: PTU trigger on ignition cycles doesn't work properly
### Bug Report
When the amount of ignition cycles notified by HMI via `BasicCommunication.OnIgnitionCycleOver` gets equal to the value of
-- `exchange_after_x_ignition_cycles` field (`module_config` section) of policies database, SDL must trigger a PTU sequence.

Currently SDL sometimes doesn't increment it's internal counter of ignition cycles

##### Reproduction Steps
1. Make sure `exchange_after_x_ignition_cycles` in local PT = 5
1. Start SDL, HMI, connect Mobile, register App
2. Perform 5 ignition cycles

##### Expected Behavior
SDL starts PTU on a 5th ignition cycle

##### Observed Behavior
SDL doesn't start PTU on a 5th ignition cycle

* SDL Core Version: https://github.com/smartdevicelink/sdl_core/tree/develop (e7b5c75)
* ATF: https://github.com/smartdevicelink/sdl_atf/tree/develop
* Testing Against: ATF (script, report and logs are in attachment)

- [build_options/109_ATF_PTU_Trigger_IGN_cycles_PROPRIETARY](https://github.com/smartdevicelink/sdl_atf_test_scripts/blob/develop/test_scripts/Policies/build_options/109_ATF_PTU_Trigger_IGN_cycles_PROPRIETARY.lua)
- [2019-09-01_01-00-01.tar.gz](https://github.com/smartdevicelink/sdl_core/files/3563110/2019-09-01_01-00-01.tar.gz)

mkorniichuk:
Fixes #3003

### Risk

### Summary
During vehicle data item parameter storage to DB, the new (and the same) row will be inserted to parent-parameter relation table for each one's version. As a result, during the retrieve of a composite vehicle data item, such parameter could be retrieved several times. And after several cycles, parameter array will exceed max boundary and therefore will be invalid, so the storage process fails and some data (including meta as `exchange_after_x_ignition_cycles`) will not be stored.

### CLA

### DIFF
--- a/src/components/policy/policy_external/src/sql_pt_representation.cc
+++ b/src/components/policy/policy_external/src/sql_pt_representation.cc
@@ -2660,9 +2660,16 @@ bool SQLPTRepresentation::InsertVehicleDataItem(
   }
 
   if (vehicle_data_item.params->is_initialized()) {
+    std::map<std::string, std::string> stored_vehicle_data_item_params;
     for (const auto& param : *(vehicle_data_item.params)) {
       if (!InsertVehicleDataItem(param)) {
         return false;
+      }
+
+      if (stored_vehicle_data_item_params.end() !=
+          stored_vehicle_data_item_params.find(param.name)) {
+        LOG4CXX_DEBUG(logger_, "Parameter already stored.");
+        continue;
       }
 
       if (!query.Prepare(sql_pt::kInsertVehicleDataItemParams)) {
@@ -2685,6 +2692,7 @@ bool SQLPTRepresentation::InsertVehicleDataItem(
                 << ". Error: " << query.LastError().text());
         return false;
       }
+      stored_vehicle_data_item_params[param.name] = param.key;
     }
   }
 
--- a/src/components/policy/policy_regular/src/sql_pt_representation.cc
+++ b/src/components/policy/policy_regular/src/sql_pt_representation.cc
@@ -2687,10 +2687,17 @@ bool SQLPTRepresentation::InsertVehicleDataItem(
   }
 
   if (vehicle_data_item.params->is_initialized()) {
+    std::map<std::string, std::string> stored_vehicle_data_item_params;
     for (const auto& param : *(vehicle_data_item.params)) {
       if (!InsertVehicleDataItem(param)) {
         return false;
       }
+
+      if (stored_vehicle_data_item_params.end() !=
+          stored_vehicle_data_item_params.find(param.name)) {
+        LOG4CXX_DEBUG(logger_, "Parameter already stored.");
+        continue;
+      };
 
       if (!query.Prepare(sql_pt::kInsertVehicleDataItemParams)) {
         LOG4CXX_ERROR(logger_,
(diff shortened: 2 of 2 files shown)


src/components/policy/policy_external/src/sql_pt_representation.cc
```diff
@@ -2660,11 +2660,18 @@ bool SQLPTRepresentation::InsertVehicleDataItem(
...
 
   if (vehicle_data_item.params->is_initialized()) {
+    std::map<std::string, std::string> stored_vehicle_data_item_params;
     for (const auto& param : *(vehicle_data_item.params)) {
       if (!InsertVehicleDataItem(param)) {
         return false;
       }
 
+      if (stored_vehicle_data_item_params.end() !=
+          stored_vehicle_data_item_params.find(param.name)) {
```
jacobkeeler: It doesn't seem like this would work for parameters which have multiple definitions, such as:
```
                        {
                            "name": "utcMonth",
                            "key": "OEM_REF_UTC_MONTH",
                            "type": "Integer",
                            "mandatory": false,
                            "minvalue": 1,
                            "maxvalue": 12,
                            "since": "5.0"
                        },
                        {
                            "name": "utcMonth",
                            "key": "OEM_REF_UTC_MONTH",
                            "type": "Integer",
                            "mandatory": true,
                            "minvalue": 1,
                            "maxvalue": 12,
                            "since": "2.0",
                            "until": "5.0"
                        },
```

mkorniichuk: @jacobkeeler the point is it actually shouldn't. `InsertVehicleDataItem` will store both items from your example to `vehicle_data_item_definitions` table. 
But the `vehicle_data_item_parameters` table should contain only a single row with name and key of the parameter. Further, when retrieving items, using the name and key from `vehicle_data_item_parameters` table, both of these items will be retrieved.

jacobkeeler: This was by design, to my knowledge. The correct schema would be selected based on the version of the mobile application. 

If we aren't going to be handling versioning for these parameters, then older definitions of parameters should be removed from the preloaded pt. Otherwise there is some ambiguity over which schema definition would be used. In this case, I suppose the best approach would be to use the latest definition, but use the original `since` value:
```
                        {
                            "name": "utcMonth",
                            "key": "OEM_REF_UTC_MONTH",
                            "type": "Integer",
                            "mandatory": false,
                            "minvalue": 1,
                            "maxvalue": 12,
                            "since": "2.0"
                        },
```

mkorniichuk: @jacobkeeler Versioning is okay, and yes, you are right, this was by design. This fix is needed to make it work as it was by design. Now it duplicates results upon request of the item (as many times as you find same name in the table)

jacobkeeler: @mkorniichuk 👍 Alright, looking at the structure of the tables, I can see what you're saying now

PR #3034, opened 2019-09-17
Fix IAP device transport switching sequence

AKalinich-Luxoft:
Fixes [#3044_core](https://github.com/smartdevicelink/sdl_core/issues/3044) and  [#2262_scripts](https://github.com/smartdevicelink/sdl_atf_test_scripts/issues/2262)

### Risk

### Testing Plan
Covered with ATF scripts

### Summary
There were done 2 main changes to fix that issue:
1. Update `IsApplicationSwitched()` function - In order to check device transport switching properly,
SDL should check device handle of initial application and is it in reconnection list. In the current impl
SDL wrongly checked current device handle instead of initial, as a result apps have never be evaluated as reconnecting.
2. Update `CommandHolder` class - After introduction `CommandSource` enum there was possible to specify any different source for HMI and mobile commands. However, `CommandHolder` class
logic was not updated to reflect these changes and still working with the hardcoded values. As a result, commands factory of plugins may not able to create commands which were suspended during device transport switching and then restored back. This commit contains updated `CommandHolder` impl which reflects changes done and properly handles initial command source.

### Tasks Remaining:
- [x] [Fix affected unit tests]
- [x] [Finish internal review]
- [x] [Check with full regression test]

### CLA


AByzhynar:
@AKalinich-Luxoft This PR could break feature “The same app from multiple devices”. Did you check regression tests?

AKalinich-Luxoft:
@theresalech PR is ready for Livio team review

AKalinich-Luxoft:
> @AKalinich-Luxoft This PR could break feature “The same app from multiple devices”. Did you check regression tests?

@AByzhynar yes, full ATF regression testing were done and it's passed

PR #3066, opened 2019-10-03
Fix latest coverity issues

jacobkeeler:
Fixes CID 204315, 204316

### Risk

### Summary
Fix potential resource leak and minor dead code

##### Bug Fixes
* Fix potential resource leak and minor dead code

### CLA

### DIFF
--- a/src/components/application_manager/src/policies/policy_handler.cc
+++ b/src/components/application_manager/src/policies/policy_handler.cc
@@ -337,8 +337,11 @@ bool PolicyHandler::LoadPolicyLibrary() {
 bool PolicyHandler::CreateManager() {
   void* policy_handle = dlopen(kLibrary.c_str(), RTLD_LAZY);
   const char* error = dlerror();
-  if (error) {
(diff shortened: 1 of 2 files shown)


src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -337,8 +337,11 @@ bool PolicyHandler::LoadPolicyLibrary() {
 bool PolicyHandler::CreateManager() {
   void* policy_handle = dlopen(kLibrary.c_str(), RTLD_LAZY);
   const char* error = dlerror();
-  if (error) {
-    LOG4CXX_ERROR(logger_, error);
+  if (!policy_handle) {
```
iCollin: `const char* error = dlerror();` isn't needed outside this block

jacobkeeler: Technically speaking you are correct, though it seems standard to call `dlerror` after a related operation. This deque's the first unread error message, if available. It might not be relevant in this case, but I don't see much harm in including it here

PR #3068, opened 2019-10-08
Replace DCHECK in Create Window Request

JackLivio:

### Risk

### Testing Plan
- [Use HMI that does not include the widget type in its capabilities. Base widgets branch.](https://github.com/smartdevicelink/generic_hmi/tree/feature/base_widgets_rpc_support)
- Connect app and send a CreateWindow Request.

### Summary
Replace a DCHECK that was crashing core with an if condition checking against the find results instead.

### CLA


src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/create_window_request.cc
```diff
@@ -333,7 +333,10 @@ bool CreateWindowRequest::DoesExceedMaxAllowedWindows(
         return false;
       });
 
-  DCHECK(find_res != windowTypeSupported->end());
+  if (find_res == windowTypeSupported->end()) {
+    LOG4CXX_WARN(logger_, "Requested Window Type is not supported by the HMI");
+    return false;
```
jacobkeeler:
```suggestion
    return true;
```
?
It seems to me that the error case would be appropriate here.

PR #3071, opened 2019-10-10
Add select function to pipe write call

### LINKED ISSUE #3072: Pipe Streaming Thread Blocks Indefinitely
### Bug Report
When doing a video stream with the Pipe streamer configuration, if there is no pipe consumer (aka user did not start gstreamer), the pipe file descriptor becomes full / busy and the pipe streamer write call will block indefinitely.

##### Reproduction Steps
1. Connect nav app
2. Start video stream without starting gstreamer
3. Shutdown Core

##### Expected Behavior
Core shuts down gracefully.

##### Observed Behavior
Core hangs on video streaming thread write call.

* OS/Version: Android and Ubuntu 18
* SDL Core Version: release/6.0.0 release candidate / develop
* Testing Against: Android Test Suite application

JackLivio:
Fixes #3072 

### Risk

### Testing Plan
Connect navigation app and start a video stream.

### Summary
Adds a select call before a write to the pipe file descriptor. This select function checks to see if the pipe fd is ready to receive data within a 10 second specified timeout. 

This fixes an issue where if the HMI is not consuming the video streaming data, the pipe streamer file descriptor would block the thread forever after a couple of seconds of streaming.

If the FD resource becomes full/busy, Core will disconnect and fail the stream.

### CLA

### DIFF
--- a/src/components/media_manager/src/pipe_streamer_adapter.cc
+++ b/src/components/media_manager/src/pipe_streamer_adapter.cc
@@ -112,34 +112,47 @@ bool PipeStreamerAdapter::PipeStreamer::Send(
   struct timeval tv;
   tv.tv_sec = 10;
   tv.tv_usec = 0;
-  int retval = select(pipe_fd_ + 1, NULL, &wfds, NULL, &tv);
-  if (retval == -1) {
-    LOG4CXX_ERROR(logger_,
-                  "Failed writing data to pipe "
-                      << named_pipe_path_ << ". Errno: " << strerror(errno));
-    return false;
-  } else if (retval) {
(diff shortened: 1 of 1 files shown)


iCollin:
I thought we had agreed to not drop any data. I don't know that it is worth it to drop partial frames from the stream to fix a weird case of nothing consuming the pipe.

However, in my testing of playing a 2 minute 42 second video (17.7MB), only one frame was partially dropped so I have approved this implementation. GStreamer was able to handle the lost data, and no issue was caused for our use-case.
It is up to the author to decide which defect is preferable.

In a worst-case scenario, I imagine a different HMI consuming data from the pipe in small increments, causing many frames to be partially dropped and degrading playback performance or even causing a failure to play.

PR #3099, opened 2019-10-22
Synchronize backup and reset policy functions

### LINKED ISSUE #3098: ExpectPolicyTableReset unit test is unstable
### Bug Report
Sometimes (~30%) `ResetPolicyTable_WithPreloadedFile_ExpectPolicyTableReset` test which is in `policy_handler_test` may fail.

##### Reproduction Steps
1. Build SDL with UT
2. Run described unit test
3. Check results

##### Expected Behavior
Test is passed

##### Observed Behavior
Test is failed with message:
```
11:40:58 Uninteresting mock function call - taking default action specified at:
11:40:58 /src/components/application_manager/test/policy_handler_test.cc:176:
11:40:58     Function call: event_dispatcher()
11:40:58           Returns: @0x326e938 8-byte object <F8-A0 E9-02 00-00 00-00>
11:40:58 Stack trace:
11:40:58 /src/components/application_manager/test/policy_handler_test.cc:370: Failure
11:40:58 Value of: policy_handler_.ResetPolicyTable()
11:40:58   Actual: false
11:40:58 Expected: true
```

* SDL Core Version: https://github.com/smartdevicelink/sdl_core/tree/develop (1d3a401)
* Testing Against: N/A

  - Build log: [build.log.tar.gz](https://github.com/smartdevicelink/sdl_core/files/3757646/build.log.tar.gz)

AKalinich-Luxoft:
Fixes #3098 

### Risk

### Testing Plan
Covered with unit tests

### Summary
This is an issue specific for `EXTERNAL_PROPRIETARY` flow only. This issue is exposed by `PolicyHandlerTest` unit test: `ResetPolicyTable_WithPreloadedFile_ExpectPolicyTableReset`
which sometimes is failing.

The main problem was that database reset(drop) and backup can be done simultaneously in different threads which were not synchronized. In that case, database drop often may be failed because another thread is trying to write something into the DB table which is going to be dropped. This causes above unit test to fail sometimes.

To make that unit test and policy component more stable, these two threads were synchronized by additing a waiting function, so one thread will be able to wait for another thread to finish its own staff.

### CLA

### DIFF
--- a/src/components/policy/policy_external/src/cache_manager.cc
+++ b/src/components/policy/policy_external/src/cache_manager.cc
@@ -2682,10 +2682,14 @@ bool CacheManager::LoadFromFile(const std::string& file_name,
 bool CacheManager::ResetPT(const std::string& file_name) {
   LOG4CXX_AUTO_TRACE(logger_);
   is_unpaired_.clear();
+
+  backuper_->WaitForBackupIsDone();
+
   if (!backup_->RefreshDB()) {
     LOG4CXX_ERROR(logger_, "Can't re-create policy database. Reset failed.");
     return false;
   }
+
   sync_primitives::AutoLock lock(cache_lock_);
   pt_.reset(new policy_table::Table());
   const bool result = LoadFromFile(file_name, *pt_);
@@ -3097,6 +3101,7 @@ void CacheManager::OnDeviceSwitching(const std::string& device_id_from,
 CacheManager::BackgroundBackuper::BackgroundBackuper(
     CacheManager* cache_manager)
     : cache_manager_(cache_manager)
+    , backup_is_in_progress_(false)
     , stop_flag_(false)
     , new_data_available_(false) {
   LOG4CXX_AUTO_TRACE(logger_);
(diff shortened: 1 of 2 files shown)


src/components/policy/policy_external/src/cache_manager.cc
```diff
@@ -3147,6 +3162,16 @@ void CacheManager::BackgroundBackuper::DoBackup() {
   backup_notifier_.NotifyOne();
 }
 
+void CacheManager::BackgroundBackuper::WaitForBackupIsDone() {
+  LOG4CXX_AUTO_TRACE(logger_);
+  sync_primitives::AutoLock auto_lock(backup_done_lock_);
+  if (!backup_is_in_progress_) {
```
AByzhynar: @AKalinich-Luxoft I would rewrite it in more simple way:
```cpp
  if (backup_is_in_progress_) {
    backup_done_.Wait(auto_lock); 
  }
```
instead of this one: 

```cpp
  if (!backup_is_in_progress_) {
    return;
  }
  backup_done_.Wait(auto_lock);
```

Can we do it?

AKalinich-Luxoft: @AByzhynar yes, looks much cleaner. Thanks
Fixed in https://github.com/smartdevicelink/sdl_core/pull/3099/commits/4a60b806dbcad635a765411ad4588b01017baffc

src/components/policy/policy_external/src/cache_manager.cc
```diff
@@ -3121,13 +3126,23 @@ void CacheManager::BackgroundBackuper::threadMain() {
   LOG4CXX_AUTO_TRACE(logger_);
   sync_primitives::AutoLock lock(need_backup_lock_);
   while (!stop_flag_) {
+    backup_is_in_progress_.exchange(true);
     {
       sync_primitives::AutoUnlock need_backup_lock(need_backup_lock_);
       InternalBackup();
     }
+
     if (new_data_available_ || stop_flag_) {
```
LuxoftAKutsan: @AKalinich-Luxoft this case will exit while loop with `backup_is_in_progress_=true`. It it correct?

AKalinich-Luxoft: @LuxoftAKutsan yes, looks like potentially some thread which is in `WaitForBackupIsDone()` might be stuck if backuper thread will be requested to stop. It's a good point. I moved `backup_is_in_progress_` setting before `continue` so now there will be a guarantee that `backup_is_in_progress_` will never be set to `true` if loop was exited by some reason.
See https://github.com/smartdevicelink/sdl_core/pull/3099/commits/32d697135fa021d2cf45366d4ebf0ac9e941accd

PR #3146, opened 2019-11-28
Test for double data cleanup

### LINKED ISSUE #3135: CCB:  Fix undefined behavior after SO cleanup
### Bug Report
SDL crash in the random places.

##### Reproduction Steps
This issue is often observed during cable unplug
while audio/video streaming is active.

##### Expected Behavior
SDL no crashes

##### Observed Behavior
SDL crash in the random places.

#### OS & Version Information

OS/Version: Ubuntu 16.04
SDL Core Version: https://github.com/smartdevicelink/sdl_core/tree/develop (e137556512215ded94228e33d21ab4c3e1af3017)
Testing : Unit tests

SNiukalov:
Fixes #3135

### Risk

### Testing Plan
Stability test

### Summary
After `cleanup_data()` call, `delete` is used to deallocate memory by pointer, however pointer is still holding a value of address of the already deallocated memory. By that reason SO still have an ability to access that data what might cause UB including core crashes in the random places.

This issue is often observed during cable unplug while audio/video streaming is active. To avoid such problem, NULL pointer is assigned explicitly to avoid unwanted memory access. Also this fixes possible "double free memory" issues.

### CLA

### DIFF
--- a/src/components/smart_objects/test/CMakeLists.txt
+++ b/src/components/smart_objects/test/CMakeLists.txt
@@ -53,5 +53,11 @@ set(EXCLUDE_PATHS
   SmartObjectConvertionTime_test.cc
 )
 
+# Enable detect Double-free, invalid free
+# AddressSanitizer is a fast memory error detector. 
+# It consists of a compiler instrumentation module and a run-time library.
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -fno-omit-frame-pointer -fsanitize=address")
+set(CMAKE_LINKER_FLAGS_DEBUG "${CMAKE_LINKER_FLAGS_DEBUG} -fno-omit-frame-pointer -fsanitize=address")
+
 collect_sources(SOURCES "${CMAKE_CURRENT_SOURCE_DIR}" "${EXCLUDE_PATHS}")
 create_test(smart_object_test "${SOURCES}" "${LIBRARIES}")
--- a/src/components/smart_objects/test/SmartObjectUnit_test.cc
+++ b/src/components/smart_objects/test/SmartObjectUnit_test.cc
@@ -31,6 +31,7 @@
  */
 
 #include "gmock/gmock.h"
+#define final  // Disable error: cannot derive from ‘final’ base
 #include "smart_objects/smart_object.h"
 
 namespace test {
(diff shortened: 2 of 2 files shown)


src/components/smart_objects/test/SmartObjectUnit_test.cc
```diff
@@ -585,6 +586,26 @@ TEST(MapEraseTest, SmartObjectTest) {
...
+  };
+
+  DerivedSmartObject obj(SmartType_String);
+  ASSERT_EQ(SmartType_String, obj.getType());
+
+  obj = "test string_value";
+  ASSERT_EQ(std::string("test string_value"), obj.asString());
+
+  obj.cleanup_data();
+  obj.cleanup_data();
```
AByzhynar: @SNiukalov It's confusing...Why did you call it twice?

src/components/smart_objects/src/smart_object.cc
```diff
@@ -734,16 +734,28 @@ void SmartObject::duplicate(const SmartObject& OtherObject) {
 void SmartObject::cleanup_data() {
   switch (m_type) {
     case SmartType_String:
-      delete m_data.str_value;
+      if (m_data.str_value) {
+        delete m_data.str_value;
+        m_data.str_value = NULL;
```
AByzhynar: @SNiukalov Use `nullptr` everywhere instead of `NULL`

SNiukalov: Fixed in:https://github.com/smartdevicelink/sdl_core/pull/3146/commits/6915e49d17c1a2bb31e654ab1fa2fc3b668b9040

src/components/smart_objects/test/SmartObjectUnit_test.cc
SNiukalov: @AByzhynar 
In this case we try testing stability work clean_data function.
In previously implementation we will get crash or indefine behevior after twice call this function.
Unit tests will be finished with aborting.
But now this test case will working correctly.

Fixed in:https://github.com/smartdevicelink/sdl_core/pull/3146/commits/6915e49d17c1a2bb31e654ab1fa2fc3b668b9040

src/components/smart_objects/test/CMakeLists.txt
```diff
@@ -53,5 +53,11 @@ set(EXCLUDE_PATHS
   SmartObjectConvertionTime_test.cc
 )
 
+# Enable detect Double-free, invalid free
+# AddressSanitizer is a fast memory error detector. 
+# It consists of a compiler instrumentation module and a run-time library.
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -fno-omit-frame-pointer -fsanitize=address")
```
AKalinich-Luxoft: @SNiukalov do you really need address sanity enabled for unit tests? I believe this tool for debug purposes only

SNiukalov: @AKalinich-Luxoft 
Yes you are right. But without this option, we can't detect Double-free, becuase test always will finish normal. But with this option test will be aborted and show backtrace.

SNiukalov:
@theresalech PR is ready for Livio team review. Thank you.

src/components/smart_objects/src/smart_object.cc
```diff
@@ -734,16 +734,28 @@ void SmartObject::duplicate(const SmartObject& OtherObject) {
...
+      if (m_data.array_value) {
+        delete m_data.array_value;
+        m_data.array_value = nullptr;
+      }
       break;
     case SmartType_Binary:
-      delete m_data.binary_value;
+      if (m_data.binary_value) {
+        delete m_data.binary_value;
+        m_data.array_value = nullptr;
```
iCollin: while it's functionally the same, this is a copy paste error :)

SNiukalov:
@iCollin 
Thanks, yes, you're right.
All changes for smart_object.cc were discarded
This mistake occurred because we have different issues (3135 and 3109)
PR rebased to the last develop commit.

Issue #3138, opened 2019-11-25
Sometimes SDL sends BasicCommunication.PolicyUpdate twice


LitvinenkoIra:
### Bug Report
SDL sends BasicCommunication.PolicyUpdate twice

##### Occurrences:
Sometimes (5%) 

##### Reproduction Steps
1. SDL is built with EXTERNAL_PROPRIETARY policy flag
2. HMI and SDL are started
3. App is registered and activated

##### Expected Behavior
SDL->HMI: BasicCommunication.PolicyUpdate

##### Observed Behavior
SDL send BasicCommunication.PolicyUpdate twice

##### OS & Version Information
* OS/Version: Ubuntu 18.04
* SDL Core Version: https://github.com/smartdevicelink/sdl_core/commit/891497bc9fb23434242dbe92e1f13dd1273fcb21
* ATF: https://github.com/smartdevicelink/sdl_atf/commit/37da7aba8929e14dbd692befdea8540c5dd1dbbb
* Script: [308_ATF_Check_app_registration_language_gui.zip](https://github.com/smartdevicelink/sdl_core/files/3887258/308_ATF_Check_app_registration_language_gui.zip)

##### Logs
[BC.PolicyUpdate_twice.zip](https://github.com/smartdevicelink/sdl_core/files/3887247/BC.PolicyUpdate_twice.zip)

JackLivio:
Fixed by #3203

PR #3203, opened 2019-12-27
Fix BC.PolicyUpdate double sending

LitvinenkoIra:
Fixes https://github.com/smartdevicelink/sdl_core/issues/3138

### Risk

### Testing Plan
http://172.26.147.12/view/Issue_3138/

### Summary
Sometimes SDL sends BasicCommunication.PolicyUpdate twice.
I believe that the main problem is that when SDL receives the second PTU trigger, in the PolicyManagerImpl::StartPTExchange() ->  update_status_manager_.IsUpdatePending() returned **false** because the state is still "UPDATE_NEEDED". 

On the latest release - SDL should change the status to "UPDATING" right after sending the BasicCommunication.PolicyUpdate to the HMI. But in some cases, SDL can process the second PTU trigger before changing the status to the "UPDATING".

Taking into account all the above I have created another one internal status "UPDATE_PENDING" and use this status for cases when SDL knows that update is required and until the snapshot will be sent to the HMI. 
So SDL will change the status to "UPDATE_PENDING" before calling RequestPTUpdate(); in PolicyManagerImpl::StartPTExchange(); -> IsUpdatePending() for "UPDATE_PENDING" will return **true**; ->  After sending the first BC.PolicyUpdate the status will be changed to "UPDATING" -> SDL won't request a PolicyUpdate twice

### DIFF
--- a/src/components/policy/policy_external/include/policy/status.h
+++ b/src/components/policy/policy_external/include/policy/status.h
@@ -163,7 +163,8 @@ class UpdateNeededStatus : public Status {
 };
 
 /**
- * @brief The UpdatePendingStatus class represents 'update pending' status
+ * @brief The UpdatePendingStatus class represents cases when SDL knows that an
+ * update is required and but before the snapshot is sent to the HMI
  */
 class UpdatePendingStatus : public Status {
  public:
--- a/src/components/policy/policy_external/include/policy/status.h
+++ b/src/components/policy/policy_external/include/policy/status.h
@@ -52,6 +52,7 @@ enum UpdateEvent {
   kOnResetPolicyTableRequireUpdate,
   kOnResetPolicyTableNoUpdate,
   kScheduleUpdate,
+  kPendingUpdate,
   kScheduleManualUpdate,
   kOnResetRetrySequence,
   kNoEvent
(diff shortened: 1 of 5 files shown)


src/components/policy/policy_external/include/policy/update_status_manager.h
```diff
@@ -149,6 +149,8 @@ class UpdateStatusManager {
    */
   void ScheduleUpdate();
 
+  void PendingUpdate();
```
mkorniichuk: @LitvinenkoIra how about doxygen?

src/components/policy/policy_external/src/update_status_manager.cc
```diff
@@ -164,9 +165,15 @@ bool UpdateStatusManager::IsUpdatePending() const {
...
+  LOG4CXX_AUTO_TRACE(logger_);
   ProcessEvent(kScheduleUpdate);
 }
 
+void UpdateStatusManager::PendingUpdate() {
+  LOG4CXX_AUTO_TRACE(logger_);
+  ProcessEvent(kPendingUpdate);
+}
+
 void UpdateStatusManager::ScheduleManualUpdate() {
```
mkorniichuk: @LitvinenkoIra good place for **`LOG4CXX_AUTO_TRACE`**

JackLivio:
@LitvinenkoIra Was the original issue only seen in external policies?

src/components/policy/policy_external/include/policy/status.h
```diff
@@ -161,6 +162,37 @@ class UpdateNeededStatus : public Status {
   bool IsUpdateRequired() const OVERRIDE;
 };
 
+/**
+ * @brief The UpdatePendingStatus class represents 'update pending' status
```
JackLivio: I think the description of the class should be more functional. The reasoning you gave in the PR description was really good. Maybe use some of that description here?

> The UpdatePendingStatus class represents cases when SDL knows that update is required and but before the snapshot is sent to the HMI.

LitvinenkoIra:
@JackLivio Yes, the original issue was only seen in external policy in case SDL receives two PTU triggers one by one

src/components/policy/policy_external/src/update_status_manager.cc
LitvinenkoIra: @mkorniichuk Done in 7d5222c6ca6e483113b51eae3332d9e94ec5444a

src/components/policy/policy_external/include/policy/status.h
LitvinenkoIra: @JackLivio Done in 7d5222c6ca6e483113b51eae3332d9e94ec5444a

dboltovskyi:
@theresalech Please notice this PR is ready for PM review

PR #3215, opened 2020-01-17
Fix potential deadlock in application manager

### LINKED ISSUE #3158: CCB: Fixing deadlock during transport switching
### Bug Report
SDL crash observed when iPhone device is connected via BT over Wifi and later connected via USB

### Detailed analysis:
IsAppInReconnectMode acquires the reregister_wait_list_lock_ and also OnDeviceSwitchingStart/OnDeviceSwitchingFinish acquires the same reregister_wait_list_lock_.
The situation posible when threadX implicitly acquire reregister_wait_list_lock_ via calling IsAppInReconnectMode during responce senging from RequestController::onTimer and waiting_for_response_.
The threadY can call OnDeviceSwitchingStart/OnDeviceSwitchingFinish at the same time and acquire reregister_wait_list_lock_ and waiting for opportunity to acquire waiting_for_response_.

##### Reproduction Steps
2. iPhone is connected to the SYNC via BT over Wifi
3. SyncProxyTester(Navigation/Projection) App is registered to SYNC
4. Select Apps from Status/Interaction bar and Select SyncProxyTester app
5. From iPhone select Menu -> AV Streaming -> Start Video
6. Connect iPhone via USB
7. Disconnect USB, and wait for all apps to re-register.
8. Connect iPhone via USB again.
9. Observe the HMI behavior.

##### Expected Behavior
Apps should register to SYNC

##### Observed Behavior
Apps doesnot register to SYNC.
SDL is crashing

#### Occurrence: 
10/10

#### OS & Version Information
* OS/Version : SYNC4

mkorniichuk:
Fixes #3158 

### Risk

### Summary
The current fix is about to reduce the scope of **reregister_wait_list_lock_ptr_** usage. Because this may lead to a deadlock when during transport switching SDL has to send a message either to the Mobile app or HMI.

Example: If transport switching is timed out, UnregisterApplication will take place. Within UnregisterApplications ManageHMICommand or ManageMobileCommand of rpc_service will be triggered. And these handlers have calls of functions within application_manager with acquiring the same **reregister_wait_list_lock_ptr_**.

### CLA

UPD: 21.01.2020

### DIFF
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -1575,12 +1575,13 @@ void ApplicationManagerImpl::OnDeviceSwitchingStart(
     const connection_handler::Device& device_from,
     const connection_handler::Device& device_to) {
   LOG4CXX_AUTO_TRACE(logger_);
+  ReregisterWaitList wait_list;
   {
     auto apps_data_accessor = applications();
 
     std::copy_if(apps_data_accessor.GetData().begin(),
                  apps_data_accessor.GetData().end(),
-                 std::back_inserter(reregister_wait_list_),
+                 std::back_inserter(wait_list),
                  std::bind1st(std::ptr_fun(&device_id_comparator),
                               device_from.mac_address()));
   }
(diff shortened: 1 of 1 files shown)


mkorniichuk:
@LuxoftAKutsan from my perspective this fix is not about the Unit test since it requires a lot of real components eventually it won't be clear what exactly it would tests. Good point about ATF test but it requires much more time to create one.

mkorniichuk:
@theresalech please, be informed current PR is ready for review.

AStasiuk:
@theresalech could you please tell me when it's planned to have PM review?

AKalinich-Luxoft:
@iCollin transport switching is not related to applications itself, it is mostly related to device. So when device switches the transport, `OnDeviceSwitchingFinish` will unregister all apps registered from that device which was not reregistered.
If your question is about two devices which are trying to switch transport at the same time - it depends on environment as SDL receives signal to initiate switching from some external component outside of SDL. If this external component handles this case and does not send switching signal until switching for another device is finished, then we are fine. Otherwise, it might be a bug as one of the devices may have less time for apps reconnection.

iCollin:
@AKalinich-Luxoft Thank you for your explanation. I meant devices when I said apps the first time in my question. Let's hope that something external is prepared for that case :)

PR #3230, opened 2020-02-03
Fix ApplicationSet reference variables are not lock-protected

### LINKED ISSUE #3149: SDL Core crash after quickly and repeatedly connect/disconnect USB cable
##### Bug Report
SDL Core crash after quickly and repeatedly connect/disconnect the USB cable

##### Preconditions
Install more than 3 SDL apps on the phone

##### Reproduction Steps
Quickly and repeatedly connect/disconnect USB cable

##### Expected Behavior
SDL core should not crash

##### Observed Behavior
SDL Core crash after quickly and repeatedly connect/disconnect the USB cable.

The reason seems to be that AM Pool is not thread-safe, it is possible to use the Object just removed when users quickly and repeatedly connect/disconnect the USB cable. 
Please refer to the attached logs for further details.

`src/components/application_manager/src/policies/policy_handler.cc`

* iOS Version:10.2
* SDL iOS Version:6.3.0
* Android Version: Android 9
* SDL Android Version: 4.9.0
* Testing Against: SDL Core 5.1.3 + HMI running on our development environment.

##### Occurrence
10%

##### Recovery Method
Acc off/on

Logs:logs.txt
[logs.txt](https://github.com/smartdevicelink/sdl_core/files/3903624/logs.txt)

zhouxin627:
Fixes #3149 

### Risk

### Testing Plan
Covered by Unit tests.

### Summary
**Reason**
When App registered and another App unregistered, multiple threads access the same variable: application.
`AM FromMobile` thread free `applicationPtr`, and `AM Pool` thread uses the same `applicationPtr` which had freed by `AM FromMobile` thread.

**Solution**
`ApplicationSet` reference variables are not lock-protected, change it to a local variable.

**Sequence**
image

### CLA

### DIFF
--- a/src/components/application_manager/src/policies/policy_handler.cc
+++ b/src/components/application_manager/src/policies/policy_handler.cc
@@ -426,7 +426,8 @@ bool PolicyHandler::ClearUserConsent() {
 uint32_t PolicyHandler::GetAppIdForSending() const {
   LOG4CXX_AUTO_TRACE(logger_);
   POLICY_LIB_CHECK(0);
-  const ApplicationSet& accessor =
+  // fix ApplicationSet access crash
+  const ApplicationSet accessor =
       application_manager_.applications().GetData();
 
   HMILevelPredicate has_none_level(mobile_api::HMILevel::HMI_NONE);
--- a/src/components/application_manager/src/policies/policy_handler.cc
+++ b/src/components/application_manager/src/policies/policy_handler.cc
@@ -427,8 +427,7 @@ uint32_t PolicyHandler::GetAppIdForSending() const {
   LOG4CXX_AUTO_TRACE(logger_);
   POLICY_LIB_CHECK(0);
   // fix ApplicationSet access crash
-  const ApplicationSet accessor =
-      application_manager_.applications().GetData();
+  const ApplicationSet accessor = application_manager_.applications().GetData();
 
   HMILevelPredicate has_none_level(mobile_api::HMILevel::HMI_NONE);
   Applications apps_without_none_level;


zhouxin627:
@theresalech please, be informed current PR is ready for review.

JackLivio:
@zhouxin627 I have some questions about the fix.

1.
Code similar to what you have changed exists in a lot of areas of the project:

```
const auto& accessor = application_manager_.applications().GetData();
```

Do you think those need to be changed to not use a reference to the ApplicationSet?

2. When reviewing the DataAccessor class, it appears that the ApplicationSet provided by the DataAccessor is synchronized by the lock `applications_list_lock_ptr_`.

Also all modifications to the applications_ variable are protected by the same lock variable that is passed to the DataAccessor. 

Is there something wrong with the locking system itself? Using a reference to the applications().getData() should theoretically be safe.

zhouxin627:
Hi @JackLivio,

1. I'm sorry. We only investigated this issue and modified `GetAppIdForSending()`.
I think one part of the similar code may not involve multiple threads access, the other part may need to be changed. We didn't check them out one by one since the scope and impact are large and not directly related to the issue. 

2. `ApplicationSet` is obtained through the `DataAccessor`, and `DataAccessor` contains the lock `applications_list_lock_ptr_`.
In theory, `applications_list_lock_ptr_` should protect `ApplicationSharedPtr` from being destructed during use.
But the way of `const ApplicationSet& accessor = application_manager_.applications().GetData();`
will cause the `DataAccessor` to be destructed after the statement is executed, then release the lock `applications_list_lock_ptr_`.
The reference of `ApplicationSet` is not protected by a lock, 
so it is possible for multiple threads to read and write to the `ApplicationSharedPtr` and cause a crash.

JackLivio:
@zhouxin627 Thank you very much for the responses.

PR #3244, opened 2020-02-15
Add function UnwrapAppPolicies to CacheManager::Init() method

### LINKED ISSUE #2966: SDL incorectly assigns default permissions for an app if they were defined in preloaded
### Bug Report
SDL incorectly assigns default permissions for an app if they were defined in preloaded

##### Reproduction Steps
1. Update SDL preloaded file and add default permissions for some app. E.g.:
```
"app_policies": 
{
  "app":"default",
  "default": {
    ...
    "groups": ["Base-4"],
    ...
  },
  ...
}
```
2. Register app
3. Check `OnHMIStatus` and `OnPermissionsChange` are sent to app

##### Expected Behavior
`OnHMIStatus` and `OnPermissionsChange` notifications are sent to app
Since these are in `Base-4` func. group and allowed at `NONE` HMI level

##### Observed Behavior
`OnHMIStatus` and `OnPermissionsChange` notifications are **not** sent to app

* SDL Core Version: https://github.com/smartdevicelink/sdl_core/tree/develop (18ca3b7)
* ATF: https://github.com/smartdevicelink/sdl_atf/tree/develop

[2019-07-18_11-24-16.tar.gz](https://github.com/smartdevicelink/sdl_core/files/3407264/2019-07-18_11-24-16.tar.gz)
[002_default_app_policies_issue.zip](https://github.com/smartdevicelink/sdl_core/files/4230367/002_default_app_policies_issue.zip)

ychernysheva:
Fixes #2966

### Risk
This PR makes [no] API changes.

### Testing Plan
Covered within existing unit tests and ATF tests.

### Summary
This Pull Request is intended to fix issue with default permissions for applications, which have already presented in PT preloaded file. Main cause of this issue is absence of function UnwrapAppPolicies in CacheManager::Init() method, as result of which section with permissions for existed app wasn't mapped with default parameters (if PTU didn't happen, because earlier function UnwrapAppPolicies was used only during PTU).

### CLA

### DIFF
--- a/src/components/policy/policy_external/src/cache_manager.cc
+++ b/src/components/policy/policy_external/src/cache_manager.cc
@@ -2629,10 +2629,7 @@ bool CacheManager::Init(const std::string& file_name,
       LOG4CXX_DEBUG(logger_,
                     "Check if snapshot valid: " << std::boolalpha << result);
 
-      bool unwrap_result =
-          UnwrapAppPolicies(pt_->policy_table.app_policies_section.apps);
-      result &= unwrap_result;
-      if (!unwrap_result) {
+      if (!UnwrapAppPolicies(pt_->policy_table.app_policies_section.apps)) {
         LOG4CXX_ERROR(logger_, "Cannot unwrap application policies");
       }
       if (result) {
--- a/src/components/policy/policy_external/src/cache_manager.cc
+++ b/src/components/policy/policy_external/src/cache_manager.cc
@@ -2629,6 +2629,12 @@ bool CacheManager::Init(const std::string& file_name,
       LOG4CXX_DEBUG(logger_,
                     "Check if snapshot valid: " << std::boolalpha << result);
 
+      result &= UnwrapAppPolicies(pt_->policy_table.app_policies_section.apps);
+      if (!result) {
+        LOG4CXX_ERROR(logger_, "Cannot unwrap application policies");
+        return result;
+      }
+
       if (result) {
         backup_->UpdateDBVersion();
         Backup();
--- a/src/components/policy/policy_external/src/cache_manager.cc
+++ b/src/components/policy/policy_external/src/cache_manager.cc
@@ -2629,10 +2629,11 @@ bool CacheManager::Init(const std::string& file_name,
       LOG4CXX_DEBUG(logger_,
                     "Check if snapshot valid: " << std::boolalpha << result);
 
-      if (!UnwrapAppPolicies(pt_->policy_table.app_policies_section.apps)) {
+      bool unwrap_result = UnwrapAppPolicies(pt_->policy_table.app_policies_section.apps);
+      result &= unwrap_result;
+      if (!unwrap_result){
         LOG4CXX_ERROR(logger_, "Cannot unwrap application policies");
       }
-
       if (result) {
         backup_->UpdateDBVersion();
         Backup();
--- a/src/components/policy/policy_external/src/cache_manager.cc
+++ b/src/components/policy/policy_external/src/cache_manager.cc
@@ -2632,7 +2632,6 @@ bool CacheManager::Init(const std::string& file_name,
       result &= UnwrapAppPolicies(pt_->policy_table.app_policies_section.apps);
       if (!result) {
         LOG4CXX_ERROR(logger_, "Cannot unwrap application policies");
-        return result;
       }
 
       if (result) {
--- a/src/components/policy/policy_external/src/cache_manager.cc
+++ b/src/components/policy/policy_external/src/cache_manager.cc
@@ -2629,8 +2629,7 @@ bool CacheManager::Init(const std::string& file_name,
       LOG4CXX_DEBUG(logger_,
                     "Check if snapshot valid: " << std::boolalpha << result);
 
-      result &= UnwrapAppPolicies(pt_->policy_table.app_policies_section.apps);
-      if (!result) {
+      if (!UnwrapAppPolicies(pt_->policy_table.app_policies_section.apps)) {
         LOG4CXX_ERROR(logger_, "Cannot unwrap application policies");
       }
 
--- a/src/components/policy/policy_external/src/cache_manager.cc
+++ b/src/components/policy/policy_external/src/cache_manager.cc
@@ -2629,9 +2629,10 @@ bool CacheManager::Init(const std::string& file_name,
       LOG4CXX_DEBUG(logger_,
                     "Check if snapshot valid: " << std::boolalpha << result);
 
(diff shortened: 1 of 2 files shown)


src/components/policy/policy_regular/src/cache_manager.cc
```diff
@@ -1783,6 +1784,7 @@ bool CacheManager::Init(const std::string& file_name,
       LOG4CXX_INFO(logger_, "Policy Table was inited successfully");
 
       result = LoadFromFile(file_name, *pt_);
+      UnwrapAppPolicies(pt_->policy_table.app_policies_section.apps);
```
LuxoftAKutsan: `UnwrapAppPolicies` is bool so it may fail. 
Good practice - to check return results. 

I would propose putting `UnwrapAppPolicies` after `snapshot->is_valid();`. 
And make the `result` depending from `UnwrapAppPolicies` : `result &= UnwrapAppPolicies(...`

Also, looks like we need to add additional Unit test for Init Function, that will check that is actually does unwrapping policies.

ychernysheva: 1. Updated test script is attached to original issue.
2. Unfortunately, class CacheManager isn't covered by unit tests at all, this problem is reflected in issue # 2449.
3. Fix for external flow is added in commit 4c96187dc1b2f5f8bab7b820dfdf9d5713ba136f.
4. Full regression is passed successfully.

src/components/policy/policy_external/src/cache_manager.cc
```diff
@@ -2629,6 +2629,12 @@ bool CacheManager::Init(const std::string& file_name,
       LOG4CXX_DEBUG(logger_,
                     "Check if snapshot valid: " << std::boolalpha << result);
 
+      result &= UnwrapAppPolicies(pt_->policy_table.app_policies_section.apps);
```
iCollin: this doesn't make sense, I don't think you can use the same result variable here as it is used for later logic.

I would just put this function call inside the if condition right below.

src/components/policy/policy_external/src/cache_manager.cc
```diff
@@ -2629,6 +2629,12 @@ bool CacheManager::Init(const std::string& file_name,
...
+      if (!result) {
+        LOG4CXX_ERROR(logger_, "Cannot unwrap application policies");
+        return result;
```
iCollin: Should we really return if we cannot unwrap app policies?

What if the snapshot was invalid? 
Do we want to skip reporting the snapshot errors and doing `ex_backup_->RemoveDB();`?

LuxoftAKutsan: If Unwrap Policies was unsuccessful, it means, that preloaded_pt is wrong, and we should not start sdl.  The second option it to make UnwrapAppPolicies function void, and do not care about it's result.

src/components/policy/policy_external/src/cache_manager.cc
LuxoftAKutsan: Unwrapping policies is preparation step for policies data, In case if it is unsuccessful for some reason we should understand it as wrong policies data, and follow corresponding strategy. I agree that return on `UnwrapAppPolicies = false` is redundant and probably even malicious. I would keep appropriate error message in the `if` and then follow to `else` statement for result checking.

iCollin: That sounds fine to me, and so does the return void idea because there is also already a log in the fail case of `UnwrapAppPolicies`. Personally I would leave the function as bool and also log here.

I just don't think we should change the result variable as it would create a case where we lose functionality.

ychernysheva: Thank you for review, I've implemented the proposed changes in commits 4b62a2c03b898dd096d50b38ea1dc15227b973c9 and 7412f0c07518e6e21d7bea98c813cad7c28d3330.

ychernysheva: Additionally I've made some changes for external flow in commit 6f62d19ed04e987f41e0c24e66c1ec388955a9f9

iCollin: @ychernysheva thank you for your updates.

@LuxoftAKutsan I apologize, I did not understand your message correctly when I last replied. I see now that you did propose to do `result &=` with the return value, but I am still not convinced this is a good idea.

A policy table could still exist with an app policy: `"default": null,`, but I don't think it is okay to add a requirement that this key exists at this time. I am weary of changing behavior like this just because, no matter how unlikely, it is possible someone is using a policy table without the default key. I doubt anyone is doing this, but I don't see any advantage to taking the risk of breaking their flow.

If there is anything that would be broken by having the "default" key missing in app policies, please let me know and I will reconsider but otherwise this requirement should not be added.

LuxoftAKutsan: @iCollin  there is no problem if default is absent in policy table. 
But if default is absent and there are some application with 
`app: "default"`. that is the issue, and UnwrapAppPolicies should return false and SDL. Such policy table is invalid. but `is_valid` won't signal about it.

ychernysheva: @iCollin I've changed code in external flow in commit b877b37e3dd1b1fe337530e769d64aba04273069, and now it's similar to fix in proprietary flow. I hope that now it's correct from your perspective? If it's not, please request additional changes, and I'll be happy to implement them.

src/components/policy/policy_external/src/cache_manager.cc
```diff
@@ -2629,6 +2629,9 @@ bool CacheManager::Init(const std::string& file_name,
       LOG4CXX_DEBUG(logger_,
                     "Check if snapshot valid: " << std::boolalpha << result);
 
+      if (!UnwrapAppPolicies(pt_->policy_table.app_policies_section.apps)) {
+        LOG4CXX_ERROR(logger_, "Cannot unwrap application policies");
+      }
```
iCollin: Please add a newline after this if block.

src/components/policy/policy_external/src/cache_manager.cc
ychernysheva: @iCollin And I also want to propose squashing this commits before merging, because history of this review isn't so important. I think that it may be good idea, isn't it?

src/components/policy/policy_external/src/cache_manager.cc
ychernysheva: Newline was added in commit 9ddff2ba34b270929df979e63ef50043e363490a

PR #3277, opened 2020-03-05
Fix SDLCore failed to retry updating the Policy Table

### LINKED ISSUE #3232: SDLCore failed to retry updating the Policy Table
### Bug Report
SDLCore failed to retry updating the Policy Table

##### Preconditions
Connect the head unit to any Android/iOS phone

##### Reproduction Steps
1. The first time of updating Policy Table is failed that caused by network problems
2. Fix the network problems soon.

##### Expected Behavior
SDLCore retries updating Policy Table is successful.

##### Observed Behavior
SDLCore retries updating Policy Table is failed.
It sends the correct URL to the phone for the first time of updating the Policy Table.
But when a retry is requested, it sends an empty URL to the phone,
so it certainly failed to update the Policy Table.

* SDL Core Version: SDL Core 6.0.1
* Testing Against:SDL Core 6.0.1 + HMI running on our development environment.

##### Occurrence:
100%

##### Recovery Method:
None

zhouxin627:
Fixes #3232 

### Risk

### Testing Plan
Covered by Unit tests.

### Summary
**Reason**
When the application receives `OnSystemRequest` RPC without URL parameters, it can not download the file from the server. As a result, PT updates failed.

**Solution**
When SDL is built in default policy mode (PROPRIETARY), there are two types of PTU, `DefaultIteration` and `RetryIteration`. when PTU type is `DefaultIteration`, SDL calls `SendPolicyUpdate` function, which will send `OnSystemRequest` RPC with parameters `URL` to application.
when PTU type is `RetryIteration`, SDL calls `SendPolicySnapshotNotification` function, which will send `OnSystemRequest` RPC without parameters `URL` to application.

The solution is that abandon the differential processing of PTU when PTU is triggered SDL always call `SendPolicyUpdate` function no matter what type of PTU.
it can ensure that `OnSystemRequset` RPC received by application contains the URL parameter.

**Sequence**
image

### CLA


zhouxin627:
Hi @theresalech please, be informed current PR is ready for review.

theresalech:
Thanks @zhouxin627 !  We will review as time allows before the 6.1 Release.

src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -1544,21 +1544,11 @@ void PolicyHandler::OnSnapshotCreated(const BinaryMessage& pt_string,
...
-    }
-
-  } else {
-    MessageHelper::SendPolicyUpdate(
-        policy_snapshot_full_path,
-        TimeoutExchangeSec(),
-        policy_manager_->RetrySequenceDelaysSeconds(),
-        application_manager_);
-  }
+  MessageHelper::SendPolicyUpdate(
```
LuxoftAKutsan: From PR description I can see that changes `Covered by Unit tests.` 

But I do not see any changes in Unit tests. 
If before changes all unit tests were passed, and after changes it still passed I would add additional unit test that should fail without appropriate changes, but pass with it.

zhouxin627: Hello @LuxoftAKutsan sorry, I have not done any changes in Unit tests. Please add additional unit test as you need.

jacobkeeler:
@zhouxin627 It seems like this fix breaks existing flows for PROPRIETARY policy mode. I've just submitted a potential solution in #3329

zhouxin627:
@jacobkeeler Thank you. We are still confirming #3329 and I'll give you feedback tomorrow.

zhouxin627:
Hi @jacobkeeler, The modification method of #3329 is that get the URL from the `policytable` when retry. But if HMI wants to do some custom processing for URL(At present, our style is customized according to the car model and other information), this method cannot be done in the retry process.

If Livio wants to retry without HMI notification. Then, we recommend caching the URL of the previous HMI notification. When retry is performed, the cached URL will be used for this processing.

jacobkeeler:
@zhouxin627 is there a reason that you aren't using `EXTERNAL_PROPRIETARY` mode? In that mode, the HMI is responsible for controlling the retry sequence as well as selecting the URLs used by the phone.

Beyond that, you really shouldn't be modifying these URLs manually to my knowledge. You should be able to customize these URLs directly through the policy table rather than modifying them manually. To my knowledge, this is the purpose of the Policy Table snapshot that is sent to the policy server, it contains information which can be used to customize the contents of the policy table specifically for the vehicle.

Since Core is responsible for the retry sequence for `PROPRIETARY` mode, it needs to be responsible through cycling through the available URLs as well.

zx5656:
@jacobkeeler Thank you for your explanation. I understood. By the way, why the implementation of proprietary mode in sdlcore 5.0 is different from that in sdlcore 5.1 and later(Before sdlcore 5.0, HMI will be notified when retrying), do you know which evolution proposal is responsible for this part of the change?

zhouxin627:
@jacobkeeler Thank you for your explanation. I understood. By the way, why the implementation of proprietary mode in sdlcore 5.0 is different from that in sdlcore 5.1 and later(Before sdlcore 5.0, HMI will be notified when retrying), do you know which evolution proposal is responsible for this part of the change?

jacobkeeler:
From what I can see, it was implemented in https://github.com/smartdevicelink/sdl_core/pull/2921, which was tied to [SDL 0211](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0211-ServiceStatusUpdateToHMI.md). It doesn't look like the proposal directly related to this flow change, but unfortunately there isn't any documentation to indicate the actual retry flow for each policy mode.

The main reasons that I see for this change are
1. The `BC.PolicyUpdate` message contains retry information for EXTERNAL_PROPRIETARY policy mode, which indicates that this message is meant to start a full PTU cycle rather than just being used for a retry.
2. With `BC.PolicyUpdate`, there is no way for the HMI to differentiate whether or not it is a retry or a new sequence.

For these reasons, it doesn't make sense for `BC.PolicyUpdate` to be sent on each retry, and this follows the behavior of EXTERNAL_PROPRIETARY policy mode as well.

jacobkeeler:
@zhouxin627 After some discussion with @JackLivio, we decided the best approach to this problem would probably be to cache the URL for reuse _only_ if it is provided by the HMI. This would mean that only one URL is used for retries in that case, but it would make it so existing implementations would not be broken, and the proper flow could still be performed if a URL is not provided. Core will perform the normal URL cycling if no URL is provided by the HMI.

This change has been pushed to #3329

zhouxin627:
@jacobkeeler Great!  I'll test it and give you a feedback ASAP.

jacobkeeler:
Closing in favor of #3329

PR #3280, opened 2020-03-09
Fix/SDL does not send BC.PolicyUpdate to HMI if no apps are registered

### LINKED ISSUE #3279: SDL does not send BC.PolicyUpdate to HMI if no apps are registered
### Bug Report
Currently, if there are no registered apps, triggering a PTU results in the HMI sending a `BasicCommunication.OnSystemRequest` with `fileName: null`. 

The PTS filename is never set in the HMI because SDL does not send a `BasicCommunication.PolicyUpdate` request(which contains the PTS path) to the HMI. This prevents us from being able to use the HMI vehicle modem to handle the PTU if no apps are registered.

##### Reproduction Steps
Using the sdl hmi with external proprietary checked:
1. Build core with external proprietary
2. Start the policy server
3. Modify the `0x07` endpoint in the PT
4. Register app, successful PTU should occur
5. Unregister app
6. Change the kilometers to 5000 to trigger a PTU

##### Expected Behavior
1. SDL sends a `BasicCommunication.PolicyUpdate` request to the HMI.
2. The HMI sends a `BasicCommunication.OnSystemRequest(PROPRIETARY)` notification to SDL with a valid PTS `fileName`.

##### Observed Behavior
1. SDL does NOT send a `BasicCommunication.PolicyUpdate` request to the HMI.
2. The HMI sends a `BasicCommunication.OnSystemRequest(PROPRIETARY)` notification to SDL with `fileName: null`.

* OS/Version: Ubuntu 18.04
* SDL Core Version: origin/develop (https://github.com/smartdevicelink/sdl_core/commit/94f92bb7eb440c0ee878896b6281f38009d64094)
* Testing Against: Test Suite

ShobhitAd:
Fixes #3279 

### Risk

### Testing Plan
Test Suite app and ATF

### Summary
Temporary fix: Change the `PolicyHandler::CanUpdate` function to return true even when no apps are registered

### CLA

### DIFF
--- a/src/components/application_manager/src/policies/policy_handler.cc
+++ b/src/components/application_manager/src/policies/policy_handler.cc
@@ -1953,7 +1953,7 @@ void PolicyHandler::OnPTUFinished(const bool ptu_result) {
 }
 
 bool PolicyHandler::CanUpdate() {
-  return 0 != GetAppIdForSending();
+  return true;
 }
 
 void PolicyHandler::RemoveDevice(const std::string& device_id) {
--- a/src/components/application_manager/src/policies/policy_handler.cc
+++ b/src/components/application_manager/src/policies/policy_handler.cc
@@ -1953,7 +1953,7 @@ void PolicyHandler::OnPTUFinished(const bool ptu_result) {
 }
 
 bool PolicyHandler::CanUpdate() {
-  return true;
+  return 0 != GetAppIdForSending();
 }
 
 void PolicyHandler::RemoveDevice(const std::string& device_id) {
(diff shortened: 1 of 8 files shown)


src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -1953,7 +1953,7 @@ void PolicyHandler::OnPTUFinished(const bool ptu_result) {
 }
 
 bool PolicyHandler::CanUpdate() {
-  return 0 != GetAppIdForSending();
+  return true;
```
JackLivio: Instead of changing the CanUpdate() behavior to return true, I would omit calling the function where applicable.

ShobhitAd: Agreed, that seems like a much better approach. Fixed in https://github.com/smartdevicelink/sdl_core/pull/3280/commits/4bb0606f3bede5abd33b28cd1d8f89d3a9b29ecd

PR #3333, opened 2020-04-11
Fix/Processing deprecated mandatory parameter

### LINKED ISSUE #3330: SetInteriorVehicleData request is processed as invalid without deprecated mandatory parameter
### Bug Report
`SetInteriorVehicleData` request is processed as invalid without deprecated mandatory parameter

##### Reproduction Steps
1. HMI and SDL are started
2. Remote_control application with version 6.0 and upper is registered and activated
3. Mobile app requests `SetInteriorVehicleData` without id parameter.

##### Expected Behavior
According to API:
> <param name="id" type="SupportedSeat" mandatory="true" deprecated="true" since="6.0">
So `SetInteriorVehicleData` request should be processed successfully without deprecated parameter.

##### Observed Behavior
SDL rejects `SetInteriorVehicleData` request with `INVALID_DATA` resultCode because of deprecated parameter missing

* sdl_hmi branch: [develop](https://github.com/smartdevicelink/sdl_hmi/tree/develop)
* commit: e8baa2d8f3f4b43d261323e0dd2c30843509e925
* sdl_core branch: https://github.com/smartdevicelink/sdl_core/tree/6.1.0_RC

##### Logs
[SmartDeviceLinkCore_seat_id_deprecated.zip](https://github.com/smartdevicelink/sdl_core/files/4456461/SmartDeviceLinkCore_seat_id_deprecated.zip)

SNiukalov:
Fixes #3330 

### Risk

### Testing Plan
Use scripts from
[smartdevicelink/sdl_atf_test_scripts#2388](https://github.com/smartdevicelink/sdl_atf_test_scripts/pull/2388)

### Summary
Object schema item does not take into account an deprecated property when processing a mandatory parameter.
InterfaceGenerator add depereted property in all fileds if function depereted.

* Add deprecated property to compare when processing a mandatory parameter.
* Remove inheritance of the deprecated attribute from function.

### CLA

### DIFF
--- a/src/components/smart_objects/src/object_schema_item.cc
+++ b/src/components/smart_objects/src/object_schema_item.cc
@@ -126,8 +126,8 @@ errors::eType CObjectSchemaItem::validate(
 
     std::set<std::string>::const_iterator key_it = object_keys.find(key);
     if (object_keys.end() == key_it) {
-      if (correct_member && correct_member->mIsMandatory == true &&
-          correct_member->mIsRemoved == false) {
+      if (correct_member && correct_member->mIsMandatory &&
+          !correct_member->mIsDeprecated && !correct_member->mIsRemoved) {
         std::string validation_info = "Missing mandatory parameter: " + key;
         report__->set_validation_info(validation_info);
         return errors::MISSING_MANDATORY_PARAMETER;
--- a/tools/InterfaceGenerator/generator/generators/SmartFactoryBase.py
+++ b/tools/InterfaceGenerator/generator/generators/SmartFactoryBase.py
@@ -1156,7 +1156,7 @@ class CodeGenerator(object):
                     schema_items_decl=self._gen_schema_items_decls(
                         function.params.values()),
                     schema_item_fill=self._gen_schema_items_fill(
-                        function.params.values(), function.since, function.until, function.deprecated, function.removed),
+                        function.params.values(), function.since, function.until, function.removed),
                     schema_params_fill=self._gen_schema_params_fill(
                         function.message_type.name)),
                 1))


src/components/smart_objects/src/object_schema_item.cc
```diff
@@ -127,6 +127,7 @@ errors::eType CObjectSchemaItem::validate(
     std::set<std::string>::const_iterator key_it = object_keys.find(key);
     if (object_keys.end() == key_it) {
       if (correct_member && correct_member->mIsMandatory == true &&
```
AByzhynar, line 129: @SNiukalov Can we change it to more short & simple line: 
```cpp
if (correct_member && correct_member->mIsMandatory &&
    !correct_member->mIsDeprecated &&
    !correct_member->mIsRemoved)
```

SNiukalov: @AByzhynar 
Thanks, fixed in:https://github.com/smartdevicelink/sdl_core/pull/3333/commits/8340fe94900632930a4438cda26dea613e9f6282

SNiukalov:
@LuxoftAKutsan 
Thanks, summary updated.

tools/InterfaceGenerator/generator/generators/SmartFactoryBase.py
```diff
@@ -1156,7 +1156,7 @@ def _gen_function_impl(self, function, namespace, class_name):
                     schema_items_decl=self._gen_schema_items_decls(
                         function.params.values()),
                     schema_item_fill=self._gen_schema_items_fill(
-                        function.params.values(), function.since, function.until, function.deprecated, function.removed),
```
AByzhynar, line 1159: @SNiukalov Please change history spreading to all RPC parameters together with deprecation. 
SO if the entire RPC is deprecated from <version> and <History> is present - 
All RPC params should be deprecated considering <History>.

JackLivio:
@SNiukalov @AByzhynar 

Thank you for submitting this fix but unfortunately the behavior reported in the issue is expected behavior. Although the SeatControlData ID parameter is deprecated in RPC spec version 6.0, it is still mandatory and must be included. On the next major version change the parameter will be removed from the spec.

A couple of rules we have been following since the implementation of SDL Core Mobile Versioning:

1. All parameters must go through a period of deprecation before being removed.
2. Changing a parameter mandatory status from true/false is a breaking change and requires a major version change. 
3. Deprecating a mandatory parameter still requires that it is included in the request.

JackLivio:
Closing PR since original Issue was closed. https://github.com/smartdevicelink/sdl_core/issues/3330#issuecomment-617016250

GetmanetsIrina:
@JackLivio, thanks for your detailed explanation.
I have one question about this info: 
> On the next major version change the parameter will be removed from the spec.
Is this a general logic for all deprecated parameters and mobile API should contain the deprecated parameters only with the current major version?

PR #3342, opened 2020-04-21
Fix capability window update when window_id is missing

JackLivio:

### Risk

### Testing Plan
Send a setDisplayLayout with LARGE_GRAPHIC_ONLY using the generic HMI. Confirm that there are no mainfield text capabilities or softbutton capabilities included in the OnsystemCapabilityUpdated notification to mobile.

### Summary
If the window_id was missing from the display updates, `has_window_id()` would return false repeatedly so everytime the main window was updated, the window_capabilities array would grow in size while holding onto old versions of the main window capability. 

When display_capability(window_id) was called by the mobile notification run() method, the outdated main window capabilities were sent to the app instead of the updated capabilities. 

### CLA

### DIFF
--- a/src/components/application_manager/src/application_data_impl.cc
+++ b/src/components/application_manager/src/application_data_impl.cc
@@ -588,7 +588,7 @@ void DynamicApplicationDataImpl::set_display_capabilities(
   };
 
   for (uint32_t i = 0; i < incoming_window_capabilities.length(); ++i) {
-    auto window_id = 0;
+    int64_t window_id = 0;
     if (incoming_window_capabilities[i].keyExists(strings::window_id)) {
       window_id = incoming_window_capabilities[i][strings::window_id].asInt();
     }
--- a/src/components/application_manager/src/application_data_impl.cc
+++ b/src/components/application_manager/src/application_data_impl.cc
@@ -568,9 +568,11 @@ void DynamicApplicationDataImpl::set_display_capabilities(
   display_capabilities_.reset(
       new smart_objects::SmartObject(display_capabilities));
 
-  auto has_window_id = [&tmp_window_capabilities](const WindowID window_id) {
+  auto has_window_id = [&tmp_window_capabilities](const WindowID window_id,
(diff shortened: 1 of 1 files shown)


src/components/application_manager/src/application_data_impl.cc
```diff
@@ -568,30 +568,37 @@ void DynamicApplicationDataImpl::set_display_capabilities(
...
+      found_index = -1;
       return false;
     }
 
     for (auto element : *tmp_window_capabilities_arr) {
       if (element.keyExists(strings::window_id)) {
         if (window_id == element[strings::window_id].asInt())
           return true;
-      } else if (window_id == 0) {
+      } else if (window_id == 0 || window_id == -1) {
```
jacobkeeler, line 583: I assume that `-1` is what is returned by `incoming_window_capabilities[i][strings::window_id].asInt()` if the key isn't present? I think it would be better to check `keyExists` rather than blindly using `asInt()` if possible.
Maybe:
```
auto window_id = 0;
if (incoming_window_capabilities[i].hasKey(strings::window_id)) {
  window_id = incoming_window_capabilities[i][strings::window_id].asInt();
}
...
```

src/components/application_manager/src/application_data_impl.cc
```diff
@@ -568,37 +568,37 @@ void DynamicApplicationDataImpl::set_display_capabilities(
...
 
   for (uint32_t i = 0; i < incoming_window_capabilities.length(); ++i) {
-    const auto window_id =
-        incoming_window_capabilities[i][strings::window_id].asInt();
-    int found_index = 0;
-    if (!has_window_id(window_id, found_index)) {
-      tmp_window_capabilities[tmp_window_capabilities.length()] =
-          incoming_window_capabilities[i];
-    } else if (found_index >= 0) {
+    auto window_id = 0;
```
iCollin, line 591:
```suggestion
    int64_t window_id = 0;
```

PR #3350, opened 2020-04-30
do not retry ptu if no apps are connected

iCollin:

### Risk

### Testing Plan
Manually, as well as ATF

### Summary
If Core cannot find an app to which it should send a PTU retry, do not attempt to send a retry.

### CLA

### DIFF
--- a/src/components/application_manager/src/policies/policy_handler.cc
+++ b/src/components/application_manager/src/policies/policy_handler.cc
@@ -1621,7 +1621,7 @@ void PolicyHandler::OnSnapshotCreated(const BinaryMessage& pt_string,
     uint32_t app_id_for_sending = 0;
     const std::string& url =
         GetNextUpdateUrl(PTUIterationType::RetryIteration, app_id_for_sending);
-    if (0 != url.length()) {
+    if (0 != url.length() && 0 != app_id_for_sending) {
       MessageHelper::SendPolicySnapshotNotification(
           app_id_for_sending, pt_string, url, application_manager_);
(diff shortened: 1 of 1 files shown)


jacobkeeler:
Should target a release branch `release/6.1.1`

src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -1621,7 +1621,7 @@ void PolicyHandler::OnSnapshotCreated(const BinaryMessage& pt_string,
     uint32_t app_id_for_sending = 0;
     const std::string& url =
         GetNextUpdateUrl(PTUIterationType::RetryIteration, app_id_for_sending);
-    if (0 != url.length()) {
+    if (0 != url.length() && 0 != app_id_for_sending) {
```
jacobkeeler, line 1624: This check might be good for safety, but I think a check should also be in the GetNextUpdateUrl function:
```
  app_id = ChoosePTUApplication(iteration_type);
+ if (app_id == 0) {
+   return std::string();
+ }

  // Use cached URL for retries if it was provided by the HMI
  if (PTUIterationType::RetryIteration == iteration_type &&
      !retry_update_url_.empty()) {
    return retry_update_url_;
  }
```

iCollin: fix adopted in https://github.com/smartdevicelink/sdl_core/pull/3350/commits/5198341c7e12055746c434093b3d0c3ddf79e660

PR #3363, opened 2020-05-12
Fix/Add Startup PTU trigger check

### LINKED ISSUE #3279: SDL does not send BC.PolicyUpdate to HMI if no apps are registered

ShobhitAd:
Fixes #3279 

### Risk

### Testing Plan
ATF Test and manual tests

### Summary
Currently, the IgnitionCycles and Days PTU triggers are not checked in the regular policy mode when SDL starts.

### CLA

### DIFF
--- a/src/components/application_manager/src/policies/policy_handler.cc
+++ b/src/components/application_manager/src/policies/policy_handler.cc
@@ -787,9 +787,9 @@ void PolicyHandler::OnSystemRequestReceived() const {
   policy_manager_->ResetTimeout();
 }
 
-void PolicyHandler::TriggerPTUIfRequired() {
+void PolicyHandler::TriggerPTUOnStartupIfRequired() {
 #ifndef EXTERNAL_PROPRIETARY_MODE
-  policy_manager_->TriggerPTUIfRequired();
+  policy_manager_->TriggerPTUOnStartupIfRequired();
 #endif
 }
 
--- a/src/components/application_manager/src/policies/policy_handler.cc
+++ b/src/components/application_manager/src/policies/policy_handler.cc
@@ -787,6 +787,12 @@ void PolicyHandler::OnSystemRequestReceived() const {
   policy_manager_->ResetTimeout();
 }
 
+void PolicyHandler::TriggerPTUIfRequired() {
+#ifndef EXTERNAL_PROPRIETARY_MODE
+  policy_manager_->TriggerPTUIfRequired();
+#endif
+}
+
 void PolicyHandler::GetRegisteredLinks(
     std::map<std::string, std::string>& out_links) const {
   DataAccessor<ApplicationSet> accessor = application_manager_.applications();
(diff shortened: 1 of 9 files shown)


src/components/application_manager/src/policies/policy_handler.cc
```diff
@@ -787,6 +787,12 @@ void PolicyHandler::OnSystemRequestReceived() const {
   policy_manager_->ResetTimeout();
 }
 
+void PolicyHandler::TriggerPTUIfRequired() {
+#ifndef EXTERNAL_PROPRIETARY_MODE
+  policy_manager_->TriggerPTUIfRequired();
```
iCollin, line 792: I think the policy mode check should go where the call is originally triggered, at the end of `ApplicationManagerImpl::OnHMIStartedCooperation`

ShobhitAd: The `TriggerPTUIfRequired` function is not implemented for external policies. In this case, it would be fine if I have the `EXTERNAL_POLICY` flag check in OnHMIStartedCooperation but If I had to use it somewhere else I’d have to add the policy flag check for each call. It might be neater for this to be handled in the policy handler. 

For now I think we could  add a note in the policy_handler function brief stating that its only implemented for regular policies at the moment

ShobhitAd: Made the change in https://github.com/smartdevicelink/sdl_core/pull/3363/commits/454e5fbda47c6a7c0c14c8d8206354a33e1a3685

PR #3375, opened 2020-05-19
Fix dangerous work with app commands

### LINKED ISSUE #3170: CCB: Fix dangerous work with app commands
### Bug Report
Crash in AttachSchema SDL

### Detailed analysis:
There was related to a thread safety violation while working with application internal commands
array. This array is a thread-safe itself as it is protected with mutexes, however FindCommand()
function is returning a raw pointer to an internal array element so the external component is able to get access to this element at any time even without locking the mutex. Core crash which was found in DeleteCommand request instance is happening by exactly the same reason -
DeleteCommand is using a raw pointer to find the command, however, this command might be destroyed from another thread at this point in time. As a result, working thread can access to a destroyed object by pointer. As a quick solution, each FindCommand() can be followed
by a command accessor which prevents unexpected command destruction. Working thread will use the temporary copy of command, but not command itself. However, the better solution would be to change the design and to avoid using of the raw pointers.

##### Reproduction Steps
2. Register app
3. Add some commands and then send DeleteCommand RPC
4. Observe behavior

##### Expected Behavior
SDL crashed
[bt.txt](https://github.com/smartdevicelink/sdl_core/files/3932171/bt.txt)

##### Observed Behavior
No crash

#### OS & Version Information
* OS/Version : SYNC4

LuxoftAKutsan:
Fixes #3170

### Risk

### Testing Plan
Tested on SYNC version of SDL
For open source version will be created test script. 

### Summary
There was related to a thread safety violation
while working with application internal commands
array. This array is a thread safe itself as it is
protected with mutexes, however FindCommand()
function is returing a raw pointer to an internal
array element so the external component is able to
get access to this element at any time even without
locking the mutex.
Core crash which was found in DeleteCommand request
instance is happening by exactly the same reason -
DeleteCommand is using a raw pointer to find the command,
however this command might be destroyed from another
thread at this point of time. As a result, working
thread can access to a destroyed object by pointer.
As a quick solution, each FindCommand() was followed
by a command accessor which prevents unexpected command
destruction. Working thread will use the temporary
copy of command, but not command itself.
However, the better solution would be to change a
design and to avoid using of the raw pointers.

### Tasks Remaining:
- [x] Luxoft Internal review
- [ ] [Task 2]

### CLA


src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/delete_command_request.cc
```diff
@@ -79,9 +79,15 @@ void DeleteCommandRequest::Run() {
   const int32_t cmd_id =
       (*message_)[strings::msg_params][strings::cmd_id].asInt();
 
-  smart_objects::SmartObject* command = application->FindCommand(cmd_id);
+  smart_objects::SmartObject command;
+  {
+    smart_objects::SmartObject* command_ptr = application->FindCommand(cmd_id);
+    if (command_ptr) {
+      command = *command_ptr;
```
mvorobio, line 86: @LuxoftAKutsan Seems there is still a chance to get the command object destroyed before copying it. Is this ok?

jacobkeeler: @LuxoftAKutsan

Agreed with @mvorobio, I think the only way to make this truly safe is to add a function which returns by value instead of pointer.

theresalech:

LuxoftAKutsan:
@theresalech this work is planned. We will notify when fix with test scripts will be available.

IGapchuk:
Hi. @jacobkeeler please be informed we've updated the `FindCommand` method to return value instead of raw pointer. Also, we've updated the next methods: `FindSubMenu` and `FindChoiceSet`, because they have the same behavior - return by raw pointer instead of return by value. Please take a look at the 33e9a6116bcb895d7570596ad5899ff93f330be5. The related Unit tests were updated too. Thank you!

JackLivio:
@LuxoftAKutsan Seeing some failed atf smoke tests relating to AddCommand and AddSubmenu on the latest commit for this pr. Confirmed with @dboltovskyi.

IGapchuk:
Hi, @JackLivio ! We've merged the `smartdevicelink/develop` branch into the fix branch (2e40cc73fc5da364dbc009a1ba5fe890fc6dd0f0) and restarted full regression.

JackLivio:
@IGapchuk What were the test results after the merge of develop?

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/add_sub_menu_request.cc
```diff
@@ -85,7 +85,10 @@ void AddSubMenuRequest::Run() {
   }
 
   const int32_t menu_id = received_msg_params[strings::menu_id].asInt();
-  if (app->FindSubMenu(menu_id)) {
+
+  const auto sub_menu = app->FindSubMenu(menu_id);
+
+  if (smart_objects::SmartType_Null == sub_menu.getType()) {
```
JackLivio, line 91:
```suggestion
  if (smart_objects::SmartType_Null != sub_menu.getType()) {
```

JackLivio: Also please fix the log message below, it should say that that submenu id already exsits

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/add_sub_menu_request.cc
```diff
@@ -88,7 +88,7 @@ void AddSubMenuRequest::Run() {
 
   const auto sub_menu = app->FindSubMenu(menu_id);
 
-  if (smart_objects::SmartType_Null == sub_menu.getType()) {
+  if (smart_objects::SmartType_Null != sub_menu.getType()) {
     LOG4CXX_ERROR(logger_, "Menu with id " << menu_id << " is not found.");
```
JackLivio, line 92:
```suggestion
    LOG4CXX_ERROR(logger_, "Menu with id " << menu_id << " already exists.");
```

LuxoftAKutsan: Fixed

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/add_sub_menu_request.cc
LuxoftAKutsan: Fixed, also fixed one more fixed place 63cdee2954

LuxoftAKutsan:
Backmerged develop

PR #3380, opened 2020-05-20
Fix trying to unlock mutex not held by any thread

### LINKED ISSUE #3186: CCB: Fix trying to unlock mutex not held by any thread
### Bug Report
Data races related to MessageQueue

### Detailed analysis:
Mutex released before call Broadcast.

##### Reproduction Steps
1. TBD

##### Expected Behavior
SDL should not crash

##### Observed Behavior
SDL crashed

#### OS & Version Information
* OS/Version : SYNC4

LuxoftAKutsan:
Fixes #3186

### Risk

### Testing Plan
Regression automated testing

### Summary

Fix trying to unlock mutex not held by any thread
Helgrind considers that this issue is an error because we release mutex before call Broadcast.

- [ ] Luxoft internal review 
- [ ] Regression automated testing

### CLA


src/components/include/utils/message_queue.h
```diff
@@ -171,13 +171,11 @@ bool MessageQueue<T, Q>::IsShuttingDown() const {
...
-    if (shutting_down_) {
-      return;
-    }
-    queue_.push(element);
+  sync_primitives::AutoLock auto_lock(queue_lock_);
+  if (shutting_down_) {
+    return;
   }
+  queue_.push(element);
   queue_new_items_.Broadcast();
```
jacobkeeler, line 179: It seems to me like this change would cause a DCHECK, I'm surprised that this is supposed to fix an issue. If one thread has called `Wait`, that would mean that when the `Broadcast` is received we would reach `AssertFreeAndMarkTaken`, which will cause a `NOTREACHED` call if the lock is still held.

LuxoftAKutsan: Fix moved from SYNC4 codebase. Additional testing and description will be provided

theresalech:

LuxoftAKutsan:
@theresalech this work is planned. We will notiy when fix with test scripts will be available.

src/components/include/utils/message_queue.h
LuxoftAKutsan: `Broadcast()` will unlock the `queue_lock_` and let the process that is waiting on `Wait()` to lock the `queue_lock_`. 

In another case between `Broadcast()` from thread1 and `Wait()` from thread2 some other thread may acquire `queue_lock_` that will prevent ability for thread2 to squre

LuxoftAKutsan:
@jacobkeeler @theresalech full regression tested on this PR. Results are similar to develop

src/components/include/utils/message_queue.h
JackLivio: Confirmed logic from online resources: https://docs.oracle.com/cd/E19455-01/806-5257/sync-27/index.html

> Call pthread_cond_broadcast() under the protection of the same mutex that is used with the condition variable being signaled. Otherwise, the condition variable could be signaled between the test of the associated condition and blocking in pthread_cond_wait(), which can cause an infinite wait.

JackLivio (approved):

PR #3382, opened 2020-05-20
Fix/3178 crash on shutdown

### LINKED ISSUE #3178: CCB: SDL crash on set_delegate on shutdown
### Bug Report
SDL crash on set_delegate on shutdown 

### Detailed analysis:
The core crash shows evidence that something around set_delegate is
crashing, on the other hand ~MessageLoopThread is deleting the thread
and not assigning the thread pointer to NULL which is necessary to avoid
undefined behavior

##### Reproduction Steps
2. Perform Ign Off

##### Expected Behavior
SDL should not crash

##### Observed Behavior
SDL crashed

#### OS & Version Information
* OS/Version : SYNC4

LuxoftAKutsan:
Fixes #3178

### Risk

### Testing Plan
Script should be created.

### Summary

    We used aggregation in Thread class and we dont have contorol over
    delegate life cycle, so logically need first to delete container and then delete delegate.
    we are calling join() -> stop() Inside destuctor of Thread
    we are using delegate_->exitThreadMain(); Inside the stop()
    
    Shutdown () was removed because it is redundant. Destructor Thread will make this.

    The core crash shows evidence that something around set_delegate is
    crashing, on the other hand ~MessageLoopThread is deleting the thread
    and not assigning the thread pointer to NULL which is necessary to avoid
    undefined behavior

### Tasks Remaining:
- [ ] Create script for the issue
- [ ] Internal Luxoft review
- [ ] Check regression

### CLA


AByzhynar:
@LuxoftAKutsan In the PR description you have stated the following:
```
Shutdown () was removed because it is redundant.
``` 
As I can see it was not removed. Remove it please or change PR description.

src/components/include/utils/threads/message_loop_thread.h
```diff
@@ -148,8 +148,9 @@ MessageLoopThread<Q>::MessageLoopThread(const std::string& name,
 template <class Q>
 MessageLoopThread<Q>::~MessageLoopThread() {
   Shutdown();
-  delete thread_delegate_;
   threads::DeleteThread(thread_);
+  thread_ = NULL;
```
AByzhynar, line 152: @LuxoftAKutsan Why do you set thread_ pointer to `NULL` after deletion but did not set `thread_delegate_`? It is a pointer as well.

JackLivio:
Potential related fix #3388

iCollin:
With these proposed changes, at the time `thread_delegate_` is deleted, it will hold a reference to `thread_` which has been deleted. Within the destructor of the delegate we find this code:

```
ThreadDelegate::~ThreadDelegate() {
  if (thread_) {
    thread_->set_delegate(NULL);
  }
}
```

This will cause an invalid write.

iCollin:
Thread delegates must be destroyed before threads.

Otherwise, when the delegate is destroyed it will attempt to write on the thread pointer that has been deleted.

```
ThreadDelegate::~ThreadDelegate() {
  if (thread_) {
    thread_->set_delegate(NULL);
  }
}
```

src/components/include/utils/threads/message_loop_thread.h
```diff
@@ -148,8 +148,9 @@ MessageLoopThread<Q>::MessageLoopThread(const std::string& name,
 template <class Q>
 MessageLoopThread<Q>::~MessageLoopThread() {
   Shutdown();
-  delete thread_delegate_;
```
jacobkeeler, line 151: Thread delegate must be deleted before the thread itself. All threads in the project follow this pattern

theresalech:

LuxoftAKutsan:
@theresalech this work is planned. We will notify when fix with test scripts will be available.

LuxoftAKutsan:
Agree with PM comments.

Tested attached script on current develop looks like issued covered with #3388 fix.

PR #3385, opened 2020-05-20
Filter unknown enums when removing unknown parameters

### LINKED ISSUE #2857: Core rejects RAI if unknown enums are present
### Bug Report
Core by default rejects any message with `INVALID_DATA` if it contains any unknown enum values, but this should not be done for the RegisterAppInterface request, since this message is sent before version negotiation occurs

##### Reproduction Steps
1. Send a RegisterAppInterface request with a new AppHMIType (ex. "SHOPPING") and an existing AppHMIType (ex. "MEDIA")

##### Expected Behavior
Core sends a `SUCCESS` response with the negotiated version, which can be used by the application to determine that the new AppHMIType is not supported by the module

##### Observed Behavior
Core responds with `INVALID_DATA`

* OS/Version: Ubuntu 18.04
* SDL Core Version: 5.0.0
* Testing Against: iOS test app

jacobkeeler:
Fixes #2857

### Risk

### Testing Plan
Follow issue reproduction steps in #2857 

### Summary
Change schema logic to filter out unknown enum values when removing unknown parameter

##### Bug Fixes
* Fixes issue where RAI was rejected if it contained new enums before version negotiation was completed.

### CLA

### DIFF
--- a/src/components/smart_objects/include/smart_objects/array_schema_item.h
+++ b/src/components/smart_objects/include/smart_objects/array_schema_item.h
@@ -113,6 +113,8 @@ class CArraySchemaItem : public ISchemaItem {
   void BuildObjectBySchema(const SmartObject& pattern_object,
                            SmartObject& result_object) OVERRIDE;
 
+  TypeID GetType() OVERRIDE;
+
  private:
   /**
    * @brief Constructor.
@@ -124,6 +126,16 @@ class CArraySchemaItem : public ISchemaItem {
   CArraySchemaItem(const ISchemaItemPtr ElementSchemaItem,
                    const TSchemaItemParameter<size_t>& MinSize,
                    const TSchemaItemParameter<size_t>& MaxSize);
+
+  /**
+   * @brief Removes unknown parameters from array.
+   * @param Object Array to remove unknown parameters.
+   * @param MessageVersion The version to check against for which parameters are
+   *                       unknown.
+   **/
+  void RemoveUnknownParams(SmartObject& Object,
+                           const utils::SemanticVersion& MessageVersion);
+
   /**
    * @brief SchemaItem for array elements.
    **/
--- a/src/components/smart_objects/include/smart_objects/enum_schema_item.h
+++ b/src/components/smart_objects/include/smart_objects/enum_schema_item.h
@@ -166,6 +166,7 @@ class TEnumSchemaItem : public CDefaultSchemaItem<EnumType> {
                       ElementSignatures);
   SmartType getSmartType() const OVERRIDE;
   EnumType getDefaultValue() const OVERRIDE;
+  TypeID GetType() OVERRIDE;
   /**
    * @brief Set of allowed enumeration elements.
    **/
@@ -410,6 +411,11 @@ SmartType TEnumSchemaItem<EnumType>::getSmartType() const {
 template <typename EnumType>
 EnumType TEnumSchemaItem<EnumType>::getDefaultValue() const {
   return EnumType::INVALID_ENUM;
+}
+
+template <typename EnumType>
+TypeID TEnumSchemaItem<EnumType>::GetType() {
+  return TYPE_ENUM;
 }
 
 template <typename EnumType>
--- a/src/components/smart_objects/include/smart_objects/schema_item.h
+++ b/src/components/smart_objects/include/smart_objects/schema_item.h
@@ -49,6 +49,16 @@ namespace ns_smart_objects {
 class SmartObject;
 class SMember;
 
+enum TypeID {
+  TYPE_NONE,
+  TYPE_OBJECT,
+  TYPE_ARRAY,
+  TYPE_STRING,
+  TYPE_NUMBER,
+  TYPE_ENUM,
+  TYPE_BOOLEAN
+};
+
 /**
  * @brief Base schema item.
  **/
@@ -142,6 +152,13 @@ class ISchemaItem {
    * @return value of any parameter
    */
   virtual size_t GetMemberSize();
+
+  /**
+   * @brief Get type ID of schema
+   *
+   * @return The type ID of this schema
+   */
+  virtual TypeID GetType();
 
   virtual ~ISchemaItem() {}
 };
--- a/src/components/smart_objects/src/array_schema_item.cc
+++ b/src/components/smart_objects/src/array_schema_item.cc
@@ -29,6 +29,8 @@
 // POSSIBILITY OF SUCH DAMAGE.
 
 #include "smart_objects/array_schema_item.h"
+
+#include "smart_objects/enum_schema_item.h"
 #include "smart_objects/smart_object.h"
 
 namespace ns_smart_device_link {
@@ -100,6 +102,28 @@ void CArraySchemaItem::applySchema(
           Object[i], remove_unknown_parameters, MessageVersion);
     }
   }
+  if (remove_unknown_parameters) {
+    RemoveUnknownParams(Object, MessageVersion);
+  }
+}
+
+void CArraySchemaItem::RemoveUnknownParams(
+    SmartObject& Object, const utils::SemanticVersion& MessageVersion) {
+  if (SmartType_Array == Object.getType()) {
+    auto array = Object.asArray();
+    for (auto it = array->begin(); it != array->end();) {
+      rpc::ValidationReport report("");
+      const bool is_invalid_enum =
+          TYPE_ENUM == mElementSchemaItem->GetType() &&
+          (mElementSchemaItem->validate(*it, &report, MessageVersion, false) !=
+           errors::OK);
+      if (is_invalid_enum) {
+        it = array->erase(it);
+      } else {
+        ++it;
+      }
+    }
+  }
 }
 
 void CArraySchemaItem::unapplySchema(SmartObject& Object,
@@ -125,6 +149,10 @@ void CArraySchemaItem::BuildObjectBySchema(const SmartObject& pattern_object,
   }
   // empty array
   result_object = SmartObject(SmartType_Array);
+}
+
+TypeID CArraySchemaItem::GetType() {
+  return TYPE_ARRAY;
 }
 
 CArraySchemaItem::CArraySchemaItem(const ISchemaItemPtr ElementSchemaItem,
--- a/src/components/smart_objects/src/object_schema_item.cc
+++ b/src/components/smart_objects/src/object_schema_item.cc
@@ -35,6 +35,7 @@
 
 #include "generated_msg_version.h"
 #include "smart_objects/always_false_schema_item.h"
+#include "smart_objects/enum_schema_item.h"
 #include "smart_objects/smart_object.h"
 
 namespace {
@@ -165,10 +166,6 @@ void CObjectSchemaItem::applySchema(
     return;
   }
 
-  if (remove_unknown_parameters) {
-    RemoveFakeParams(Object, MessageVersion);
-  }
-
   SmartObject default_value;
 
   for (const auto& item : mMembers) {
@@ -184,6 +181,10 @@ void CObjectSchemaItem::applySchema(
       member.mSchemaItem->applySchema(
           Object[key], remove_unknown_parameters, MessageVersion);
     }
+  }
+
+  if (remove_unknown_parameters) {
+    RemoveUnknownParams(Object, MessageVersion);
   }
 }
 
@@ -232,6 +233,10 @@ size_t CObjectSchemaItem::GetMemberSize() {
   return mMembers.size();
 }
 
+TypeID CObjectSchemaItem::GetType() {
+  return TYPE_OBJECT;
+}
+
 boost::optional<SMember&> CObjectSchemaItem::GetMemberSchemaItem(
     const std::string& member_key) {
   auto it = mMembers.find(member_key);
@@ -250,7 +255,7 @@ void CObjectSchemaItem::AddMemberSchemaItem(const std::string& member_key,
 CObjectSchemaItem::CObjectSchemaItem(const Members& members)
     : mMembers(members) {}
 
-void CObjectSchemaItem::RemoveFakeParams(
+void CObjectSchemaItem::RemoveUnknownParams(
     SmartObject& Object, const utils::SemanticVersion& MessageVersion) {
   for (const auto& key : Object.enumerate()) {
     std::map<std::string, SMember>::const_iterator members_it =
@@ -259,7 +264,13 @@ void CObjectSchemaItem::RemoveFakeParams(
     if (mMembers.end() != members_it) {
       const SMember* member =
           GetCorrectMember(members_it->second, MessageVersion);
-      if (!member || member->mIsRemoved) {
+      rpc::ValidationReport report("");
+      const bool is_invalid_enum =
+          TYPE_ENUM == member->mSchemaItem->GetType() &&
+          (member->mSchemaItem->validate(
+               Object.getElement(key), &report, MessageVersion, false) !=
+           errors::OK);
+      if (!member || member->mIsRemoved || is_invalid_enum) {
         Object.erase(key);
       }
       continue;
--- a/src/components/smart_objects/include/smart_objects/bool_schema_item.h
+++ b/src/components/smart_objects/include/smart_objects/bool_schema_item.h
@@ -51,6 +51,8 @@ class CBoolSchemaItem : public CDefaultSchemaItem<bool> {
       const TSchemaItemParameter<bool>& DefaultValue =
           TSchemaItemParameter<bool>());
 
+  TypeID GetType() OVERRIDE;
+
  private:
   explicit CBoolSchemaItem(const TSchemaItemParameter<bool>& DefaultValue);
   SmartType getSmartType() const OVERRIDE;
--- a/src/components/smart_objects/include/smart_objects/number_schema_item.h
+++ b/src/components/smart_objects/include/smart_objects/number_schema_item.h
@@ -82,6 +82,8 @@ class TNumberSchemaItem : public CDefaultSchemaItem<NumberType> {
       const utils::SemanticVersion& MessageVersion = utils::SemanticVersion(),
       const bool allow_unknown_enums = false) OVERRIDE;
 
+  TypeID GetType() OVERRIDE;
+
  private:
   /**
    * @brief Get smart type for this NumberType.
@@ -189,6 +191,11 @@ errors::eType TNumberSchemaItem<NumberType>::validate(
     return errors::OUT_OF_RANGE;
   }
   return errors::OK;
+}
+
+template <typename NumberType>
+TypeID TNumberSchemaItem<NumberType>::GetType() {
+  return TYPE_NUMBER;
 }
 
 template <typename NumberType>
--- a/src/components/smart_objects/include/smart_objects/object_schema_item.h
+++ b/src/components/smart_objects/include/smart_objects/object_schema_item.h
@@ -127,8 +127,8 @@ class CObjectSchemaItem : public ISchemaItem {
    * @brief Apply schema.
    * @param Object Object to apply schema.
    * @param remove_unknown_parameters contains true if need to remove unknown
-   *parameters
-   * from smart object otherwise contains false.
+   *                                  parameters from smart object otherwise
+   *                                  contains false.
    **/
   void applySchema(SmartObject& Object,
                    const bool remove_unknown_parameters,
@@ -138,7 +138,7 @@ class CObjectSchemaItem : public ISchemaItem {
    * @brief Unapply schema.
    * @param Object Object to unapply schema.
    * @param remove_unknown_parameters contains true if need to remove unknown
-   *parameters
+   *                                  parameters
    **/
   void unapplySchema(SmartObject& Object,
                      const bool remove_unknown_parameters) OVERRIDE;
@@ -156,6 +156,8 @@ class CObjectSchemaItem : public ISchemaItem {
    */
   size_t GetMemberSize() OVERRIDE;
 
+  TypeID GetType() OVERRIDE;
+
   boost::optional<SMember&> GetMemberSchemaItem(
       const std::string& member_key) OVERRIDE;
 
@@ -171,25 +173,27 @@ class CObjectSchemaItem : public ISchemaItem {
   CObjectSchemaItem(const Members& Members);
 
   /**
-   * @brief Removes fake parameters from object.
-   * @param Object Object to remove fake parameters.
+   * @brief Removes unknown parameters from object.
+   * @param Object Object to remove unknown parameters.
+   * @param MessageVersion The version to check against for which parameters are
+   *                       unknown.
    **/
-  void RemoveFakeParams(SmartObject& Object,
-                        const utils::SemanticVersion& MessageVersion);
+  void RemoveUnknownParams(SmartObject& Object,
+                           const utils::SemanticVersion& MessageVersion);
 
   /**
    * @brief Returns the correct schema item based on message version.
    * @param member Schema member
    * @param MessageVersion Semantic Version of mobile message.
    * @return Pointer to correct schema item if item found or nullptr, if item
-   *was not found.
+   *         was not found.
    **/
   const SMember* GetCorrectMember(const SMember& member,
                                   const utils::SemanticVersion& messageVersion);
 
   /**
    * @brief Map of member name to SMember structure describing the object
-   *member.
+   *        member.
    **/
   Members mMembers;
   DISALLOW_COPY_AND_ASSIGN(CObjectSchemaItem);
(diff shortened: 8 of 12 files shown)


src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -298,7 +298,7 @@ void CommandRequestImpl::SendResponse(
     response[strings::msg_params] = *response_params;
   }
 
-  if (info) {
+  if (info && !info.empty()) {
```
JackLivio, line 301: Does not compile since its a char array, need to check if first character is '\0'

```suggestion
  if (info && *info != '\0') {
```

src/components/application_manager/src/commands/command_request_impl.cc
```diff
@@ -318,7 +318,17 @@ void CommandRequestImpl::SendResponse(
   }
 
   response[strings::msg_params][strings::success] = success;
-  response[strings::msg_params][strings::result_code] = result_code;
+  if ((result_code == mobile_apis::Result::SUCCESS ||
+       result_code == mobile_apis::Result::WARNINGS) &&
+      !warning_info().empty()) {
+    response[strings::msg_params][strings::info] =
+        (info && !info.empty()) ? std::string(info) + "\n" + warning_info()
```
JackLivio, line 325:
```suggestion
        (info && *info != '\0') ? std::string(info) + "\n" + warning_info()
```

src/components/smart_objects/include/smart_objects/enum_schema_item.h
```diff
@@ -314,6 +317,21 @@ const ElementSignature TEnumSchemaItem<EnumType>::getSignature(
...
 
+template <typename EnumType>
+bool TEnumSchemaItem<EnumType>::filterInvalidEnums(
+    SmartObject& Object,
+    const utils::SemanticVersion& MessageVersion,
+    rpc::ValidationReport* report__) {
+  rpc::ValidationReport dummy_report("");
+  if (validate(Object, &dummy_report, MessageVersion, false) != errors::OK) {
+    std::string validation_info =
+        "Filtered invalid value - " + Object.asString();
```
JackLivio, line 328: Not sure if this is the validation info passed back to mobile but if it is, then it needs to follow the format mentioned in the proposal:

> Invalid enums were removed: <param_1>:<enum_value_1>,<enum_value_2>;<param_2>:<enum_value_3> ...

src/components/smart_objects/include/smart_objects/enum_schema_item.h
```diff
@@ -128,15 +128,17 @@ class TEnumSchemaItem : public CDefaultSchemaItem<EnumType> {
       const std::vector<ElementSignature>& signatures,
       const utils::SemanticVersion& MessageVersion);
 
+  bool filterInvalidEnums(SmartObject& Object,
```
JackLivio, line 131: Please add method description of params and return value

src/components/smart_objects/include/smart_objects/schema_item.h
```diff
@@ -91,13 +101,27 @@ class ISchemaItem {
...
 
+  /**
+   * @brief Filter invalid enum values
+   *
+   * @param Object Object to check for invalid enum values
+   * @param MessageVersion the version of the schema to use for validation
+   * @param report__ object for reporting enums which were removed during the
+   * process
+   *
+   * @return true if the value being checked should be filtered, false otherwise
```
JackLivio, line 112: The return value is a bit confusing. It says if true it should be filtered, but isnt the enum implementation of this method doing the filtering?

src/components/smart_objects/src/array_schema_item.cc
```diff
@@ -90,6 +92,31 @@ errors::eType CArraySchemaItem::validate(
...
+      strVal << index;
+      if (mElementSchemaItem->filterInvalidEnums(
+              *it, MessageVersion, &report__->ReportSubobject(strVal.str()))) {
+        it = array->erase(it);
+      } else {
+        ++it;
+      }
+      index++;
+    }
+    if (array->empty()) {
```
JackLivio, line 113: Should this method return true if there was at least enum item removed?

src/components/smart_objects/src/smart_schema.cc
```diff
@@ -59,8 +59,16 @@ ISchemaItemPtr CSmartSchema::getSchemaItem() {
...
-                               const utils::SemanticVersion& MessageVersion) {
+                               const utils::SemanticVersion& MessageVersion,
+                               rpc::ValidationReport* report__) {
   mSchemaItem->applySchema(Object, remove_unknown_parameters, MessageVersion);
+  if (remove_unknown_parameters) {
+    rpc::ValidationReport dummy_report("");
+    if (!report__) {
+      report__ = &dummy_report;
+    }
+    mSchemaItem->filterInvalidEnums(Object, MessageVersion, report__);
```
JackLivio, line 70: Should the return value be checked here?

JackLivio, line 70: Its a little confusing that the remove invalid enums was put in applySchema and not in `RemoveUnkownParams`. Any reason why it was split like this?

JackLivio, line 70: Should the call to filter unknown enums be moved to RemoveUnknownParams? The code is all over the place and its pretty hard to follow whats happening where.

src/components/application_manager/src/rpc_service_impl.cc
```diff
@@ -245,6 +246,12 @@ bool RPCServiceImpl::ManageMobileCommand(
           app->hmi_level(mobile_apis::PredefinedWindows::DEFAULT_WINDOW);
     }
 
+    commands::CommandRequestImpl* request =
+        dynamic_cast<commands::CommandRequestImpl*>(command.get());
+    if (request != NULL) {
+      request->set_warning_info(warning_info);
```
LuxoftAKutsan, line 252: Can we move set_warning_info to base class?

src/components/formatters/include/formatters/CSmartFactory.h
```diff
@@ -277,7 +279,8 @@ template <class FunctionIdEnum, class MessageTypeEnum, class StructIdEnum>
 bool CSmartFactory<FunctionIdEnum, MessageTypeEnum, StructIdEnum>::attachSchema(
     ns_smart_device_link::ns_smart_objects::SmartObject& object,
     const bool remove_unknown_parameters,
-    const utils::SemanticVersion& MessageVersion) {
+    const utils::SemanticVersion& MessageVersion,
+    rpc::ValidationReport* report__) {
```
LuxoftAKutsan, line 283: Do we need 2 underscores after parameter name?

src/components/include/application_manager/rpc_service.h
```diff
@@ -54,10 +54,15 @@ class RPCService {
    * @brief ManageMobileCommand convert message to mobile command and execute it
    * @param message pointer to received message
    * @param origin origin of command
+   * @param warning_info warning message to send on a successful response. Only
+   * applies to requests from mobile.
    * @return true if command is executed, otherwise return false
    */
-  virtual bool ManageMobileCommand(const commands::MessageSharedPtr message,
-                                   commands::Command::CommandSource source) = 0;
+  virtual bool ManageMobileCommand(
```
LuxoftAKutsan, line 61: I do understand benefit of adding default argument, we will not have additional method and mock. 
But IMHO and according to [google coding stye](https://google.github.io/styleguide/cppguide.html#Default_Arguments) default arguments in virtual functions is bad practice. 
I would propose to make additional function `ManageMobileCommandExtended`/`ManageMobileCommandWarning`/`ManageMobileInfoCommand` ( naming is complicated) with warning code,  and keep ManageMobileCommand as proxy to it.

src/components/smart_objects/include/smart_objects/array_schema_item.h
```diff
@@ -80,13 +80,17 @@ class CArraySchemaItem : public ISchemaItem {
       const utils::SemanticVersion& MessageVersion = utils::SemanticVersion(),
       const bool allow_unknown_enums = false) OVERRIDE;
 
+  bool filterInvalidEnums(SmartObject& Object,
```
LuxoftAKutsan, line 83: Shouldn't we start from capital letter?

src/components/smart_objects/include/smart_objects/schema_item.h
```diff
@@ -91,13 +101,27 @@ class ISchemaItem {
...
+   **/
+  virtual bool filterInvalidEnums(SmartObject& Object,
```
LuxoftAKutsan, line 114: I see check of return value only in unit tests. Do we need it?

src/components/smart_objects/src/array_schema_item.cc
```diff
@@ -90,6 +92,31 @@ errors::eType CArraySchemaItem::validate(
   return errors::OK;
 }
 
+bool CArraySchemaItem::filterInvalidEnums(
+    SmartObject& Object,
+    const utils::SemanticVersion& MessageVersion,
+    rpc::ValidationReport* report__) {
+  if (SmartType_Array == Object.getType()) {
```
LuxoftAKutsan, line 99: I would use reverse if end early exist. Just as suggestion.

src/components/smart_objects/src/array_schema_item.cc
```diff
@@ -90,6 +92,31 @@ errors::eType CArraySchemaItem::validate(
...
+    int index = 0;
+    auto array = Object.asArray();
+    for (auto it = array->begin(); it != array->end();) {
+      std::stringstream strVal;
+      strVal << index;
```
LuxoftAKutsan, line 104: We ca use https://en.cppreference.com/w/cpp/string/basic_string/to_string

src/components/smart_objects/src/array_schema_item.cc
```diff
@@ -90,6 +92,31 @@ errors::eType CArraySchemaItem::validate(
   return errors::OK;
 }
 
+bool CArraySchemaItem::filterInvalidEnums(
```
LuxoftAKutsan, line 95: I feel this function as rather complicated one. 
I would try probably to sacrifice performance but make it more easy to read and obvious.

src/components/smart_objects/src/object_schema_item.cc
```diff
@@ -157,6 +158,31 @@ errors::eType CObjectSchemaItem::validate(
   return errors::OK;
 }
 
+bool CObjectSchemaItem::filterInvalidEnums(
+    SmartObject& Object,
+    const utils::SemanticVersion& MessageVersion,
+    rpc::ValidationReport* report__) {
+  for (const auto& key : Object.enumerate()) {
```
LuxoftAKutsan, line 165: Why don't you check  `Object.getType` ?

src/components/smart_objects/include/smart_objects/schema_item.h
jacobkeeler: Yes, it is used recursively within the function. If a particular element returns true, that means that it should be removed from the containing structure. This return value is not used at the root level because that would mean that the full structure would be destroyed.

src/components/smart_objects/src/array_schema_item.cc
jacobkeeler: Are you just suggesting the operands be swapped? I'm a bit confused by this comment

src/components/smart_objects/src/array_schema_item.cc
jacobkeeler: No, that would mean that the array should be removed, which doesn't make sense in that case

jacobkeeler: That said, it would probably be better to return if the list is shorter than the minimum length, rather than explicitly checking for an empty list

src/components/smart_objects/src/object_schema_item.cc
jacobkeeler: Ah, I added the `GetType` function for use with my original (and simpler) implementation of this, it didn't turn out to be necessary in the final implementation, so I can remove that function.

src/components/smart_objects/src/smart_schema.cc
jacobkeeler: I could leave a log here if that works, not much else we could do with it though in this case.

src/components/smart_objects/src/array_schema_item.cc
jacobkeeler: Perhaps I might be able to add some extra comments for clarity. I don't think there's really a way to make it simpler though

src/components/smart_objects/include/smart_objects/enum_schema_item.h
jacobkeeler: It's in the `schema_item.h` header, we should only have it in one place

src/components/smart_objects/include/smart_objects/array_schema_item.h
jacobkeeler: Perhaps by normal case rules, I was following the standards of this class in this case (ex. `validate`, `applySchema`, etc.)

src/components/formatters/include/formatters/CSmartFactory.h
jacobkeeler: Ah, just following previous examples, I can remove it

src/components/application_manager/src/rpc_service_impl.cc
jacobkeeler: I can't see it being useful for any other type of message, but I do agree that casting it down is a bit ugly. I can move it to make things a bit more clean.

jacobkeeler:
@LuxoftAKutsan ah yes, this was originally just to fix the one issue, but after discovering the proposal I changed it to match the proposal requirements. I added the proposal to the PR description accordingly.

src/components/smart_objects/src/smart_schema.cc
jacobkeeler: Ah, sadly the logger is not available in the smart objects component, so there really isn't anything to do here.

src/components/formatters/include/formatters/CSmartFactory.h
jacobkeeler: Fixed in https://github.com/smartdevicelink/sdl_core/pull/3385/commits/bee65b5d8fe27964ed09db991e9ecd8c3b7d8d2c

src/components/application_manager/src/rpc_service_impl.cc
jacobkeeler: Fixed in https://github.com/smartdevicelink/sdl_core/pull/3385/commits/6840997e72ef482f670c747b72ac78ecc8d29fb3

src/components/smart_objects/src/array_schema_item.cc
jacobkeeler: Added comments in https://github.com/smartdevicelink/sdl_core/pull/3385/commits/147b7bca79fe459cdceeeee6d7b8a74133890571

src/components/smart_objects/src/object_schema_item.cc
jacobkeeler: I ended up using it in this class to filter out single invalid enums from the object

src/components/smart_objects/src/array_schema_item.cc
jacobkeeler: Fixed in https://github.com/smartdevicelink/sdl_core/pull/3385/commits/bee65b5d8fe27964ed09db991e9ecd8c3b7d8d2c

src/components/smart_objects/src/array_schema_item.cc
jacobkeeler: Ah, I think I see your point, it should return true ONLY if at least one item is removed. Fixed in https://github.com/smartdevicelink/sdl_core/pull/3385/commits/bee65b5d8fe27964ed09db991e9ecd8c3b7d8d2c

src/components/include/application_manager/rpc_service.h
jacobkeeler: Changed to overload in https://github.com/smartdevicelink/sdl_core/pull/3385/commits/446fe5c5c95865f6d7d088a088cbd5d655de2851

jacobkeeler:
@LuxoftAKutsan @JackLivio addressed comments

src/components/smart_objects/src/array_schema_item.cc
LuxoftAKutsan: Consider following approach : 
```cpp
if (SmartType_Array != Object.getType()) { 
  return false;
  // This block may be removed within removing Object.getType(
}

auto array = Object.asArray();
auto elements_to_remove = std::find_if(array->begin(), array->end(), []() {
  mElementSchemaItem->filterInvalidEnums(...)
});
for (element: elements_to_remove) {
  array.eraze(element);
}
// min size staff using elements_to_remove.size()
}
```
I didn't compile it, so there could be issues with syntax. 
 
I see this benefits : 
 - No nested structures 
 - No iterator crutches.

I agree that it probably less efficient then previous one. But we do not expect huge amount of elements in `Object`, do we?

src/components/smart_objects/src/object_schema_item.cc
```diff
@@ -157,6 +158,39 @@ errors::eType CObjectSchemaItem::validate(
   return errors::OK;
 }
 
+bool CObjectSchemaItem::filterInvalidEnums(
+    SmartObject& Object,
+    const utils::SemanticVersion& MessageVersion,
+    rpc::ValidationReport* report) {
+  bool valid = true;
+  for (const auto& key : Object.enumerate()) {
+    std::map<std::string, SMember>::const_iterator members_it =
```
LuxoftAKutsan, line 167: good place for auto

src/components/smart_objects/src/object_schema_item.cc
```diff
@@ -157,6 +158,39 @@ errors::eType CObjectSchemaItem::validate(
   return errors::OK;
 }
 
+bool CObjectSchemaItem::filterInvalidEnums(
```
LuxoftAKutsan, line 161: Consider following approach. 
```cpp


bool CObjectSchemaItem::filterInvalidEnums(
    SmartObject& Object,
    const utils::SemanticVersion& MessageVersion,
    rpc::ValidationReport* report) {
  
  auto filter_invalid_enums = [this, &MessageVersion, report](member) {
      const SMember* member =
          GetCorrectMember(members_it->second, MessageVersion);
      rreturn member->mSchemaItem->filterInvalidEnums(
              Object[key], MessageVersion, &report->ReportSubobject(key));
  };

  auto should_be_erazed = [](member) -> bool {
        // The member is safe to filter if it is non-mandatory
        // Object is no longer valid if the member is mandatory.
        // To maintain the general structure of the message, only leaf nodes
        // (individual enum values) should be filtered in this case.
      return member->mSchemaItem->GetType() == TYPE_ENUM && !member->mIsMandatory;
  };

  bool valid = true;
  for (const auto& key : Object.enumerate()) {
    auto members_it = mMembers.find(key);
    DCHECK_OR_RETURN(mMembers.end() != members_it, false);  //  object should match schema
    if (filter_invalid_enums(*members_it)) {
        if (member->mIsMandatory) {
          valid = false;
        }
        if (should_be_erazed(member)) {
            Object.erase(key); 
        }
    }
  }
  return !valid;
}
```

The same as above, I didn't check it and it not so performant. but I feel it as more readable.

src/components/smart_objects/test/ArraySchemaItem_test.cc
```diff
@@ -345,6 +349,92 @@ TEST(test_map_validate, test_ArraySchemaItemTest) {
...
+
+  rpc::ValidationReport report("RPC");
+  EXPECT_TRUE(item->filterInvalidEnums(obj, utils::SemanticVersion(), &report));
+  EXPECT_NE(std::string(""), rpc::PrettyFormat(report));
+
+  EXPECT_EQ((size_t)0, obj.length());
+
+  report = rpc::ValidationReport("RPC");
+  EXPECT_EQ(errors::OUT_OF_RANGE, item->validate(obj, &report));
+  EXPECT_NE(std::string(""), rpc::PrettyFormat(report));
```
LuxoftAKutsan, line 410: Why do we expect empty report?

src/components/smart_objects/test/ArraySchemaItem_test.cc
```diff
@@ -345,6 +349,92 @@ TEST(test_map_validate, test_ArraySchemaItemTest) {
...
+}
```
LuxoftAKutsan, line 411: Would be good also to cover with UT : 
 - filtering of mandatory\not mandatory
 - filtering of nested structures with mandatory\not mandatory
 - warning message

src/components/smart_objects/test/ArraySchemaItem_test.cc
jacobkeeler: We expect a non-empty report (`EXPECT_NE`). I don't think we should enforce a strict format, just verify that some warning was generated

src/components/smart_objects/include/smart_objects/array_schema_item.h
```diff
@@ -113,6 +117,8 @@ class CArraySchemaItem : public ISchemaItem {
   void BuildObjectBySchema(const SmartObject& pattern_object,
                            SmartObject& result_object) OVERRIDE;
 
+  TypeID GetType() OVERRIDE;
```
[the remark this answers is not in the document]
jacobkeeler: Per my later comment (https://github.com/smartdevicelink/sdl_core/pull/3385#discussion_r436099463), I found that it was needed in the latest changes to `object_schema_item.cc`

src/components/smart_objects/include/smart_objects/enum_schema_item.h
```diff
@@ -314,10 +317,25 @@ const ElementSignature TEnumSchemaItem<EnumType>::getSignature(
   return ret;
 }
 
+template <typename EnumType>
+bool TEnumSchemaItem<EnumType>::filterInvalidEnums(
+    SmartObject& Object,
+    const utils::SemanticVersion& MessageVersion,
+    rpc::ValidationReport* report) {
+  rpc::ValidationReport dummy_report("");
+  if (validate(Object, &dummy_report, MessageVersion, false) != errors::OK) {
```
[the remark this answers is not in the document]
jacobkeeler: Since `false` is the default value for this parameter, I could just remove it from this call

src/components/smart_objects/src/object_schema_item.cc
jacobkeeler: Seems a lambda would be overkill for "should_be_erazed" when a local variable would work fine. IMO unnecessary lambdas make things less clear overall. 

I'll see what I can do to clean it up.

src/components/smart_objects/test/ArraySchemaItem_test.cc
jacobkeeler: Added tests: https://github.com/smartdevicelink/sdl_core/pull/3385/commits/e547238c73dc223287a8ac46cfbe6a6e92e3c18c

src/components/smart_objects/src/object_schema_item.cc
jacobkeeler: Fixed in https://github.com/smartdevicelink/sdl_core/pull/3385/commits/e547238c73dc223287a8ac46cfbe6a6e92e3c18c

src/components/smart_objects/include/smart_objects/enum_schema_item.h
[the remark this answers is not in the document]
jacobkeeler: Fixed in https://github.com/smartdevicelink/sdl_core/pull/3385/commits/e547238c73dc223287a8ac46cfbe6a6e92e3c18c

src/components/application_manager/include/application_manager/rpc_handler_impl.h
```diff
@@ -176,6 +176,7 @@ class RPCHandlerImpl : public RPCHandler,
 
   bool ConvertMessageToSO(const Message& message,
                           smart_objects::SmartObject& output,
+                          std::string& warning_info,
```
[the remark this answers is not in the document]
jacobkeeler: Fixed in https://github.com/smartdevicelink/sdl_core/pull/3385/commits/e547238c73dc223287a8ac46cfbe6a6e92e3c18c

src/components/smart_objects/src/object_schema_item.cc
jacobkeeler: Refactored in https://github.com/smartdevicelink/sdl_core/pull/3385/commits/b4452abf1c2f607f6dc68cb96889cdaff8363dde

jacobkeeler:
@LuxoftAKutsan addressed new comments

LuxoftAKutsan:
I see that proposal implementation merged to develop. 
Is there ant of following components: 
 - Regression testing result
 - Manual test plan. 
 - Don't we expect any changes in automated test coverage that may trigger changes in sdl_core?

PR #3389, opened 2020-05-21
Fix problem with sending onAppInterfaceUnregistered notification during execution of UnregisterAppInterface request

### LINKED ISSUE #1178: Core sends an onAppInterfaceUnregistered notification when UnregisterAppInterface is sent from Mobile
It seems as though we have an issue with core sending both an `onAppInterfaceUnregistered` notification and an `UnregisterAppInterface` response when the mobile proxy sends over `UnregisterAppInterface`. 
This should not be the case, as the notification should only be sent if there is an unregistration from the core side that was not caused by the mobile side. The notification also has a reason of -1, which is not a valid value. 

We should remove the notification from being sent when the `UnregisterAppInterface` RPC is sent.

ychernysheva:
Fixes #1178 

### Risk

### Testing Plan
Corresponding unit test is updated

### Summary
This pull request is intended to fix issue with sending onAppInterfaceUnregistered notification during execution of UnregisterAppInterface request. Root cause of this issue is a sending OnAppInterfaceUnregistered notification with invalid unregistration reason in method UnregisterAppInterfaceRequest::Run().
That's why this notification was removed from UnregisterAppInterfaceRequest::Run() method and corresponding unit test.

### CLA

### DIFF
--- a/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/unregister_app_interface_request.cc
+++ b/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/unregister_app_interface_request.cc
@@ -49,11 +49,6 @@ void UnregisterAppInterfaceRequest::Run() {
     return;
   }
 
-  rpc_service_.ManageMobileCommand(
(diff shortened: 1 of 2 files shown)


ychernysheva:
@theresalech this PR is ready for review.
Thank you.

PR #3417, opened 2020-06-02
Add SDL Core Daemon Script

### LINKED ISSUE #2124: [SDL 0154] Add SDL Core Daemon script
Proposal: [Add SDL Core Daemon script](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0154-add-sdl-core-daemon-script.md)

> This proposal is for the addition of a daemon script allowing simple control over the lifecycle of SDL Core. This script would allow users to easily start/restart/stop SDL Core in the background without needing to create their own scripts to manage this behavior, as is currently the case.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/444

jacobkeeler:
Fixes #2124

### Risk

### Testing Plan
Test all of the commands in the daemon script in several configurations and verify that it responds appropriately:
```
./core.sh start
./core.sh stop
./core.sh restart
./core.sh kill
```

### Summary
Added SDL Core daemon script

##### Enhancements
* Added SDL Core daemon script, allowing a user to easily start and stop Core as a background process

### CLA


src/appMain/core.sh
```diff
@@ -0,0 +1,74 @@
...
+    RESULT=0
+  fi
+
+  if [ -f "$CORE_PID_FILE" ]; then
+    rm $CORE_PID_FILE
+  fi 
+  return $RESULT
+}
+
+if [ x$1 == xstop ]; then
```
iCollin, line 52: i always have preferred quotes to doing this letter in front hack to avoid the unary operator warning, but I am not a big bash guy. Why do you prefer this approach to using quotes?

src/appMain/main.cc
```diff
@@ -135,7 +135,8 @@ int32_t main(int32_t argc, char** argv) {
   // Logger initialization
   INIT_LOGGER("log4cxx.properties", profile_instance.logs_enabled());
 
-  threads::Thread::SetNameForId(threads::Thread::CurrentId(), "MainThread");
+  threads::Thread::SetNameForId(threads::Thread::CurrentId(),
```
iCollin, line 138: not that this is a bad thing, but couldn't we just use the proper process name in the script instead of doing this change? pgrep/pkill smartDeviceLinkCore would work fine.

src/appMain/core.sh
```diff
@@ -0,0 +1,74 @@
...
+  else
+    echo "Core is not running (or was started outside of this script)"
+  fi
+elif [ x$1 == xstart ]; then
+  core_start
+elif [ x$1 == xkill ]; then
+  core_stop
+  if [ -n "$(pgrep $CORE_PROCESS_NAME)" ]; then
+    echo "Killing all lingering instances of SDL Core"
+    killall -9 $CORE_PROCESS_NAME
```
iCollin, line 70: pgrep is used above, i suggest we use pkill here to have less dependencies

src/appMain/core_external_proprietary.sh
```diff
@@ -0,0 +1,127 @@
...
+    rm $CORE_PID_FILE
+  fi 
+  return $RESULT
+}
+
+function pm_install_dependencies() {
+  pip3 list | grep -F tornado > /dev/null
+  if [ $? -eq 1 ]; then
+    echo "Installing tornado python package"
+    sudo pip3 install tornado
```
iCollin, line 57: why did we remove -H from sudo? does pip3 not use home dir for installs? it may be default on your install to enable -H but I think we should keep this flag.

src/appMain/core_external_proprietary.sh
```diff
@@ -0,0 +1,127 @@
...
+  fi
+}
+
+function pm_start() {
+  if [ -f "$PM_PID_FILE" ] && [ -n "$(ps -p $(cat $PM_PID_FILE) -o pid=)" ]; then
+    echo "Policy Server is already running"
```
iCollin, line 63: core_start() checks pgrep too, i think that check would be useful here as well.

src/appMain/core_external_proprietary.sh
```diff
@@ -0,0 +1,127 @@
...
+  # Clear PID regardless of whether the process was running or not
+  if [ -f "$PM_PID_FILE" ]; then
+    rm $PM_PID_FILE
+  fi
+  return $RESULT
+}
+
+if [ x$1 == xstop ]; then
+  core_stop
+  if [ ! "$?" -eq 0 ]; then
```
iCollin, line 93:
```suggestion
  if [ "$?" -ne 0 ]; then
```

src/appMain/core_external_proprietary.sh
```diff
@@ -0,0 +1,127 @@
...
+    echo "Core is not running (or was started outside of this script)"
+  fi
+
+  pm_stop
+  if [ ! "$?" -eq 0 ]; then
```
iCollin, line 98:
```suggestion
  if [ "$?" -ne 0 ]; then
```

src/appMain/core.sh
jacobkeeler: This was just carried over from the original PR, no real preference either way. I can change it

src/appMain/core.sh
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
+cd $(dirname $0)
+DIR=$(pwd)
+PID_DIR=~/.sdl
```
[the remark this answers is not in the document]
jacobkeeler: Nah, I believe the reason for this choice was that the original script used `/var/run` which was resulting in permissions issues. $DIR should work fine.

src/appMain/core.sh
jacobkeeler: Was unaware of `pkill`, I believe these are all installed by default, but I believe `pkill` would end up being cleaner

src/appMain/core_external_proprietary.sh
jacobkeeler: Seems this was an change carried over from the example in the proposal, which was based on an older version of the script, will fix

src/appMain/daemon.sh
```diff
@@ -0,0 +1,143 @@
...
+
+  if [ "$EXTERNAL_POLICIES" == "true" ]; then
+    pm_stop
+    pkill -9 -f $PM_APPLICATION_NAME
+    if [ "$?" -eq 0 ]; then
+      echo "Killed all lingering instances of the Policy Server"
+    fi
+  fi
+else
+  echo $1 $COMMAND
```
iCollin, line 141: should we keep this line?

src/appMain/daemon.sh
```diff
@@ -0,0 +1,142 @@
...
+    rm $CORE_PID_FILE
+  fi 
+  return $RESULT
+}
+
+function pm_install_dependencies() {
+  pip3 list | grep -F tornado > /dev/null
+  if [ $? -eq 1 ]; then
+    echo "Installing tornado python package"
+    sudo -H pip3 install tornado
```
LuxoftAKutsan, line 55: Would be nice to specify tornado in dependencies, but not force installation during sdl_run

Issue #3162, opened 2019-12-06
CCB: SDL Crashes while testing QT 5.12


LitvinenkoIra:
### Bug Report
SDL Crashes while testing QT 5.12

### Detailed analysis:
SDL added the new application registered to the application list but failed to update the plugin manager before "onVehicleData" notification is received from the HMI from different thread, since the application list contains extra application than to the registered app extension, it was throwing null extension ptr which while referencing caused a crash by the RPC handler thread. Worst case the delay between the application list update and extension update was about 70-100ms and before that if onvehicledata is received, then it will result in crash.

##### Reproduction Steps
1. Start SDL, HMI, connect Mobile
2. Connect iOS with SYNC via BT. Disconnect BT reconnect BT  -->  SDL Crash observed
3. Connect iOS with SYNC via USB + Android via BT – Unplug iOS --> SDL Crash observed

##### Expected Behavior
SDL should not crash

##### Observed Behavior
SDL crashed

#### OS & Version Information
* OS/Version : SYNC4
* SDL Core Version: https://github.com/smartdevicelink/sdl_core/releases/tag/5.0.1 
* Testing Against: SYNC4 Applink

E-SAITO-TMC:
@theresalech -san, We were able to reproduce this issue with Core 6.0.1. This is a high priority issue for Toyota. Could you please include this in the Core 6.2 release planning?

theresalech:
@LitvinenkoIra can you please advise if you are able to reproduce this issue in the latest Core release?

LuxoftAKutsan:
Fix from SYNC 4 is partially applicable to open sdl. Will be applied. 
Only regression will be checked because issue related to IOs devices

theresalech:
@LuxoftAKutsan, thank you for the additional information!  Please let us know when a PR is ready for Livio's review.

PR #3428, opened 2020-06-09
fix SDL Crash seen in SDL upon receiving rpc - OnEmergencyEvent with "enabled:true"

zhouxin627:
Fixes #2808 

This PR is **ready** for review.

### Risk
This PR makes **no** API changes.

### Testing Plan
Covered by manual testing

### Summary
SDLCore received `OnAppRegistered` RPC. There contain 2 steps:
1. `application_manager` add Application to the list of registered applications.
2. change Application `HMILevel` from `INVALID_ENUM` to the default value.
The issue occurs time is between step 1 and step 2, HMI send `OnEmergencyEvent(true)` RPC to SDLCore at this time, which cause Application new `HMILevel` change 
to `INVALID_ENUM`，then SDL crashes with `DCHECK` failure in `RecordHmiStateChanged` function because new `HMILevel` is `INVALID_ENUM`. 
image
Code modification are as follows:				

- If some application‘s `HMILevel` value is `INVALID_ENUM`, SDLCore should not change `HMILevel` to any valid `HMILevel` in `hmi_state.cc`					

 (In order to avoid `HMILevel`  has incorrect change in some cases，for example, HMI sends a `OnEventChanged(DEACTIVATE_HMI)` event to SDLCore when the `HMILevel` is `INVALID_ENUM`, 						
which will cause `HMILevel` value change to `BACKGROUND`  from `INVALID_ENUM`, as result,  SDLCore will send `OnHmiStatus(BACKGROUND)` to mobileApp）						

- If some application‘s old `HMILevel` and new `HMILevel` are both `INVALID_ENUM`(Invalid value), SDLCore should not run `OnStateChanged` function to avoid `DCHECK` failure.					

### CLA
- [x] I have signed [the CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform)

PR #3430, opened 2020-06-09
Fix SIGSEGV Crash subscribed_data

### LINKED ISSUE #3172: CCB: Fix SIGSEGV Crash subscribed_data
### Bug Report
SmartDeviceLink crash with SIGSEGV

### Detailed analysis:
unprotected access to the VehicleInfoAppExtension::subscribed_data from different threads

##### Reproduction Steps
2. Register app
3. Perform multiple RPC scenarios and send OnVehicleDataNotification
4. Observe behavior

##### Expected Behavior
SDL should not crash

##### Observed Behavior
SDL crashed
[bt.txt](https://github.com/smartdevicelink/sdl_core/files/3932241/bt.txt)

#### OS & Version Information
* OS/Version : SYNC4

zhouxin627:
Fixes #3172 #3162 

### Risk

### Testing Plan
Covered by manual testing

### Summary
There is a time interval between registering an application and adding `VehicleInfoAppExtension`,
and then registering a new App when you already subscribe to the Vehicle and have an information notification, because the App object for `QueryInterface()` already exists, but `VehicleInfoAppExtension` is not Added yet. So we add judgment protection for Invalid object usage. 
image

### CLA

### DIFF
--- a/src/components/application_manager/rpc_plugins/vehicle_info_plugin/src/commands/mobile/on_vehicle_data_notification.cc
+++ b/src/components/application_manager/rpc_plugins/vehicle_info_plugin/src/commands/mobile/on_vehicle_data_notification.cc
@@ -79,8 +79,7 @@ void OnVehicleDataNotification::Run() {
       auto& ext = VehicleInfoAppExtension::ExtractVIExtension(*app);
       // Fix for Invalid object usage
       if (nullptr == &ext) {
-        LOG4CXX_ERROR(
-              logger_, "ExtractVIExtension is nullptr ");
+        LOG4CXX_ERROR(logger_, "ExtractVIExtension is nullptr");
         return false;
       }
       return ext.isSubscribedToVehicleInfo(name);
--- a/src/components/application_manager/rpc_plugins/vehicle_info_plugin/src/commands/mobile/on_vehicle_data_notification.cc
+++ b/src/components/application_manager/rpc_plugins/vehicle_info_plugin/src/commands/mobile/on_vehicle_data_notification.cc
@@ -77,6 +77,12 @@ void OnVehicleDataNotification::Run() {
     auto subscribed_to_ivi_predicate = [&name](const ApplicationSharedPtr app) {
       DCHECK_OR_RETURN(app, false);
       auto& ext = VehicleInfoAppExtension::ExtractVIExtension(*app);
+      // Fix for Invalid object usage
+      if (nullptr == &ext) {
+        LOG4CXX_ERROR(
+              logger_, "ExtractVIExtension is nullptr ");
+        return false;
+      }
       return ext.isSubscribedToVehicleInfo(name);
     };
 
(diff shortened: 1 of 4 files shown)


LuxoftAKutsan:
Looks like  that automated script that can reproduce the issue with some probability can be created .

zhouxin627:
Hi @LuxoftAKutsan, thank you for your comment. We also considered to register the app first and then access the vehicle data of the app. The method is `RegisterApplication`、`AddExtension` first, and then `QueryInterface`, but the changing of timing sequence has a great impact on the existing design. Therefore, using NULL object protection can also solve the crash problem. And It is also correct not to send `OnVehicleData` notification when the registration process is not over. It is a failsafe scene. It also can be covered by the current modification scheme. Based on the above points, we believe that temporary solutions can solve this problem.

zhouxin627:
Hi @LuxoftAKutsan, do you have a plan of making a PR to fix this issue.

LuxoftAKutsan:
Possible fix  for https://github.com/smartdevicelink/sdl_core/issues/3162 
is available in SYNC4
Unfortunately it is not the approach that I was described in my previous comment.
Not sure if this changes will come into 7.0 .
Here is the diff 

[diff.txt](https://github.com/smartdevicelink/sdl_core/files/5097847/diff.txt)

IGapchuk:
Hi @zhouxin627 ! We believe that https://github.com/smartdevicelink/sdl_core/pull/3430#issuecomment-676530924 is not full fix diff and we can provide the additional diff with changes for safely work with AppExtensions. Please, take a look at attached
[bugfix_3172_3162.txt](https://github.com/smartdevicelink/sdl_core/files/5246340/bugfix_3172_3162.txt). We believe it will be helpful.
If you have some additional questions or suggestions feel free to share them with us. Thank you!

JackLivio:
@zhouxin627 @LuxoftAKutsan 

I am working on a fix to resolve this issue. I have reordered the app registration process slightly to add the plugins earlier to an app instance. The app will not be accessible by the applications accessor until all of the plugins have been added. This should prevent the crashes seen by missing app on the plugin level.

https://github.com/smartdevicelink/sdl_core/pull/3526/files

Separate note:

I attempted to resolve the issue seen here: https://github.com/smartdevicelink/sdl_core/pull/3428

But INVALID_ENUM plays an important roll in the app resumption process, especially with processing widget windows during resumption. The fix became to messy and i offered a simpler solution for that crash in the other pr.

LuxoftAKutsan:
Looks like Jacks #3526 should fix this particular issue. 

But the root cause is more global and in the future, we will work on better documentation and reworking application registration.

The main problem is that there are many activities that need to be done during app registration (Resumption, update policy, add to plugins, to applicaiton_manager,  HMI notification, etc ... ). 
Some of this activities may prevent registration, some of them need to be done after another. 
We do not have a picture of the application registration sequence. 

We need to create appropriate documentation with list of activities that need to be done on app registration, there order, and dependencies.
Create separate events for plugins to create app extension and to notify a plugin that app is finally registered. 
Make an application accessible for external events only after the full registration cycle will be finished.

JackLivio:
Closing in favor of #3526

PR #3447, opened 2020-06-18
Additional Submenus and Driver Distraction Limits

### LINKED ISSUE #2084: [SDL 0148] Template Improvements: Additional SubMenus
Proposal: [Template Improvements: Additional SubMenus](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0148-template-additional-submenus.md)

> Currently, SDL is by far the most restrictive among its competitors for drilling down into menus. It only allows the top menu and one submenu (additional submenus can be "faked" using Choice Sets). CarPlay, for example, allows at least 3 levels of drilling, this was the most I could find in the apps I use (in addition to tabs at the top, which is a kind-of 4th level). CarPlay allows as much drilling as desired by the user in the car, but prevents scrolling while moving. For future updates, we should allow further menu depth. For driver distraction reasons, a later proposal will allow restriction of menu depth while driving to replicate current restrictions if desired.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/423

Steering Committee Decision:

> The Steering Committee has voted to accept this proposal with the following revisions: update proposal to only include details regarding "Allowing Submenu's with parentID's that refer another Submenu." A separate proposal will be entered to discuss "Allowing Core to define driver distraction parameters for HMI."

>When discussing submenus, a few additional points were made:

>Right now, this is only considered for submenus, but driver distraction guidelines could be extended to include choice sets for interactions as well (including list length)
Proposal outlines the same behavior as is currently in place for submenus, but extends to allow a submenu to have a submenu

>The proposal .md was updated on March 22, 2018 to include the agreed upon revisions.

JackLivio:
Fixes #2084 & #2123

This PR is **not** for review.

### Risk

### Testing Plan
- Test AddSubMenu with parent ids at multiple levels
- Test submenudepth and command list length with driver distraction on/off

### Summary
The implementations of proposals 0148-template-additional-submenus and 0152-driver-distraction-list-limits are co-dependent and need to be implemented together.

This pr is considered a menu overhaul for the 6.2 release.

### CLA


src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/add_sub_menu_request.cc
```diff
@@ -130,6 +132,9 @@ void AddSubMenuRequest::Run() {
   }
   msg_params[strings::menu_params][strings::menu_name] =
       received_msg_params[strings::menu_name];
+  msg_params[strings::menu_params][strings::parent_id] =
+      received_msg_params[strings::parent_id];
```
jacobkeeler, line 136:
```suggestion
  msg_params[strings::menu_params][strings::parent_id] = parent_id;
```

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/get_system_capability_request.cc
```diff
@@ -163,6 +163,17 @@ void GetSystemCapabilityRequest::Run() {
                              all_services);
       break;
     }
+    case mobile_apis::SystemCapabilityType::DRIVER_DISTRACTION:
+      if (hmi_capabilities.driver_distraction_capability() &&
+          hmi_capabilities.driver_distraction_supported()) {
```
jacobkeeler, line 168: Seems checking `driver_distraction_supported` might be redundant here. This value should be used in the RAI response (see https://github.com/smartdevicelink/rpc_spec/pull/263/files#r453856989)

src/components/application_manager/src/application_data_impl.cc
```diff
@@ -855,12 +855,13 @@ smart_objects::SmartObject* DynamicApplicationDataImpl::FindSubMenu(
 }
 
 bool DynamicApplicationDataImpl::IsSubMenuNameAlreadyExist(
-    const std::string& name) {
+    const std::string& name, const int32_t parent_id) {
```
jacobkeeler, line 858:
```suggestion
    const std::string& name, const uint32_t parent_id) {
```

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/add_sub_menu_request.cc
```diff
@@ -94,7 +94,9 @@ void AddSubMenuRequest::Run() {
   const std::string& menu_name =
       received_msg_params[strings::menu_name].asString();
 
-  if (app->IsSubMenuNameAlreadyExist(menu_name)) {
+  const int32_t parent_id = received_msg_params[strings::parent_id].asInt();
```
jacobkeeler, line 97:
```suggestion
  const uint32_t parent_id = received_msg_params[strings::parent_id].asUInt();
```
Also, you should check if this value exists, otherwise a bad value is likely returned (this would only affect the case where an older proxy is sending the message, otherwise the value would be filled with the default value of 0).

src/components/application_manager/include/application_manager/application_data_impl.h
```diff
@@ -330,7 +330,7 @@ class DynamicApplicationDataImpl : public virtual Application {
   CommandsMap commands_;
   mutable std::shared_ptr<sync_primitives::RecursiveLock> commands_lock_ptr_;
   SubMenuMap sub_menu_;
-  mutable std::shared_ptr<sync_primitives::Lock> sub_menu_lock_ptr_;
+  mutable std::shared_ptr<sync_primitives::RecursiveLock> sub_menu_lock_ptr_;
```
jacobkeeler, line 333: It looks to me that this change is not necessary since the data accessor is only used at the root level.

src/components/interfaces/HMI_API.xml
```diff
@@ -2817,7 +2817,7 @@
 <struct name="MenuParams">
   <param name="parentID" type="Integer" minvalue="0" maxvalue="2000000000" mandatory="false">
     <description>unique ID of the sub menu, the command must be added to.</description>
-    <description>If not provided, the command must be added to the top level of the in application menu.</description>
+    <description>If not provided or is 0, the command must be added to the top level of the in application menu.</description>
```
jacobkeeler, line 2820:
```suggestion
    <description>If not provided or 0, the command must be added to the top level of the in application menu.</description>
```
Minor suggestion

jacobkeeler:
`sdl_commands_test_unity` fails on my machine with this PR, please verify on your end

jacobkeeler:
Also fix merge conflicts

src/components/application_manager/include/application_manager/application_data_impl.h
JackLivio: I believe it is necessary, i got a deadlock without that change on this [line](https://github.com/smartdevicelink/sdl_core/pull/3447/files/1e59f31353831c4316ed3a70c161ac5611e47a12..6923723764dfc3514e1305a73972c0121e6bef97#diff-5ab8eae8ebc07c54021736d8a81015a8R118)

The first lock taken is [here](https://github.com/smartdevicelink/sdl_core/pull/3447/files/1e59f31353831c4316ed3a70c161ac5611e47a12..6923723764dfc3514e1305a73972c0121e6bef97#diff-5ab8eae8ebc07c54021736d8a81015a8R214).

Note that the commands accessor has a recursive lock for the same type of usage [here](https://github.com/smartdevicelink/sdl_core/pull/3447/files/1e59f31353831c4316ed3a70c161ac5611e47a12..6923723764dfc3514e1305a73972c0121e6bef97#diff-5ab8eae8ebc07c54021736d8a81015a8R175)

jacobkeeler: Ah, I see I didn't notice that that the lock was taken in the `app->RemoveSubMenu` call

JackLivio:
@jacobkeeler Merge conflicts and unit tests fixed

PR #3454, opened 2020-06-26
Feature/Supported Character Sets

ShobhitAd:
Implements [#2788](https://github.com/smartdevicelink/sdl_core/issues/2788)

### Risk

### Testing Plan
- ATF Tests
- Manual Tests using Test suite

### Summary
[Summary of PR changes]

### CLA


src/appMain/hmi_capabilities.json
```diff
@@ -10,169 +10,169 @@
             "displayName": "GENERIC_DISPLAY",
             "textFields": [{
                     "name": "mainField1",
-                    "characterSet": "TYPE2SET",
+                    "characterSet": "UTF_8",
```
iCollin, line 13: Should this just be ASCII since "All systems will support at least ASCII"

src/components/interfaces/HMI_API.xml
```diff
@@ -1432,16 +1432,25 @@
 <enum name="CharacterSet">
   <description>The list of potential character sets</description>
   <element name="TYPE2SET">
-    <description>See [@TODO: create file ref]</description>
+    <description>See [@TODO: create file ref]. Note that this element is deprecated in the MOBILE API</description>
```
iCollin, line 1435: same question from MOBILE_API, why keep `See [@TODO: create file ref]. `

src/appMain/hmi_capabilities.json
ShobhitAd: From the proposal https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0202-character-sets.md
> These character sets were chosen because they are the most widely supported character sets and are supported on all Android and iOS devices. As noted in the document, UTF-8 is the preferred character set.

Also, I think setting `UTF_8` as the default would allow us to show the full capabilities as opposed to just using the minimum supported character set `ASCII`.

src/components/interfaces/HMI_API.xml
ShobhitAd: Removed in https://github.com/smartdevicelink/sdl_core/pull/3454/commits/e630938b2b5a012d0c9f8cb9fa4f73a7e7d4a198

PR #3462, opened 2020-07-20
Feature/Add reason param to All protocol NAKs

### LINKED ISSUE #3434: [SDL 0308] Add a Reason Parameter to All Protocol NAKs
Proposal: [Add a Reason Parameter to All Protocol NAKs](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0308-protocol-nak-reason.md)

> As a maintainer of an app library, it is often difficult to understand why Core has NAKed a protocol message. Most NAKs currently have a `rejectedParams` parameter, but in practice, this has not been enough information to determine why a NAK has occurred and properly alert the developer. In place of the `rejectedParams` parameter, the `RegisterSecondaryTransportNAK` has a `reason` parameter to describe the failure reason. This is a better approach and we should expand it to all NAK parameters.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/1027

ShobhitAd:
Fixes #3434 

### Risk

### Testing Plan
- ATF
- Manual testing using test_suite

### Summary
- Add reason param to the StartService and EndService NAK payloads
- Add a reason string for all NAK cases in SDL
- Bump protocol version to `5.3.0`

### CLA


src/components/protocol_handler/src/protocol_handler_impl.cc
```diff
@@ -488,20 +492,36 @@ void ProtocolHandlerImpl::SendStartSessionNAck(
 
   uint8_t maxProtocolVersion = SupportedSDLProtocolVersion();
 
+  utils::SemanticVersion full_version;
+  session_observer_.ProtocolVersionUsed(
```
iCollin, line 496: are we sure the connection is valid when this code is run? we don't need to check the return value of ProtocolVersionUsed? If an app sends EndSession and stops running, maybe it's possible the connection could be removed by now.

src/components/protocol_handler/src/handshake_handler.cc
```diff
@@ -242,10 +256,28 @@ void HandshakeHandler::ProcessFailedHandshake(BsonObject& params,
...
+        case ServiceStatus::PTU_FAILED:
+          return "Policy Table Update failed";
+        case ServiceStatus::CERT_INVALID:
+          return "Invalid certificate";
+        case ServiceStatus::INVALID_TIME:
+          return "Failed to get system time";
+        case ServiceStatus::UNSECURE_START_FAILED:
+          return "Unsecure start failed";
+        default:
+          return "Unknown cause of failure";
```
iCollin, line 272: I don't see any references to `ProcessFailedHandshake` with any service status besides the top three (not even UNSECURE_START_FAILED)

src/components/protocol_handler/src/protocol_handler_impl.cc
```diff
@@ -1596,7 +1638,9 @@ RESULT_CODE ProtocolHandlerImpl::HandleControlMessageEndSession(
                          current_session_id,
                          packet.protocol_version(),
                          service_type,
-                         rejectedParams);
+                         rejectedParams,
+                         "Refused to end session for service of type " +
```
iCollin, line 1642: `OnSessionEndedCallback` has some descriptive logs in it with its error returns, do you think this reason param is enough for developers, or should some message be passed up from `OnSessionEndedCallback` or even deeper?

src/components/protocol_handler/src/handshake_handler.cc
```diff
@@ -202,10 +202,24 @@ void HandshakeHandler::ProcessSuccessfulHandshake(const uint32_t connection_key,
...
-                                           context_.new_session_id_,
-                                           protocol_version_,
-                                           context_.service_type_);
+
+    auto gen_reason_msg = [](const bool can_be_protected,
+                             const bool is_already_protected) -> std::string {
+      if (can_be_protected && is_already_protected) {
+        return "Service is already protected";
+      } else if (!can_be_protected && is_already_protected) {
+        return "Service can be protected but is not already protected";
```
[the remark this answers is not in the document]
ShobhitAd: Good catch, fixed in https://github.com/smartdevicelink/sdl_core/pull/3462/commits/c85e87f6e19357f22feba51f0732a4238812ff60

src/components/protocol_handler/src/handshake_handler.cc
```diff
@@ -202,10 +202,24 @@ void HandshakeHandler::ProcessSuccessfulHandshake(const uint32_t connection_key,
...
+      } else if (!can_be_protected && !is_already_protected) {
+        return "Service cannot be protected";
+      }
+      return std::string();
```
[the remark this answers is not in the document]
ShobhitAd: Fixed in https://github.com/smartdevicelink/sdl_core/pull/3462/commits/c85e87f6e19357f22feba51f0732a4238812ff60

src/components/protocol_handler/src/handshake_handler.cc
```diff
@@ -242,10 +256,28 @@ void HandshakeHandler::ProcessFailedHandshake(BsonObject& params,
   } else {
     service_status_update_handler_.OnServiceUpdate(
         this->connection_key(), context_.service_type_, service_status);
+
+    auto gen_reason_msg =
```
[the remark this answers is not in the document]
ShobhitAd: Fixed in https://github.com/smartdevicelink/sdl_core/pull/3462/commits/c85e87f6e19357f22feba51f0732a4238812ff60

src/components/protocol_handler/src/handshake_handler.cc
ShobhitAd: Yes, its for future proofing. I think even if we add a call for `ProcessFailedHandshake` for `UNSECURED_START_FAILED` in the future, updating the reason_msg is change which might be easily missed. Let me know what you think.

src/components/protocol_handler/src/protocol_handler_impl.cc
```diff
@@ -61,9 +61,10 @@ std::string ConvertPacketDataToString(const uint8_t* data,
 
 const size_t kStackSize = 131072;
 
-utils::SemanticVersion default_protocol_version(5, 2, 0);
+utils::SemanticVersion default_protocol_version(5, 3, 0);
 utils::SemanticVersion min_multiple_transports_version(5, 1, 0);
 utils::SemanticVersion min_cloud_app_version(5, 2, 0);
+utils::SemanticVersion min_reason_param_version(5, 3, 0);
```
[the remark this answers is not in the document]
ShobhitAd: I've changed these version variables to `const` in https://github.com/smartdevicelink/sdl_core/pull/3462/commits/c85e87f6e19357f22feba51f0732a4238812ff60 but I'm not sure if `constexpr` would work here(since `utils::SemanticVersion` doesn't seem to be a literal type).

src/components/protocol_handler/include/protocol_handler/protocol_handler_impl.h
```diff
@@ -420,7 +422,8 @@ class ProtocolHandlerImpl
   void SendEndSessionNAck(ConnectionID connection_id,
                           uint32_t session_id,
                           uint8_t protocol_version,
-                          uint8_t service_type);
+                          uint8_t service_type,
+                          const std::string reason = std::string());
```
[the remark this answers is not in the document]
ShobhitAd: Made reason a required param in https://github.com/smartdevicelink/sdl_core/pull/3462/commits/6d9f2a85a5689e792cd1729f4ccb83f394f62845

src/components/protocol_handler/src/protocol_handler_impl.cc
ShobhitAd: Good call. Checks added in https://github.com/smartdevicelink/sdl_core/pull/3462/commits/c785dfa7661aba9511513c072b9ce8ef26ef1cbc

src/components/protocol_handler/src/protocol_handler_impl.cc
```diff
@@ -488,20 +492,36 @@ void ProtocolHandlerImpl::SendStartSessionNAck(
...
+      for (std::string param : rejectedParams) {
+        char paramPtr[256];
+        strncpy(paramPtr, param.c_str(), sizeof(paramPtr));
+        paramPtr[sizeof(paramPtr) - 1] = '\0';
+        bson_array_add_string(&rejectedParamsArr, paramPtr);
+      }
+      bson_object_put_array(
+          &payloadObj, strings::rejected_params, &rejectedParamsArr);
+    }
+    if (!reason.empty() && min_version >= min_reason_param_version) {
```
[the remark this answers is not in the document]
ShobhitAd: I wanted to mirror the same approach which is used by `StartSessionAck` https://github.com/smartdevicelink/sdl_core/blob/master/src/components/protocol_handler/src/protocol_handler_impl.cc#L313 but it looks like `StartSessionAck` has a few more major version checks which will prevent the bson payload from being added. 

I've added changes in https://github.com/smartdevicelink/sdl_core/pull/3462/commits/e8fa2afa899678b43c2bb64e51f9abcd5114ee19  to use the `full_version` instead of calculating the `min_version`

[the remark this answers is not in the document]
iCollin: Do you know the reason for this min version logic though? Maybe it is intended behavior that we use the latest protocol version if they don't report a valid version, and this line should use min_version.

iCollin:
Hey @ShobhitAd, updates look good for the most part. I left one new comment, and also one more reply [here](https://github.com/smartdevicelink/sdl_core/pull/3462#discussion_r472276512). I also see some unit tests failing:
```
[  FAILED  ] 6 tests, listed below:
[  FAILED  ] ProtocolHandlerImplTest.StartSession_Unprotected_SessionObserverReject
[  FAILED  ] ProtocolHandlerImplTest.StartSession_Protected_SessionObserverReject
[  FAILED  ] ProtocolHandlerImplTest.StartSession_Unprotected_Multiple_SessionObserverAcceptAndReject
[  FAILED  ] ProtocolHandlerImplTest.StartSession_Audio_RejectByTransportType
[  FAILED  ] ProtocolHandlerImplTest.StartSession_Video_RejectByTransportType
[  FAILED  ] ProtocolHandlerImplTest.EndSession_SessionObserverReject
```

src/components/protocol_handler/src/handshake_handler.cc
ShobhitAd: Although it looks like this case would not be handled in the `hanshake_handler` so I've removed `UNSECURE_START_FAILED` in https://github.com/smartdevicelink/sdl_core/pull/3462/commits/c289ecf066bafed2451d064f0ed4e1799b968705

src/components/protocol_handler/src/protocol_handler_impl.cc
ShobhitAd: Implemented getting the failure reason from `OnSessionEndedCallback` in https://github.com/smartdevicelink/sdl_core/pull/3462/commits/4c55f0b4aba9f5887d71dbcc36423967e550d736

src/components/protocol_handler/src/handshake_handler.cc
```diff
@@ -208,11 +208,10 @@ void HandshakeHandler::ProcessSuccessfulHandshake(const uint32_t connection_key,
       if (can_be_protected && is_already_protected) {
         return "Service is already protected";
       } else if (!can_be_protected && is_already_protected) {
-        return "Service can be protected but is not already protected";
-      } else if (!can_be_protected && !is_already_protected) {
+        return "Service cannot be protected but is already protected";
```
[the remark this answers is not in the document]
ShobhitAd: hmm.. you're right, it doesn't seem possible for a service to be protected if it is part of the force_unprotected_services in the settings. Removed the condition in https://github.com/smartdevicelink/sdl_core/pull/3462/commits/350a638867c085a0719d54ee099452c99470b029

src/components/protocol_handler/src/handshake_handler.cc
```diff
@@ -203,22 +203,14 @@ void HandshakeHandler::ProcessSuccessfulHandshake(const uint32_t connection_key,
...
-        return "Service cannot be protected";
-      }
-    };
     protocol_handler_.SendStartSessionNAck(
         context_.connection_id_,
         context_.new_session_id_,
         protocol_version_,
         context_.service_type_,
-        gen_reason_msg(can_be_protected, is_service_already_protected));
+        (can_be_protected && is_service_already_protected)
```
iCollin, line 211: well this is getting much simpler. Do we even need to check can_be_protected? It seems that is implied by `is_service_already_protected`

src/components/protocol_handler/test/protocol_handler_tm_test.cc
```diff
@@ -906,12 +909,6 @@ TEST_F(ProtocolHandlerImplTest,
   bson_object_initialize_default(&bson_nack_params);
   bson_object_put_array(
       &bson_nack_params, protocol_handler::strings::rejected_params, &bson_arr);
-  std::string reason =
```
iCollin: is it possible to put an updated string here instead of removing this part?

JackLivio (approved):

PR #3057, opened 2019-09-30
[WIP] initial fix of socket streaming by guessing playback time

iCollin:
Partially Fixes #2945 

This PR is **ready** for review.

### Risk
This PR makes **no** API changes.

### Testing Plan
ATF, Manually

### Summary
Add logic for Core to guess how long an audio stream will take based on PCM capabilities.

##### Enhancements
* Socket streaming somewhat works

### Tasks Remaining:
- [x] Decide how much latency compensation should be added in DataSizeToMilliseconds
- [x] Restrict logic in this PR to only applicable scenarios (streaming is socket, maybe just for audio)
- [x] Test a lot more!
- [x] Make sure it builds with tests, add mock functions!

### CLA
- [x] I have signed [the CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform)

PR #3488, opened 2020-08-31
Fix streamer activity in case of suspend

### LINKED ISSUE #3160: CCB: Fix streamer activity in case of suspend
### Bug Report
SDL stops streaming data by timeout when no new data from mobile received

### Detailed analysis:
The problem of the current implementation is that SDL kills streaming thread responsible for sending a/v streaming data to HMI when streaming timeout expires.
This issue is observed when mobile app dumps 10-seconds audio file to SDL during 2 seconds and after that app does not send any data. In that case, HMI will play audio file during 2 seconds + timeout = 5 seconds. At the 5th second SDL just kills streaming thread with all pending messages, however, audio service is still open. As a result, not a whole audio file is played.

The correct behavior from SDL side in that case is not kill streaming thread when streaming timeout was expired. SDL should kill streaming thread only when service is
actually stopped.

##### Reproduction Steps
1. Start SDL and modified HMI
2. Register SPT with type Navi over TCP transport
3. Activate that app
4. In SPT go to Menu options on the mobile device --> AV Streaming -->Under Audio Streaming Encrypt Video to NONE --> Start session --> Start Audio
5. Select 30 sec file and send it to HMI
6. Wait until the whole audio file is played

##### Expected Behavior
Should be played whole audio file

##### Observed Behavior
Audio is played just a few seconds

#### OS & Version Information
* OS/Version : SYNC4

AKalinich-Luxoft:
Fixes #3160 

### Risk

### Testing Plan
[Describe how you plan to unit test the changes in this PR]

### Summary
The problem of the current implementation is that SDL kills streaming thread responsible for sending a/v streaming data to HMI when streaming timeout expires. This issue is observed when mobile app dumps 10-seconds audio file to SDL during 2 seconds and after that app does not send any data. In that case HMI will play audio file during 2 seconds + timeout = 5 seconds. At the 5th second SDL just kills streaming thread with all pending messages, however audio service is still open. As a result not a whole audio file is played.

The correct behavior from SDL side in that case is not kill streaming thread when streaming timeout was expired. SDL should kill streaming thread only when service is actually stopped.

Current SDL behavior was updated to align with a correct behavior described above.

### Tasks Remaining:
- [x] Test this fix

### CLA

### DIFF
--- a/src/components/application_manager/src/application_impl.cc
+++ b/src/components/application_manager/src/application_impl.cc
@@ -659,10 +659,9 @@ void ApplicationImpl::SuspendStreaming(
 }
 
 void ApplicationImpl::WakeUpStreaming(
-    protocol_handler::ServiceType service_type, uint32_t timer_len) {
+    protocol_handler::ServiceType service_type) {
   using namespace protocol_handler;
   SDL_LOG_AUTO_TRACE();
-  const bool is_custom_timeout = 0 != timer_len;
 
   // See the comment in StopStreaming(). Also, please make sure that we acquire
   // streaming_stop_lock_ then xxx_streaming_suspended_lock_ in this order!
@@ -680,13 +679,8 @@ void ApplicationImpl::WakeUpStreaming(
       }
     }
 
-    if (is_custom_timeout) {
-      SDL_LOG_DEBUG("New estimated video playback time is: " << timer_len);
-    }
-
-    video_stream_suspend_timer_.Start(
-        is_custom_timeout ? timer_len : video_stream_suspend_timeout_,
-        timer::kPeriodic);
+    video_stream_suspend_timer_.Start(video_stream_suspend_timeout_,
+                                      timer::kPeriodic);
   } else if (ServiceType::kAudio == service_type) {
     {  // reduce the range of audio_streaming_suspended_lock_
       sync_primitives::AutoLock lock(audio_streaming_suspended_lock_);
@@ -698,14 +692,8 @@ void ApplicationImpl::WakeUpStreaming(
         audio_streaming_suspended_ = false;
       }
     }
-
-    if (is_custom_timeout) {
-      SDL_LOG_DEBUG("New estimated audio playback time is: " << timer_len);
-    }
-
-    audio_stream_suspend_timer_.Start(
-        is_custom_timeout ? timer_len : audio_stream_suspend_timeout_,
-        timer::kPeriodic);
+    audio_stream_suspend_timer_.Start(audio_stream_suspend_timeout_,
+                                      timer::kPeriodic);
   }
 }
 
--- a/src/components/application_manager/src/application_impl.cc
+++ b/src/components/application_manager/src/application_impl.cc
@@ -615,6 +615,9 @@ void ApplicationImpl::StopStreaming(
 void ApplicationImpl::StopNaviStreaming() {
   SDL_LOG_AUTO_TRACE();
   video_stream_suspend_timer_.Stop();
+  application_manager_.OnAppStreaming(app_id(),
+                                      protocol_handler::ServiceType::kMobileNav,
+                                      StreamingState::kStopped);
   MessageHelper::SendNaviStopStream(app_id(), application_manager_);
   set_video_streaming_approved(false);
   set_video_stream_retry_number(0);
@@ -623,6 +626,9 @@ void ApplicationImpl::StopNaviStreaming() {
 void ApplicationImpl::StopAudioStreaming() {
   SDL_LOG_AUTO_TRACE();
   audio_stream_suspend_timer_.Stop();
+  application_manager_.OnAppStreaming(app_id(),
+                                      protocol_handler::ServiceType::kAudio,
+                                      StreamingState::kStopped);
   MessageHelper::SendAudioStopStream(app_id(), application_manager_);
   set_audio_streaming_approved(false);
   set_audio_stream_retry_number(0);
@@ -633,14 +639,17 @@ void ApplicationImpl::SuspendStreaming(
   using namespace protocol_handler;
   SDL_LOG_AUTO_TRACE();
 
-  if (ServiceType::kMobileNav == service_type) {
+  if (ServiceType::kMobileNav == service_type && !video_streaming_suspended_) {
     video_stream_suspend_timer_.Stop();
-    application_manager_.OnAppStreaming(app_id(), service_type, false);
+    application_manager_.OnAppStreaming(
+        app_id(), service_type, StreamingState::kSuspended);
     sync_primitives::AutoLock lock(video_streaming_suspended_lock_);
     video_streaming_suspended_ = true;
-  } else if (ServiceType::kAudio == service_type) {
+  } else if (ServiceType::kAudio == service_type &&
+             !audio_streaming_suspended_) {
     audio_stream_suspend_timer_.Stop();
-    application_manager_.OnAppStreaming(app_id(), service_type, false);
+    application_manager_.OnAppStreaming(
+        app_id(), service_type, StreamingState::kSuspended);
     sync_primitives::AutoLock lock(audio_streaming_suspended_lock_);
     audio_streaming_suspended_ = true;
   }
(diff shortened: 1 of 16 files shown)


LuxoftAKutsan (approved):

src/components/application_manager/src/application_impl.cc
```diff
@@ -661,25 +670,37 @@ void ApplicationImpl::WakeUpStreaming(
...
+        application_manager_.OnAppStreaming(
+            app_id(), service_type, StreamingState::kStarted);
         application_manager_.ProcessOnDataStreamingNotification(
             service_type, app_id(), true);
         video_streaming_suspended_ = false;
       }
     }
+
+    if (timer_len != 0) {
+      SDL_LOG_DEBUG("New estimated video playback time is: " << timer_len);
```
SNiukalov, line 682: @AKalinich-Luxoft 
Please change order comparing:
0 == timer_len

src/components/application_manager/src/application_impl.cc
```diff
@@ -661,25 +670,37 @@ void ApplicationImpl::WakeUpStreaming(
...
+    }
+
     video_stream_suspend_timer_.Start(
         timer_len == 0 ? video_stream_suspend_timeout_ : timer_len,
         timer::kPeriodic);
```
SNiukalov, line 687: @AKalinich-Luxoft 
How above

src/components/application_manager/src/application_impl.cc
```diff
@@ -661,25 +670,37 @@ void ApplicationImpl::WakeUpStreaming(
...
+        application_manager_.OnAppStreaming(
+            app_id(), service_type, StreamingState::kStarted);
         application_manager_.ProcessOnDataStreamingNotification(
             service_type, app_id(), true);
         audio_streaming_suspended_ = false;
       }
     }
+
+    if (timer_len != 0) {
+      SDL_LOG_DEBUG("New estimated audio playback time is: " << timer_len);
```
SNiukalov, line 701: @AKalinich-Luxoft 
How above

src/components/media_manager/src/media_manager_impl.cc
```diff
@@ -298,13 +298,29 @@ void MediaManagerImpl::StopStreaming(
...
 }
 
+void MediaManagerImpl::SuspendStreaming(
+    int32_t application_key, protocol_handler::ServiceType service_type) {
+  SDL_LOG_AUTO_TRACE();
+  using namespace application_manager;
+  using namespace protocol_handler;
+
+  ApplicationSharedPtr app = application_manager_.application(application_key);
+  if (app) {
```
SNiukalov, line 313: @AKalinich-Luxoft 
Maybe const auto app

SNiukalov (approved):

src/components/application_manager/src/application_impl.cc
AKalinich-Luxoft: @SNiukalov fixed in ce5bce1528d56f60e65492288d1f6d1c1f1b9372

src/components/application_manager/src/application_impl.cc
AKalinich-Luxoft: @SNiukalov fixed in ce5bce1528d56f60e65492288d1f6d1c1f1b9372

src/components/application_manager/src/application_impl.cc
AKalinich-Luxoft: @SNiukalov fixed in ce5bce1528d56f60e65492288d1f6d1c1f1b9372

src/components/media_manager/src/media_manager_impl.cc
AKalinich-Luxoft: @SNiukalov fixed in ce5bce1528d56f60e65492288d1f6d1c1f1b9372

ychernysheva (approved):

src/components/media_manager/src/media_manager_impl.cc
```diff
@@ -298,13 +298,29 @@ void MediaManagerImpl::StopStreaming(
...
+  using namespace application_manager;
+  using namespace protocol_handler;
+
+  const auto app = application_manager_.application(application_key);
+  if (app) {
+    // Reset playback time and stream size when audio file has been played
+    if (ServiceType::kAudio == service_type &&
+        "socket" == settings().audio_server_type()) {
+      stream_data_size_ = 0ull;
+      socket_audio_stream_start_time_ =
```
iCollin, line 318: when streaming is suspended, i thought there could still be unstreamed data on the core side. why would you zero `stream_data_size_` here? also since you are suspending the stream, i don't see how it makes sense to set `socket_audio_stream_start_time_` here.

perhaps i am misunderstanding, when is this method `MediaManagerImpl::SuspendStreaming` called?

iCollin (commented):
The code changes only seem to affect the stream time of a stream that was suspended at some point. Am I correct that stream timeouts are not changed by this proposal unless a stream was suspended? I am confused because the issue and PR description seem to suggest otherwise, the examples and reproduction steps do not have any suspend in them, but the titles of both sound to match the code.

src/components/media_manager/src/media_manager_impl.cc
AKalinich-Luxoft: @iCollin `stream_data_size_` should be set to zero here in order to properly calculate a new timeout for SuspendStreaming. This is required only for audio streaming and only for socket mode previously done in #3064 
Without that change, we had a failure of ATF script which verifies that timeout was calculated properly. As I remember there was a case when there was streaming few seconds, then 10 seconds wait and then proceed streaming. Because streaming was suspended and them resumed again, `socket_audio_stream_start_time_` was not refreshed properly and timeout time was wrong.
Actually, it is a code taken from `StopStreaming()` before my changes. `socket_audio_stream_start_time_` is just set to zero for completeness, however this is not necessary.

AKalinich-Luxoft: @iCollin you are right, there might be an unstreamed data on the core side when stream suspension is happening. However, `stream_data_size_` is required only for calculation of the next streaming timeout so it is safe to reset it when the actual timeout has happened.

AKalinich-Luxoft:
> The code changes only seem to affect the stream time of a stream that was suspended at some point. Am I correct that stream timeouts are not changed by this proposal unless a stream was suspended? I am confused because the issue and PR description seem to suggest otherwise, the examples and reproduction steps do not have any suspend in them, but the titles of both sound to match the code.

@iCollin you are right, the code changes does not change any streaming timeouts, but just updates SDL behavior little bit. This PR provides a new state for a streaming - `kStopped`, `kStarted`, `kSuspended` (was only stopped and started). This new state allows us to keep `messages_` and streaming thread alive when streaming timeout has occurred (it was force stopped before these changes). That's why this new streaming state allows us to fix the issue described in PR.

AKalinich-Luxoft:
@iCollin I did a few things here:
1. `develop` branch has been merged into this PR (https://github.com/smartdevicelink/sdl_core/pull/3488/commits/f29a3e722b8c2192a518c4a8a00e8a6df787f2e8)
2. Changes from #3057 has been reverted in https://github.com/smartdevicelink/sdl_core/pull/3488/commits/3fd087e06339ae2b971d5220581f8015e0fd2ecd
3. Unused code after revert has been removed in https://github.com/smartdevicelink/sdl_core/pull/3488/commits/e6abe62a639ecda3b30738c4214bdde240617e5a

Please double check this PR one more time

src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -3577,12 +3579,32 @@ void ApplicationManagerImpl::OnAppStreaming(
...
+      // Apply temporary streaming state and start activity in media_manager_
+      state_ctrl_.OnVideoStreamingStarted(app);
+      media_manager_->StartStreaming(app_id, service_type);
+      break;
+    }
+
+    case Application::StreamingState::kSuspended: {
+      // Don't stop activity in media_manager_ in that case
+      // Just cancel the temporary streaming state
+      state_ctrl_.OnVideoStreamingStopped(app);
```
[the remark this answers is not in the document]
AKalinich-Luxoft: @iCollin is just updates video streaming state via `OnHMIStatus` notification. This is an opposite action to what is done for `kStarted` case. Note that streaming may transit between suspended <-> started several times while service is opened so SDL should notify application each time on each such transition.
This also allows mobile proxy lib to decide should it stop a/v service or keep it alive. I know that SPT tracks such transitions during streaming and performs some actions.

iCollin (approved):

PR #3491, opened 2020-09-03
Use atomic bool in ApplicationManager

### LINKED ISSUE #3182: CCB: Use atomic bool in ApplicatinManager
### Bug Report
Deadlock inside class "ApplicationManagerImpl"

### Detailed analysis:
Deadlock happens because of using two mutexes for one critical section
inside classes "ApplicationManagerImpl" and "ResumeCtrlImpl".
We are trying to lock one section by different mutexes.
Resolving, by copy set of applications into temporary set and use it
instead of guarded app set which returns by function 'applications()'.
Use atomic bool for flag 'is_stopping_' in class 'ApplicationManagerImpl', remove mutex 'stopping_application_mng_lock_' which needed for guarding of i's_stopping_'.

##### Reproduction Steps
1. TBD

##### Expected Behavior
SDL should not crash

##### Observed Behavior
SDL crashed

#### OS & Version Information
* OS/Version : SYNC4

AKalinich-Luxoft:
Fixes #3182 

### Risk

### Testing Plan
Covered by unit tests

### Summary
Replace bool flag 'is_stopping_' in class ApplicationManagerImpl and remove mutex 'stopping_application_mng_lock_'.

### Tasks Remaining:
- [x] Test this fix

### CLA

### DIFF
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -229,7 +229,7 @@ ApplicationManagerImpl::ApplicationManagerImpl(
 ApplicationManagerImpl::~ApplicationManagerImpl() {
   SDL_LOG_AUTO_TRACE();
 
-  is_stopping_ = true;
+  is_stopping_.store(true);
   SendOnSDLClose();
   media_manager_ = NULL;
   hmi_handler_ = NULL;
@@ -2534,9 +2534,7 @@ bool ApplicationManagerImpl::Init(
 
 bool ApplicationManagerImpl::Stop() {
   SDL_LOG_AUTO_TRACE();
-  stopping_application_mng_lock_.Acquire();
-  is_stopping_ = true;
-  stopping_application_mng_lock_.Release();
+  is_stopping_.store(true);
   application_list_update_timer_.Stop();
   try {
     if (unregister_reason_ ==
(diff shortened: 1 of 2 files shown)


LuxoftAKutsan (approved):

src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -229,7 +229,7 @@ ApplicationManagerImpl::ApplicationManagerImpl(
 ApplicationManagerImpl::~ApplicationManagerImpl() {
   SDL_LOG_AUTO_TRACE();
 
-  is_stopping_ = true;
+  is_stopping_.exchange(true);
```
YarikMamykin, line 232: Why using `exchange` instead of `store`?
I think `exchange` could be more suitable in case we wanted to save previous value of `is_stopping_` variable.

YarikMamykin (changes requested):

src/components/application_manager/src/application_manager_impl.cc
AKalinich-Luxoft: @YarikMamykin actually, just moved a fix without any changes. However I agree that `store` is more applicable in that case. Replaced in https://github.com/smartdevicelink/sdl_core/pull/3491/commits/8209b59b0d52b22d82ab243a7c0d631de55bd662

YarikMamykin (approved):

SNiukalov (approved):

Issue #3484, opened 2020-08-28
SDL sends unexpectedDisconnect=true to HMI in case HMI close an app with RESOURCE_CONSTRAINT reason


dboltovskyi:
### Bug Report
SDL sends `unexpectedDisconnect=true` in `BC.OnAppUnregistered` notification to HMI in case HMI close an app using `BC.OnExitApplication` notification with `RESOURCE_CONSTRAINT` reason

##### Reproduction Steps
1. SDL and HMI are started
2. App1 and App2 are registered on one mobile connection
3. HMI sends `BC.OnExitApplication` with `RESOURCE_CONSTRAINT` reason related to App2 to SDL

##### Expected Behavior
SDL sends `BC.OnAppUnregistered` notification related to App2 with `unexpectedDisconnect: false`

##### Observed Behavior
SDL sends `BC.OnAppUnregistered` notification related to App2 with `unexpectedDisconnect: true`

##### OS & Version Information
* OS/Version: Ubuntu 18.04
* SDL Core Version: https://github.com/smartdevicelink/sdl_core/tree/develop (b476821b44)
* Testing Against: ATF

##### Test Case, Sample Code, and / or Example App
 - Script: [026_OnExitApplication_RESOURCE_CONSTRAINT_multiple_apps_on_one_connection](https://github.com/smartdevicelink/sdl_atf_test_scripts/blob/develop/test_scripts/WebEngine/WebEngineProjection/026_OnExitApplication_RESOURCE_CONSTRAINT_multiple_apps_on_one_connection.lua)
 - Report and logs: [2020-08-28_14-08-15_3484.tar.gz](https://github.com/smartdevicelink/sdl_core/files/5143605/2020-08-28_14-08-15_3484.tar.gz)

LitvinenkoIra:
Luxoft will prepare the fix for this issue

LitvinenkoIra:
Please notice that this is a regression issue

jacobkeeler:
@LitvinenkoIra how can this be a regression issue if `RESOURCE_CONSTRAINT` was introduced in this release?

dboltovskyi:
@jacobkeeler We're apologize for misunderstanding. By saying 'regression' we meant issue needs to be fixed in scope of `7.0` Release since it's related to the feature of this release.

jacobkeeler:
Closing with the merge of #3516

PR #3516, opened 2020-09-25
Fix sdl sends unexpected disconnect with resource constraint

SNiukalov:
Fixes #[3484](https://github.com/smartdevicelink/sdl_core/issues/3484)

### Risk

### Testing Plan
 * ATF scripts #[WebEngineProjection](https://github.com/smartdevicelink/sdl_atf_test_scripts/tree/develop/test_scripts/WebEngine/WebEngineProjection)
 * UT's

### Summary
SDL sends "unexpectedDisconnect=true" because happens data race.

In current implementation at OnExitApplicationNotification for the reason "RESOURCE_CONSTRAINT" SDL sends unregister notification to mobile with flag "final message", after that
application_manager_.UnregisterApplication is called.

The root issue is call application_manager_.UnregisterApplication.
Because SDL send unregister notification to mobile with flag "final message".

At this case in the "ProtocolHandlerImpl::Handle(const impl::RawFordMessageToMobile)"
SDL puts this session to sessions_last_message_id_.
After that at the ProtocolHandlerImpl::OnTMMessageSend SDL puts that connection to ready_to_close_connections_
and sends SendEndSession.

When this message will be sent, SDL call ProtocolHandlerImpl::OnTMMessageSend again and try to find this connection in ready_to_close_connections_
If connection found transport_manager_.Disconnect(connection_handle) is called;

After disconnect done"TransportManagerImpl::Handle(TransportAdapterEvent) case ON_DISCONNECT_DONE" will be called "ConnectionHandlerImpl::OnConnectionClosed".
Which will result in the call
"connection_handler_observer_->OnServiceEndedCallback" with "kCommon" reason and there will be a repeated call "UnregisterApplication with is_unexpected_disconnect = true". 

This PR provides the next changes:
* Rework processing of the OnExitApplication
(RESOURCE_CONSTRAINT) notification
* UTs update

### CLA


SNiukalov:
@theresalech PR is ready for Livio team review. Thank you.

src/components/connection_handler/include/connection_handler/connection_handler_impl.h
```diff
@@ -716,6 +734,11 @@ class ConnectionHandlerImpl
    */
   Connection* ending_connection_;
 
+  /**
+   * @brief List connection, which processed final message
+   */
+  std::set<transport_manager::ConnectionUID> connections_final_message;
```
jacobkeeler, line 740: It seems to me that this map could be replaced with a new bool property for the Connection class `final_message_sent()`. This would eliminate the need for the `ClosureReasonHandling` function

src/components/connection_handler/src/connection_handler_impl.cc
```diff
@@ -1688,9 +1703,11 @@ void ConnectionHandlerImpl::KeepConnectionAlive(uint32_t connection_key,
 }
 
 void ConnectionHandlerImpl::OnConnectionEnded(
-    const transport_manager::ConnectionUID connection_id) {
+    const transport_manager::ConnectionUID connection_id,
+    const CloseSessionReason close_reason) {
   SDL_LOG_INFO("Delete Connection: " << static_cast<int32_t>(connection_id)
-                                     << " from the list.");
+                                     << " from the list."
+                                     << " with reason " << close_reason);
```
jacobkeeler, line 1710: It seems to me that the `reason` could be determined here rather than passed in.

something like "If the final message was sent, send use kFinalMessage, otherwise use kCommon"

jacobkeeler (changes requested):

jacobkeeler:
Another quick question, is there a reason that `OnAppInterfaceUnregistered` is not _always_ considered the final message? Currently the logic only sets `is_final` if the `RESOURCE_CONSTRAINT` reason is used.

src/components/connection_handler/include/connection_handler/connection_handler_impl.h
SNiukalov: @jacobkeeler 
Unfortunately, we may have multiple applications on different connections.
We should be able to identify the connection with "transport_manager :: ConnectionUID".

src/components/connection_handler/src/connection_handler_impl.cc
SNiukalov: @jacobkeeler 
Yes, of course, but CloseSessionReason is an OnConnectionEnded parameter. 
We call this function in several places.
I believe that the separation of the logic of removing the connection and finding out the reason is justified in this case.

SNiukalov:
@jacobkeeler 
Yes, you right. Currently only for the RESOURCE_CONSTRAINT reason.

src/components/connection_handler/include/connection_handler/connection_handler_impl.h
jacobkeeler: `connection_list_` appears to be a 1-1 map from `ConnectionUID` to `Connection`. Are you telling me that a `Connection` can have multiple `ConnectionUID`s?

jacobkeeler: The main issue I have with this new structure is that there seems to be no guarantee that the connection IDs will actually be removed from this set during closure.

SNiukalov: @jacobkeeler 
Sorry, small misunderstanding. 
I thought you want change connections_final_message to new bool property in the connection_handler_impl.h.
Now I understand what you mean, sorry.

jacobkeeler: Ah, I also realized from your comment that there were still issues with my solution, perhaps the field should be added to the `Session` class instead, since that would relate to a single app.

jacobkeeler:
@SNiukalov I'm aware of that, I'm asking why this is the case

src/components/connection_handler/src/connection_handler_impl.cc
jacobkeeler: I still think it would be better to have this logic within the function, since `unexpected_disconnect` should always be `false` if the last message was sent, plus it is better to have the logic in one place.

LuxoftAKutsan:
@jacobkeeler there were no requirements for SDL to close session on app unregister. 
So re-registering application on the same session is actually a valid use-case.  
And [proposal 0273](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0273-webengine-projection-mode.md#4-application-lifecycle-or-what-happens-if-the-app-is-closed) explicitly says that on `RESOURCE_CONSTRAINT` _HMI to be able to **unregister and disconnect** an app_. 

We have no test cases (as I know) of the reregistering applications using the same session. But there is no guarantee that no business apps do this.
So to minimize SDL behavior changes logic is implemented strictly according to the proposal.

theresalech:
Hi @SNiukalov and @LuxoftAKutsan, can you please let us know if you have answers to the outstanding questions on this PR, or provide us with timing on when you will be able to provide those answers?

In preparation for the Core 7.0 Release Candidate delivery to the Steering Committee on Thursday (10/8), we will be entering a code freeze end of day today (10/5). If we are unable to merge this PR today, it may still be something we can include in the Core 7.0 release, but we will not be able to include it in the initial Release Candidate.

dboltovskyi:
Rebuild required

src/components/connection_handler/include/connection_handler/connection_handler_impl.h
LuxoftAKutsan: Adding additional boolean property to the interface saems to me as more complicated and less obvious soultion then set of connection id's with connections_final_messages. 
Also it iwll provide access to this information to all components that wotk with Connection/Session interfaces. For better incapsulation I would leave this solution.

src/components/connection_handler/src/connection_handler_impl.cc
LuxoftAKutsan: @jacobkeeler, I agree that extending arguments of `OnConnectionEnded` note required. 
Moved `ClosureReasonHandling` into `OnConnectionEnded` with 21fcf233f3

dboltovskyi:
Rebuild required

src/components/connection_handler/include/connection_handler/connection_handler_impl.h
jacobkeeler: I disagree, using a property seems far more appropriate in this case.

> The main issue I have with this new structure is that there seems to be no guarantee that the connection IDs will actually be removed from this set during closure.

Using a property will guarantee that this information will not linger past the lifespan of the `Session` object

theresalech:
@LuxoftAKutsan , @SNiukalov , @dboltovskyi - it looks like there is one outstanding comment on this PR. We would like to have this resolved by EOD Monday (10/26) to ensure this PR can be included in the 7.0 release. Can you please respond to the comment, or let us know if you have any concerns with this timing?

src/components/connection_handler/include/connection_handler/connection_handler_impl.h
jacobkeeler: @LuxoftAKutsan see https://github.com/smartdevicelink/sdl_core/pull/3553 for the changes I'm referring to

dboltovskyi:
@theresalech We apologize for the delay. Initially we planned to continue activities on this PR today and found out @jacobkeeler has opened https://github.com/smartdevicelink/sdl_core/pull/3553 based on our https://github.com/smartdevicelink/sdl_core/pull/3516. We're not sure do we still need to continue work on our fix or assist PM with their one.

jacobkeeler:
@dboltovskyi If you would like to incorporate the changes from #3553 into this PR, that would be fine with me. I submitted that PR in case we didn't get the changes from you today so that we could still include this fix in the release.

Alternatively, if you just want to leave a review on that PR, I could merge that one instead.

dboltovskyi:
@jacobkeeler Thank you for you response.
Please notice we have the following tests failed using the build from https://github.com/smartdevicelink/sdl_core/pull/3553
```
./test_scripts/WebEngine/WebEngineProjection/025_OnExitApplication_RESOURCE_CONSTRAINT_one_app.lua
./test_scripts/WebEngine/WebEngineProjection/027_OnExitApplication_RESOURCE_CONSTRAINT_multiple_apps_on_different_connections.lua
```
Though these are passed at: https://github.com/smartdevicelink/sdl_core/pull/3516

@AKalinich-Luxoft currently is investigating the difference between PRs which lead to fails.

AKalinich-Luxoft:
> @jacobkeeler Thank you for you response.
> Please notice we have the following tests failed using the build from #3553
> 
> ```
> ./test_scripts/WebEngine/WebEngineProjection/025_OnExitApplication_RESOURCE_CONSTRAINT_one_app.lua
> ./test_scripts/WebEngine/WebEngineProjection/027_OnExitApplication_RESOURCE_CONSTRAINT_multiple_apps_on_different_connections.lua
> ```
> 
> Though these are passed at: #3516
> 
> @AKalinich-Luxoft currently is investigating the difference between PRs which lead to fails.

@dboltovskyi @jacobkeeler please check https://github.com/smartdevicelink/sdl_core/pull/3516/commits/171419ecf92ebe9a1d5dc58083d441ffa6182f97

jacobkeeler:
As a note, it seems that the root issue that caused the failures was the fact that `session_id` was incorrectly filled out in the `OnTMMessageSend` function, it was always `0`, so attaching this flag to the session didn't work properly. It doesn't seem like there are any issues as of now tying this value to the connection directly, so this is approach that was taken instead, but the invalid session ID still remains an issue that might be worth investigating at some point

jacobkeeler (approved):

AKalinich-Luxoft:
@jacobkeeler regression PASSED on https://github.com/smartdevicelink/sdl_core/pull/3516/commits/04662fad4dd6d4e32dff8370e53a0cbf58bdfb71
PR is ready for merge

dboltovskyi:
Rebuild required

PR #3525, opened 2020-10-02
Add check for already processed resumptions to ResumptionDataProcessor

### LINKED ISSUE #3522: SDL sometimes finishes resumption successfully even if error has been received from HMI
### Bug Report
SDL sometimes finishes resumption successfully even if error has been received from HMI 

##### Occurrences:
~0.01% (very rare)

##### Reproduction Steps
1. App sends `AddCommand` request (which is stored by SDL as resumption data)
2. App adds some other resumption data (e.g. SubMenu, VD and WP subscription etc.)
3. Unexpected disconnect and reconnect are performed
4. App re-register with actual HashId
5. SDL starts resumption process for App and sends all required requests to HMI
6. HMI replies with erroneous response to one of the requests (e.g. `VR.AddCommand`)

##### Expected Behavior
SDL fails resumption and send `RESUME_FAILED` to app

##### Observed Behavior
SDL successfully finishes resumption and send `SUCCESS` to app

* OS/Version: Ubuntu 18.04
* SDL Core Version: https://github.com/smartdevicelink/sdl_core/tree/develop (a36ea82)
* Testing Against: ATF

 - ATF Script: [004_resume_one_rpc_with_error_result_code_ign_off](https://github.com/smartdevicelink/sdl_atf_test_scripts/blob/develop/test_scripts/Resumption/Handling_errors_from_HMI/004_resume_one_rpc_with_error_result_code_ign_off.lua)
 - Report and Logs: [2020-09-29_21-34-37_3522.tar.gz](https://github.com/smartdevicelink/sdl_core/files/5302722/2020-09-29_21-34-37_3522.tar.gz)

ychernysheva:
Fixes #3522

### Risk

### Summary
This PR is intended to improve synchronization in `ResumptionDataProcessorImpl::Restore()` method. Root cause of this problem is a case when some requests from HMI may be processed earlier than callback will be set in ResumptionDataProcessor. So in this case execution of `ProcessResumptionStatus` method for such application will be aborted due to absence of callback, and this behavior is valid and expected. But after this variable `is_requests_list_empty` will be assigned to true, and there is a problem, because now Restore() method doesn't have any checks for actual result of already performed operation. That's why in all such cases, regardless of actual resumption status, callback will be called with parameters mobile_apis::Result::SUCCESS and "Data resumption successful".

That's why corresponding check is added to `Restore` method, also some refactoring is performed to avoid duplication of code, redundant check `HasDataToRestore` is deleted (now in fact it will be always true), and potential source of memory leak in `EraseAppResumptionData` method is fixed (before only last request was deleted from `request_app_ids_` map).

In current environment this issue is hard to reproduce, but it may be a problem if this code is used in other environment or in other systems (for example, with proprietary hardware) without synchronization improvement.

### CLA

### DIFF
--- a/src/components/application_manager/include/application_manager/resumption/resumption_data_processor_impl.h
+++ b/src/components/application_manager/include/application_manager/resumption/resumption_data_processor_impl.h
@@ -125,14 +125,22 @@ class ResumptionDataProcessorImpl
                                const ResumptionRequest& found_request);
 
   /**
-   * @brief IsResumptionFinished checks whether some responses are still waiting
+   * @brief EraseProcessedRequest erases processed request from list of pending
+   * requests
    * @param app_id ID of application, related to event
    * @param found_request reference to found request
-   * @return true, if resumption for this application is finished, or false, if
-   * some requests aren't processed  yet
    */
-  bool IsResumptionFinished(const uint32_t app_id,
-                            const ResumptionRequest& found_request);
+  void EraseProcessedRequest(const uint32_t app_id,
+                             const ResumptionRequest& found_request);
+
+  /**
+   * @brief IsResumptionFinished checks whether some responses are still
+   * waiting
+   * @param app_id ID of application, related to event
+   * @return true, if resumption for this application is finished, or false, if
+   * some requests aren't processed yet
+   */
+  bool IsResumptionFinished(const uint32_t app_id) const;
 
   /**
    * @brief IsResumptionSuccessful checks whether overall resumption status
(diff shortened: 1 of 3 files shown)


ychernysheva:
@theresalech This fix is ready for review.
Thank you!

LuxoftAKutsan (approved):

dboltovskyi:
Rebuild required

dboltovskyi:
Rebuild required

PR #3526, opened 2020-10-02
Re-order rai logic for plugins

### LINKED ISSUE #3172: CCB: Fix SIGSEGV Crash subscribed_data

JackLivio:
Fixes #3172 #3162 

This PR is **Ready** for review.

### Risk

### Testing Plan
Send on vehicle data notification at same time app registers. Core should not crash.

### Summary
Moves the call to add the newly registered app to the applications list later in the registration process. This ensures that an app has all of its necessary plugins added before it is accessible by the applications accessor.

### CLA

### DIFF
--- a/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/register_app_interface_request.cc
+++ b/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/register_app_interface_request.cc
@@ -746,6 +746,9 @@ void RegisterAppInterfaceRequest::Run() {
     plugin.OnApplicationEvent(plugin_manager::kApplicationRegistered,
                               application);
   };
+  // To prevent timing issues, this event is called before an app is accessible
+  // by the applications accessor. This prevents incoming hmi rpcs from
+  // attempting to access an app before it has been fully initialized.
   application_manager_.ApplyFunctorForEachPlugin(on_app_registered);
   application_manager_.FinalizeAppRegistration(application, connection_key());
 
--- a/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/register_app_interface_request.cc
+++ b/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/register_app_interface_request.cc
@@ -742,6 +742,13 @@ void RegisterAppInterfaceRequest::Run() {
   SetupAppDeviceInfo(application);
   auto status_notifier = AddApplicationDataToPolicy(application);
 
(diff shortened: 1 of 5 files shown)


src/components/include/application_manager/application_manager.h
```diff
@@ -558,6 +558,9 @@ class ApplicationManager {
       const std::shared_ptr<smart_objects::SmartObject>&
           request_for_registration) = 0;
 
+  virtual void FinalizeAppRegistration(ApplicationSharedPtr application,
```
LuxoftAKutsan, line 561: Would be nice to have doxygen description

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/register_app_interface_request.cc
```diff
@@ -742,6 +742,13 @@ void RegisterAppInterfaceRequest::Run() {
   SetupAppDeviceInfo(application);
   auto status_notifier = AddApplicationDataToPolicy(application);
 
+  auto on_app_registered = [application](plugin_manager::RPCPlugin& plugin) {
+    plugin.OnApplicationEvent(plugin_manager::kApplicationRegistered,
+                              application);
+  };
+  application_manager_.ApplyFunctorForEachPlugin(on_app_registered);
```
LuxoftAKutsan, line 749: In this case it is important to remember that when the plugin received `kApplicationRegistered` event, the application is not finally registered. I would add this as a comment in the `kApplicationRegistered` enum

LuxoftAKutsan (approved):

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/register_app_interface_request.cc
JackLivio: Noting that this placement is the most important part of the change. This event needed to happen before an app was added to the applications accessor in the app manager.

jacobkeeler (approved):

Issue #3532, opened 2020-10-07
SDL does not transfer OnSystemCapabilityUpdated notification to App after IGN_OFF/IGN_ON cycle


VjKlepikov:
### Bug Report
SDL does not transfer OnSystemCapabilityUpdated notification from HMI to an App after IGN_OFF/IGN_ON cycle in case App sent Show RPC with `templateConfiguration` parameter for widget window after `Show` response

#### Preconditions:
1) In sdl_preloaded_pt.json default app_policies groups:
["Base-4", "WidgetSupport" ]
2) SDL and HMI are started
3) App is registered and activated(App has hmi level FULL/LIMITED )
4) App created and activated a widget ( the widget has "hmi_level": FULL )
5) IGN_OFF and IGN_ON are performed
6) App is registered with actual hashId
7) Widget is activated

##### Reproduction Steps
App sends "Show" request with "templateConfiguration" param for the restored widget
HMI sends BC.OnSystemCapabilityUpdated:
HMI→SDL: "BC.OnSystemCapabilityUpdated"(params)

##### Expected Behavior
SDL→App: “OnSystemCapabilityUpdated” (params)

##### Observed Behavior
SDL does not transfer OnSystemCapabilityUpdated(params) to Mobile App

##### OS & Version Information
SDL: https://github.com/smartdevicelink/sdl_core/commits/develop (9f61819)
HMI: https://github.com/smartdevicelink/sdl_hmi/tree/develop (f712d4e)

VjKlepikov:
@theresalech Please notice this is regression issue

iCollin:
Closed via https://github.com/smartdevicelink/sdl_core/pull/3550

Issue #3542, opened 2020-10-16
SDL is unable to activate application after resumption in case of two apps


dboltovskyi:
### Bug Report
SDL is unable to activate application after resumption in case of two apps

##### Reproduction Steps
1. There are two `MEDIA` apps registered: `App_1` and `App_2`
2. `App_1` is in `LIMITED` and `App_2` is in `BACKGROUND` HMI levels
3. Unexpected disconnect and reconnect are performed
4. SDL resumed HMI level of both apps:
=> `App_1` is in `LIMITED`, `App_2` is in `BACKGROUND` HMI levels
5. `App_2` activated
6. `App_1` activated

##### Expected Behavior
SDL does:
  - send `OnHMIStatus(FULL)` for `App_1`
  - send `OnHMIStatus(BACKGROUND)` for `App_2`

##### Observed Behavior
SDL does:
 - reply to `SDL.ActivateApp` request from HMI
 - send `OnHMIStatus(NONE)` for `App_1`
 - not send `OnHMIStatus` for `App_2`

##### OS & Version Information
* OS/Version: Ubuntu 18.04
* SDL Core Version: https://github.com/smartdevicelink/sdl_core/tree/release/7.0.0 (5c72cddc35)
* Testing Against: ATF

##### Test Case, Sample Code, and / or Example App
 - ATF Script: [3542_apps_activation_after_resumption.lua.tar.gz](https://github.com/smartdevicelink/sdl_core/files/5388808/3542_apps_activation_after_resumption.lua.tar.gz)
 - Report and Logs: [2020-10-15_22-16-45.345_3542.tar.gz](https://github.com/smartdevicelink/sdl_core/files/5388807/2020-10-15_22-16-45.345_3542.tar.gz)

dboltovskyi:
Please notice issue is applicable for [6.1.0](https://github.com/smartdevicelink/sdl_core/tree/release/6.1.0) and [6.0.1](https://github.com/smartdevicelink/sdl_core/tree/release/6.0.1) releases, however is not reproduced at [5.1.3](https://github.com/smartdevicelink/sdl_core/tree/5.1.3)
Probably the root cause is the fix https://github.com/smartdevicelink/sdl_core/pull/3117 which was included in `6.0.1`

dboltovskyi:
@theresalech We would like to treat this issue as regression since it's not reproduced at `5.1.3` and steps to reproduce are quite simple.

theresalech:
[...]

PR #3543, opened 2020-10-16
Fix/ OnSystemCapabilityUpdated for widget after resumption

LitvinenkoIra:
Fixes https://github.com/smartdevicelink/sdl_core/issues/3532

### Risk

### Testing Plan
ATF test scripts from https://github.com/smartdevicelink/sdl_atf_test_scripts/pull/2465

### Summary
After the widgets resumption, SDL doesn't resend the OnSCU notification with updates for the widget to the mobile app (If HMI sends this notification after the resumption is finished). 

This is happening because [here](https://github.com/smartdevicelink/sdl_core/blob/release/7.0.0/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/on_system_capability_updated_notification.cc#L221)  we are checking the values of two flags: `is_resuming_` and `is_widget_windows_resumption_` and they both are `true` (during all ign cycle) if the application was resumed and it had widgets to resume. 

I changed the `is_widget_windows_resumption_` to false (`StopWaitingForWidgets` function) if widget resumption already finished and [SDL sent OnSCU notification after registration](https://github.com/smartdevicelink/sdl_core/pull/3543/files#diff-89664857793adc72d087a6958999865d16e2ee4c595fb4c23fa4e397bad4e93eR664), in order to allow processing of subsequent OnSCU notifications.

Also, the `window_ids_to_resume_` set was cleared after receiving the unsuccessful response for at least one UI.CreateWindow. But SDL may send another UI.CreateWindow after that and receive OnSCU notification from HMI, this sequence [still should be a part of resumption](https://github.com/LuxoftSDL/sdl_core/blob/fix/on_system_cap_updated_for_widget_after_resumption/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/hmi/on_bc_system_capability_updated_notification_from_hmi.cc#L93) (SDL shouldn't resend this notification to the mobile). 
So we shouldn't clear the window_ids_to_resume_ set till the resumption is finished, but we still need to reset the display capabilities after receiving the unsuccessful response for at least one UI.CreateWindow. This is why I modified the `ResetDisplayCapabilities` function - it deletes only one window and resets the capabilities, and added the new one `StopWaitingForWindows` - it clears the window_ids_to_resume_ set.

Also `StopWaitingForWindow` renamed to the `StopWaitingForMainWindow` because we use it only for Main window [here](https://github.com/LuxoftSDL/sdl_core/blob/fix/on_system_cap_updated_for_widget_after_resumption/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/mobile/register_app_interface_request.cc#L460)

### CLA

### DIFF
--- a/src/components/application_manager/src/resumption/resumption_data_processor_impl.cc
+++ b/src/components/application_manager/src/resumption/resumption_data_processor_impl.cc
@@ -311,8 +311,11 @@ void ResumptionDataProcessorImpl::FinalizeResumption(
   } else {
     SDL_LOG_ERROR("Resumption for app " << app_id << " failed");
     callback(mobile_apis::Result::RESUME_FAILED, "Data resumption failed");
-    RevertRestoredData(application_manager_.application(app_id));
+    auto app = application_manager_.application(app_id);
+    RevertRestoredData(app);
     application_manager_.state_controller().DropPostponedWindows(app_id);
+    auto& builder = app->display_capabilities_builder();
+    builder.StopWaitingForWindows();
   }
   EraseAppResumptionData(app_id);
 }
@@ -1096,7 +1099,7 @@ void ResumptionDataProcessorImpl::CheckCreateWindowResponse(
     SDL_LOG_ERROR("UI_CreateWindow for correlation id: " << correlation_id
                                                          << " has failed");
     auto& builder = application->display_capabilities_builder();
-    builder.ResetDisplayCapabilities();
+    builder.ResetDisplayCapabilities(window_id);
     return;
   }
 
--- a/src/components/application_manager/src/resumption/resumption_data_processor_impl.cc
+++ b/src/components/application_manager/src/resumption/resumption_data_processor_impl.cc
@@ -311,8 +311,7 @@ void ResumptionDataProcessorImpl::FinalizeResumption(
   } else {
     SDL_LOG_ERROR("Resumption for app " << app_id << " failed");
     callback(mobile_apis::Result::RESUME_FAILED, "Data resumption failed");
-    auto app = application_manager_.application(app_id);
-    RevertRestoredData(app);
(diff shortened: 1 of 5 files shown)


LitvinenkoIra:
@theresalech this PR is ready for Livio review

src/components/application_manager/src/resumption/resumption_data_processor_impl.cc
```diff
@@ -311,8 +311,11 @@ void ResumptionDataProcessorImpl::FinalizeResumption(
   } else {
     SDL_LOG_ERROR("Resumption for app " << app_id << " failed");
     callback(mobile_apis::Result::RESUME_FAILED, "Data resumption failed");
-    RevertRestoredData(application_manager_.application(app_id));
+    auto app = application_manager_.application(app_id);
+    RevertRestoredData(app);
     application_manager_.state_controller().DropPostponedWindows(app_id);
+    auto& builder = app->display_capabilities_builder();
+    builder.StopWaitingForWindows();
```
iCollin, line 318: can these two lines be in DeleteWindowsSubscriptions called from RevertRestoredData?

LitvinenkoIra: I think yes, let me check the regression and I will push the updates

LitvinenkoIra: @iCollin It seems like this issue have the same root cause as https://github.com/smartdevicelink/sdl_core/issues/3542 
We need some time for the investigation but it seems like this fix should be reworked. 
I'll keep you posted

LitvinenkoIra:
@iCollin  https://github.com/smartdevicelink/sdl_core/pull/3550 fixes this issue. 
This PR is closed as redundant

PR #3546, opened 2020-10-20
Bugfix 2808 copy

JackLivio:
Updated version of pr #3428

### CLA


JackLivio:
I noticed some atf failure relating on OnEventChange. Deactivate_HMI seems to be crashing core. This fix will need some updates.

JackLivio:
Crash is fixed, ready for review.

src/components/application_manager/src/hmi_state.cc
```diff
@@ -249,6 +249,10 @@ mobile_apis::HMILevel::eType PhoneCallHmiState::max_hmi_level() const {
   using namespace helpers;
   using namespace mobile_apis;
 
+  if (parent() && HMILevel::INVALID_ENUM == parent()->hmi_level()) {
+    return parent()->hmi_level();
+  }
```
jacobkeeler, line 254:
```suggestion
  if (HMILevel::INVALID_ENUM == parent_hmi_level()) {
    return parent_hmi_level();
  }
```
This function has a null check for `parent()` when necessary

src/components/application_manager/src/hmi_state.cc
```diff
@@ -279,6 +283,10 @@ mobile_apis::HMILevel::eType DeactivateHMI::max_hmi_level() const {
...
```
jacobkeeler, line 288:
```suggestion
  if (HMILevel::INVALID_ENUM == parent_hmi_level()) {
    return parent_hmi_level();
  }
```
This function has a null check for `parent()` when necessary

src/components/application_manager/src/hmi_state.cc
```diff
@@ -323,6 +331,10 @@ mobile_apis::VideoStreamingState::eType AudioSource::video_streaming_state()
 mobile_apis::HMILevel::eType AudioSource::max_hmi_level() const {
   using namespace mobile_apis;
 
+  if (parent() && HMILevel::INVALID_ENUM == parent()->hmi_level()) {
+    return parent()->hmi_level();
+  }
```
jacobkeeler, line 336:
```suggestion
  if (HMILevel::INVALID_ENUM == parent_hmi_level()) {
    return parent_hmi_level();
  }
```
This function has a null check for `parent()` when necessary

src/components/application_manager/src/hmi_state.cc
```diff
@@ -396,6 +408,10 @@ mobile_apis::HMILevel::eType EmbeddedNavi::max_hmi_level() const {
   using namespace mobile_apis;
   using namespace helpers;
 
+  if (parent() && HMILevel::INVALID_ENUM == parent()->hmi_level()) {
+    return parent()->hmi_level();
+  }
```
jacobkeeler, line 413:
```suggestion
  if (HMILevel::INVALID_ENUM == parent_hmi_level()) {
    return parent_hmi_level();
  }
```
This function has a null check for `parent()` when necessary

jacobkeeler (changes requested):
In addition, I think these checks should be in the `hmi_level` function rather than the `max_hmi_level` function

src/components/application_manager/src/hmi_state.cc
jacobkeeler: Also should probably be in the `hmi_level` function instead of `max_hmi_level`, same goes for all of these checks

jacobkeeler (approved):

PR #3550, opened 2020-10-22
Fix application activation after resumption

### LINKED ISSUE #3542: SDL is unable to activate application after resumption in case of two apps

ychernysheva:
Fixes #3542 #3532 

### Risk

### Testing Plan
Covered by ATF test scripts

### Summary
There is a case when media application acquires wrong HMI Level (NONE instead of FULL) during activation. Root cause of this issue is a `true` value of `is_resuming` flag for application, which in fact completed resumption process. Before this flag was set to false only for applications with failed resumption, so there is a need to assign it to false for all already resumed applications regardless of successfulness of resumption.

Also note, that after widget resumption SDL doesn't resend the OnSCU notification with updates for the widget to the mobile app (If HMI sends this notification after the resumption is finished). This is happening because we are checking the value of `is_resuming_` flag, which now is `true` even if the application was resumed and it had widgets to resume. So root cause of wrong behavior of OnSCU notification is incorrect value of `is_resuming_` flag after completed resumption.

Additionally tiny refactoring was performed for two methods of ResumeController class: `RestoreAppHMIState` and `StartAppHmiStateResumption`. Their return value was changed from bool to void, because in fact we don't need this values, and determination of this value makes logic of these methods more complicated than it may be. Mentioned methods are invoked only once inside of ResumeController class, that's why this changes cannot adversely affect another logic.

### CLA

### DIFF
--- a/src/components/application_manager/include/application_manager/resumption/resume_ctrl.h
+++ b/src/components/application_manager/include/application_manager/resumption/resume_ctrl.h
@@ -84,9 +84,8 @@ class ResumeCtrl {
   /**
    * @brief Set application HMI Level and ausio_state as saved
    * @param application is application witch HMI Level is need to restore
-   * @return true if success, otherwise return false
    */
-  virtual bool RestoreAppHMIState(
+  virtual void RestoreAppHMIState(
       application_manager::ApplicationSharedPtr application) = 0;
 
   /**
@@ -255,7 +254,7 @@ class ResumeCtrl {
    * @param application - application to restore hmi level
    * and audio streaming state
    */
-  virtual bool StartAppHmiStateResumption(
+  virtual void StartAppHmiStateResumption(
       application_manager::ApplicationSharedPtr application) = 0;
 
   /**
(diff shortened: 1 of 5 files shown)


ychernysheva:
@theresalech This fix is ready for review.
Thank you!

iCollin (approved):

Issue #3549, opened 2020-10-21
SDL sometimes sends OnHMIStatus in wrong order when app is activated too fast by HMI


dboltovskyi:
### Bug Report
SDL sometimes sends OnHMIStatus in wrong order when app is activated too fast by HMI.
It's very unlikely issue may be faced with mobile apps, however it may happen with web apps.

##### Occurrences
~20-50%

##### Reproduction Steps
1. App is registered
2. HMI activates app just after receiving `BC.OnAppRegistered` notification

##### Expected Behavior
SDL sends 2 `OnHMIStatus` notifications to App: `NONE`, `FULL`

##### Observed Behavior
SDL sends 2 `OnHMIStatus` notifications to App: `FULL`, `NONE`

##### OS & Version Information
* OS/Version: Ubuntu 18.04
* SDL Core Version: https://github.com/smartdevicelink/sdl_core/tree/release/7.0.0 (3ec2b6b)
* Testing Against: ATF

##### Test Case, Sample Code, and / or Example App
 - ATF Script: [3549_Register_App_and_Activate.lua.tar.gz](https://github.com/smartdevicelink/sdl_core/files/5418966/3549_Register_App_and_Activate.lua.tar.gz)
 - Report and Logs: [2020-10-21_18-21-44.747_3549.tar.gz](https://github.com/smartdevicelink/sdl_core/files/5418962/2020-10-21_18-21-44.747_3549.tar.gz)

dboltovskyi:
@theresalech Please notice this issue is not reproducible at `6.1.1` release, so it can be treated as regression one. Luxoft team is going to provide a fix.

iCollin:
Closed via #3559

PR #3557, opened 2020-10-29
Fix/ Postpone application activation if registration is not complete yet

LitvinenkoIra:
Fixes https://github.com/smartdevicelink/sdl_core/issues/3549

### Risk

### Testing Plan
Script from https://github.com/smartdevicelink/sdl_atf_test_scripts/pull/2470

### Summary
There was an issue when the SDL receives SDLActivateAppRequest from HMI and activates the app before the registration is completed and NONE HMI level is assigned. In this case, after the registration is complete SDL will assign NONE hmi level for the app that was in FULL so additional steps will be required to activate the application one more time.

To prevent such a situation additional check for app HMI level has been added during the app activation (SDLActivateAppRequest):
if the app HMI level is INVALID_ENUM - it means that the registration isn't complete yet - so we need to postpone the activation.
After the registration is completed and HMI level NONE is assigned - need to check if we have the postponed activation for the app and if yes - start the postponed activation.

### CLA

### DIFF
--- a/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/hmi/sdl_activate_app_request.cc
+++ b/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/hmi/sdl_activate_app_request.cc
@@ -118,6 +118,15 @@ void SDLActivateAppRequest::Run() {
 
   ApplicationConstSharedPtr app =
       application_manager_.WaitingApplicationByID(app_id());
+
+  if (!app) {
+    app = application_manager_.application(app_id());
+    if (!app) {
+      SDL_LOG_WARN("Can't find application within waiting apps: " << app_id());
+      return;
+    }
+  }
+
   if (application_manager_.state_controller().IsStateActive(
           HmiState::STATE_ID_DEACTIVATE_HMI)) {
     SDL_LOG_DEBUG(
@@ -139,6 +148,18 @@ void SDLActivateAppRequest::Run() {
     application_manager_.connection_handler().ConnectToDevice(app->device());
   } else {
     const uint32_t application_id = app_id();
+    auto main_state =
+        app->CurrentHmiState(mobile_apis::PredefinedWindows::DEFAULT_WINDOW);
+    if (mobile_apis::HMILevel::INVALID_ENUM == main_state->hmi_level()) {
+      SDL_LOG_DEBUG(
+          "Application registration is not completed, HMI level hasn't set "
+          "yet, postpone activation");
+      auto& postponed_activation_ctrl = application_manager_.state_controller()
+                                            .GetPostponedActivationController();
+      postponed_activation_ctrl.AddAppToActivate(application_id,
+                                                 correlation_id());
+      return;
+    }
     policy_handler_.OnActivateApp(application_id, correlation_id());
   }
 }
@@ -186,6 +207,18 @@ void SDLActivateAppRequest::Run() {
 
   if (app_to_activate->IsRegistered()) {
     SDL_LOG_DEBUG("Application is registered. Activating.");
+    auto main_state = app_to_activate->CurrentHmiState(
+        mobile_apis::PredefinedWindows::DEFAULT_WINDOW);
+    if (mobile_apis::HMILevel::INVALID_ENUM == main_state->hmi_level()) {
+      SDL_LOG_DEBUG(
+          "Application registration is not completed, HMI level hasn't set "
+          "yet, postpone activation");
+      auto& postponed_activation_ctrl = application_manager_.state_controller()
+                                            .GetPostponedActivationController();
+      postponed_activation_ctrl.AddAppToActivate(application_id,
+                                                 correlation_id());
+      return;
+    }
     policy_handler_.OnActivateApp(application_id, correlation_id());
     return;
   } else if (app_to_activate->is_cloud_app()) {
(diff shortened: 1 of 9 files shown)


LitvinenkoIra:
@theresalech This PR is ready for Livio review. Thank you!

src/components/application_manager/src/state_controller_impl.cc
```diff
@@ -955,6 +957,15 @@ void StateControllerImpl::OnStateChanged(ApplicationSharedPtr app,
 
   if (new_state->hmi_level() == mobile_apis::HMILevel::HMI_NONE) {
     app->ResetDataInNone();
+    if (mobile_apis::HMILevel::INVALID_ENUM == old_state->hmi_level()) {
+      const uint32_t app_id = app->app_id();
+      const uint32_t corr_id =
+          postponed_activation_controller_.GetPendingActivationCorrId(app_id);
+      if (corr_id > 0) {
+        app_mngr_.GetPolicyHandler().OnActivateApp(app_id, corr_id);
```
iCollin, line 965: Why is onActivateApp called before these two lines are run?
```
app_mngr_.OnHMIStateChanged(app->app_id(), old_state, new_state);
app->usage_report().RecordHmiStateChanged(new_state->hmi_level());
```

src/components/application_manager/src/state_controller_impl.cc
```diff
@@ -955,6 +957,15 @@ void StateControllerImpl::OnStateChanged(ApplicationSharedPtr app,
 
   if (new_state->hmi_level() == mobile_apis::HMILevel::HMI_NONE) {
     app->ResetDataInNone();
+    if (mobile_apis::HMILevel::INVALID_ENUM == old_state->hmi_level()) {
```
iCollin, line 960: Can you please explain when this code will run when `GetDefaultHmiLevel` returns a status other than `NONE`?

AKalinich-Luxoft: @iCollin I believe this code will not be executed if app transits from INVALID to whatever except NONE. It is a good point and I think this block should be moved to the end of function

AKalinich-Luxoft: @iCollin I believe this code will not be executed if app transits from INVALID to whatever except NONE. It is a good point and I think this block should be moved to the end of function

src/components/application_manager/src/state_controller_impl.cc
AKalinich-Luxoft: @iCollin as I can see `OnActivateApp` indirectly triggers the same function `OnStateChanged` too.
Probably, it make sense to move these lines before

src/components/application_manager/src/state_controller_impl.cc
JackLivio: @AKalinich-Luxoft At the end of this function i suggest this is the implemented fix:

```
  if (mobile_apis::HMILevel::INVALID_ENUM == old_state->hmi_level() && app_mngr_.GetDefaultHmiLevel(app) == new_state->hmi_level()) {
    const uint32_t app_id = app->app_id();
    const uint32_t corr_id =
        postponed_activation_controller_.GetPendingActivationCorrId(app_id);
    if (corr_id > 0) {
      app_mngr_.GetPolicyHandler().OnActivateApp(app_id, corr_id);
      postponed_activation_controller_.RemoveAppToActivate(app_id);
    }
  }
```

JackLivio: This should address the possibility for a default hmi level other than NONE

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/hmi/sdl_activate_app_request.cc
```diff
@@ -186,6 +207,18 @@ void SDLActivateAppRequest::Run() {
...
+          "yet, postpone activation");
+      auto& postponed_activation_ctrl = application_manager_.state_controller()
+                                            .GetPostponedActivationController();
+      postponed_activation_ctrl.AddAppToActivate(application_id,
+                                                 correlation_id());
+      return;
+    }
     policy_handler_.OnActivateApp(application_id, correlation_id());
     return;
   } else if (app_to_activate->is_cloud_app()) {
```
iCollin, line 224: @AKalinich-Luxoft there are some cases where this delay logic will not be applied, such as if the app is a cloud app. I think the other call to `policy_handler_.OnActivateApp` in `SDLActivateAppRequest::on_event` should have the postponed_activation_ctrl logic as well.

AKalinich-Luxoft: @iCollin fixed in https://github.com/smartdevicelink/sdl_core/pull/3557/commits/1b59bec2a67c251ed6455496501a91a64c14b4b5

src/components/application_manager/src/state_controller_impl.cc
AKalinich-Luxoft: @iCollin fixed in https://github.com/smartdevicelink/sdl_core/pull/3557/commits/4605e9859af606ef434484e8c9ea094918e4d365

src/components/application_manager/src/state_controller_impl.cc
AKalinich-Luxoft: @iCollin @JackLivio fixed in https://github.com/smartdevicelink/sdl_core/pull/3557/commits/4605e9859af606ef434484e8c9ea094918e4d365

PR #3559, opened 2020-10-29
ensure apps receive default HMI status before activated status

iCollin:
### Summary

There was an issue when the SDL receives SDLActivateAppRequest from HMI and activates the app before the registration is completed and NONE (or other default) HMI level is assigned. In this case, after the registration is complete SDL will assign NONE (or other default) hmi level for the app that was in FULL so additional steps will be required to activate the application one more time.

To prevent such a situation additional check for app HMI level has been added during the app activation (SDLActivateAppRequest):
if the app HMI level is INVALID_ENUM - it means that the registration isn't complete yet - so we need to postpone the activation.
After the registration is completed and HMI level NONE (or other default) is assigned - need to check if we have the postponed activation for the app and if yes - start the postponed activation.

### CLA

### DIFF
--- a/src/components/application_manager/include/application_manager/application_manager_impl.h
+++ b/src/components/application_manager/include/application_manager/application_manager_impl.h
@@ -415,6 +415,7 @@ class ApplicationManagerImpl
    * @brief Closes all registered applications
    */
   void UnregisterAllApplications();
+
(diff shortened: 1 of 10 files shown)


JackLivio (approved):

JackLivio:
@iCollin Please update pr description

PR #3560, opened 2020-10-30
Fix threaded socket termination

AKalinich-Luxoft:
Fixes https://github.com/smartdevicelink/sdl_core/issues/3561

### Risk

### Testing Plan
Covered by ATF scripts

### Summary
There was noticed an issue with ThreadedSocket connection which often closes connection like an "unexpected disconnect"
calling `ConnectionAborted()` every time even if it is handled disconnect. Also, there were no any events raised to Transport
Manager for rare cases when connect is expected. To fix that, flow has been separated for handling expected and unexpected disconnect. Expected disconnect will be handled in case when external component calls `Disconnect()` and socket thread is alive. In that case should be called `DisconnectDone()` when thread is stopped.
Unexpected disconnect will be handled in case when internal thread has been aborted and `Abort()` was called. In that case internal thread `ConnectionAborted()` will be called from internal thread to notify Transport Manager that physical disconnect has happened. This issue is pretty hard to detect just because business logic handles both disconnect types in the same way for now, but this might be changed in the future.

### CLA

### DIFF
--- a/src/components/transport_manager/include/transport_manager/transport_adapter/threaded_socket_connection.h
+++ b/src/components/transport_manager/include/transport_manager/transport_adapter/threaded_socket_connection.h
@@ -189,7 +189,9 @@ class ThreadedSocketConnection : public Connection {
   int write_fd_;
   void threadMain();
   void Transmit();
-  void Finalize();
+  void FinalizeExpectedDisconnect();
+  void FinalizeUnexpectedDisconnect();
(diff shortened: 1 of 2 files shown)


Ypostolov (approved):

ychernysheva (approved):

LitvinenkoIra (approved):

AKalinich-Luxoft:
Reopened for `develop` in #3566

PR #3573, opened 2020-11-13
Add OnWayPointChange handling for NAVIGATION app services

### LINKED ISSUE #3569: SDL Core ignores OnWayPointChange from NAVIGATION App Service Provider
### Bug Report
NAVIGATION app services have the option to handle the `GetWayPoints` RPC, but cannot provide updates for these waypoints via `OnWayPointChange` since SDL Core will ignore this message from mobile.

##### Reproduction Steps
1. Register NAVIGATION ASP for app1 with `handledRPCs: [45]` and activate the new service
2. Subscribe app2 to waypoints using `SubscribeWayPoints`
3. Send `OnWayPointChange` from app1

##### Expected Behavior
app2 receives OnWayPointChange from app1

##### Observed Behavior
OnWayPointChange notification from app1 is ignored

* OS/Version: Ubuntu 18.04
* SDL Core Version: 7.0.0 (reproducible since 5.1.0)
* Testing Against: Java Test Suite

jacobkeeler:
Fixes #3569 

### Risk

### Testing Plan
https://github.com/smartdevicelink/sdl_atf_test_scripts/pull/2486

### Summary
Add handling for OnWayPointChange notification from NAVIGATION app service. NAVIGATION app service will take over waypoints processing when active if it specifies that it can handle these messages. This allows waypoints to be used even when a system doesn't support this feature directly.

##### Bug Fixes
* Core now handles OnWayPointChange from app when its service handles the GetWayPoints RPC.

### CLA

### DIFF
--- a/src/appMain/sdl_preloaded_pt.json
+++ b/src/appMain/sdl_preloaded_pt.json
@@ -854,8 +854,7 @@
                         "hmi_levels": [
                             "FULL",
                             "LIMITED",
-                            "BACKGROUND",
-                            "NONE"
+                            "BACKGROUND"
                         ]
                     },
                     "UnpublishAppService": {
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -4737,7 +4737,7 @@ void ApplicationManagerImpl::SubscribeAppForWayPoints(uint32_t app_id,
   }
   SDL_LOG_DEBUG("There are applications subscribed: "
                 << subscribed_way_points_apps_list_.size());
-  if (GetAppServiceManager().IsWayPointsHandled()) {
+  if (GetAppServiceManager().FindWayPointsHandler() != nullptr) {
     auto service = GetAppServiceManager().ActiveServiceForType(
         EnumToString(mobile_apis::AppServiceType::NAVIGATION));
     auto it = mobile_way_points_data_.find(service->connection_key);
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -4750,8 +4750,7 @@ void ApplicationManagerImpl::SubscribeAppForWayPoints(uint32_t app_id,
     (*way_point_notification)[strings::params][strings::connection_key] =
         app_id;
     GetRPCService().SendMessageToMobile(way_point_notification);
-  }
-  if (hmi_way_points_data_) {
+  } else if (hmi_way_points_data_) {
     smart_objects::SmartObjectSPtr way_point_notification =
         std::make_shared<smart_objects::SmartObject>(*hmi_way_points_data_);
     (*way_point_notification)[strings::params][strings::connection_key] =
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -162,6 +162,7 @@ ApplicationManagerImpl::ApplicationManagerImpl(
           std::make_shared<sync_primitives::RecursiveLock>())
     , apps_to_register_list_lock_ptr_(std::make_shared<sync_primitives::Lock>())
     , reregister_wait_list_lock_ptr_(std::make_shared<sync_primitives::Lock>())
+    , subscribed_to_hmi_way_points_(false)
     , audio_pass_thru_active_(false)
     , audio_pass_thru_app_id_(0)
     , driver_distraction_state_(hmi_apis::Common_DriverDistractionState::DD_OFF)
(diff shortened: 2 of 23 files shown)


src/appMain/sdl_preloaded_pt.json
```diff
@@ -854,8 +854,7 @@
                         "hmi_levels": [
                             "FULL",
                             "LIMITED",
-                            "BACKGROUND",
-                            "NONE"
+                            "BACKGROUND"
```
JackLivio, line 857: Changing default permissions could likely break someones integration. Is this really necessary?

src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -4725,42 +4726,70 @@ bool ApplicationManagerImpl::IsAppSubscribedForWayPoints(
...
+    if (mobile_way_points_data_.end() == it) {
+      SDL_LOG_DEBUG("No waypoint data provided by app service provider yet");
+      return;
+    }
+    smart_objects::SmartObjectSPtr way_point_notification =
+        std::make_shared<smart_objects::SmartObject>(it->second);
+    (*way_point_notification)[strings::params][strings::connection_key] =
         app_id;
-    GetRPCService().SendMessageToMobile(way_point_notification_);
+    GetRPCService().SendMessageToMobile(way_point_notification);
```
JackLivio, line 4752: Add return message here, or else mobile could be send two notifications?

src/components/application_manager/src/helpers/application_helper.cc
```diff
@@ -9,12 +9,16 @@ namespace {
 using namespace application_manager;
 void DeleteWayPoints(ApplicationSharedPtr app,
                      ApplicationManager& app_manager) {
-  app_manager.UnsubscribeAppFromWayPoints(app);
-  if (!app_manager.IsAnyAppSubscribedForWayPoints()) {
+  std::set<uint32_t> subscribed_apps =
+      app_manager.GetAppsSubscribedForWayPoints();
+  bool send_unsubscribe =
+      subscribed_apps.size() <= 1 && app_manager.IsSubscribedToHMIWayPoints();
```
JackLivio, line 15: Are unsubscribe requests never sent to mobile app service providers?

src/appMain/sdl_preloaded_pt.json
jacobkeeler: Not strictly, this change just made the most sense security-wise after looking into all of the cases for this issue. It doesn't really make too much sense to allow an app to publish an active service before the actual app is activated at least once.

As for breaking integrations, I really don't think that this file should be considered definitive in any way, policies are heavily configurable for a reason. If an app assumes that every integration has the same general policy configurations, it's going to run into issues anyway.

src/components/application_manager/src/helpers/application_helper.cc
jacobkeeler: As of now, correct. Same with subscribe requests. Initially I didn't see a way of doing so using the existing RPC passing handler interface, but it might be possible to do just manually using mobile events (similar to GetAppServiceData), it would just make the logic a lot more complicated. This would also have the counterintuitive consequence that the app service must NOT include `SubscribeWayPoints` or `UnsubscribeWayPoints` (this is still the case with this implementation, but at least the app service provider will never receive these messages), since the RPC passing handler would take over in that case.

I am also not certain how we would handle the case where an app service which handles waypoints registers while another app is already subscribed to waypoints.

src/components/application_manager/src/application_manager_impl.cc
jacobkeeler: Ah, should just be in an else-if

src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -4725,42 +4726,70 @@ bool ApplicationManagerImpl::IsAppSubscribedForWayPoints(
...
+  }
+  if (hmi_way_points_data_) {
```
jacobkeeler, line 4754:
```suggestion
  } else if (hmi_way_points_data_) {
```

src/components/application_manager/src/application_manager_impl.cc
jacobkeeler: Fixed in https://github.com/smartdevicelink/sdl_core/pull/3573/commits/109813ab47dcd266f0e66e66450f20eda4b6057b

src/appMain/sdl_preloaded_pt.json
JackLivio: > If an app assumes that every integration has the same general policy configurations, it's going to run into issues anyway.

Excellent point

JackLivio (approved):

JackLivio (approved):

dboltovskyi:
@jacobkeeler Please notice one UT starts failing within the fix:
```
16:47:21 [==========] 1212 tests from 578 test cases ran. (4358 ms total)
16:47:21 [  PASSED  ] 1211 tests.
16:47:21 [  FAILED  ] 1 test, listed below:
16:47:21 [  FAILED  ] HMICommandsNotificationsTest.OnNaviWayPointChangeSendNotificationToMobile
16:47:21 
16:47:21  1 FAILED TEST
```
Build Log: [SDL_build.log.tar.gz](https://github.com/smartdevicelink/sdl_core/files/5690189/SDL_build.log.tar.gz)

PR #3582, opened 2020-11-23
Fix getting SSL context in CryptoManager

### LINKED ISSUE #3421: SSL Context should use MACRO instead of direct struct access
### Bug Report
SDL Core will not build with security enabled when using OpenSSL 1.1 because the SSL context used inside of ResetConnection is directly accessing the underlying structure inside of the connection.  Instead the context should be referenced by the proper MACRO which will allow core to build with OpenSSL 1.1 as well as older versions.

Replacing the line linked below with the following code snippet resolves this issue
https://github.com/smartdevicelink/sdl_core/blob/master/src/components/security_manager/src/ssl_context_impl.cc#L599
```SSL_CTX* ssl_context = SSL_get_SSL_CTX(connection_);```

##### Reproduction Steps
Clone SDL Core and attempt to build on a system with security enabled while using OpenSSL 1.1

##### Expected Behavior
Build is performed successfully

##### Observed Behavior
Error when attempting to build

* OS/Version: Ubuntu 18.04
* SDL Core Version: 6.1.1

AKalinich-Luxoft:
Fixes #3421 

### Risk

### Testing Plan
Covered by unit tests

### Summary
This PR fixes way of getting SSL context from SSL connection. Direct access works well in 1.0.2 version, however this field
became unavailable in 1.1.1 version which causes SDL build to fail. In order to make this change working on all lib versions
as expected, direct access has been replaced with convenient getter function from the same library.

### CLA

### DIFF
--- a/src/components/security_manager/src/ssl_context_impl.cc
+++ b/src/components/security_manager/src/ssl_context_impl.cc
@@ -588,7 +588,7 @@ void CryptoManagerImpl::SSLContextImpl::ResetConnection() {
     SSL_shutdown(connection_);
   }
   SDL_LOG_DEBUG("SSL connection recreation");
-  SSL_CTX* ssl_context = connection_->ctx;
+  SSL_CTX* ssl_context = SSL_get_SSL_CTX(connection_);
   SSL_free(connection_);
   connection_ = SSL_new(ssl_context);
   if (mode_ == SERVER) {
--- a/tools/Utils/generate_test_certificates.py
+++ b/tools/Utils/generate_test_certificates.py
@@ -9,6 +9,7 @@
 import os
 import subprocess
 import tempfile
+import shutil
 from argparse import ArgumentParser
 from subprocess import check_call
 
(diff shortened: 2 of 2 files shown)


LitvinenkoIra (approved):

ychernysheva (approved):

dboltovskyi:
@JackLivio Please notice this PR is ready for PMs review.

However we would like stress the following findings regarding the update:
1. SDL can be built successfully either with `lbssl1.0` or `libssl1.1`
2. Secure connection can be established successfully with `SPT` mobile appication
3. There are issues with ATF and Test Scripts:
- ATF also requires an update to be built with `libssl1.1`
- ATF parallel mode requires an update in [Dockerfile](https://github.com/smartdevicelink/sdl_atf/blob/master/atf_parallels/docker/Dockerfile) to include `libssl-dev`
- It looks like new SSL version follow more strict rules regarding self-signed certificates that are used in Test Scripts
(from quick view it seems interim certificate has to include `CA:true` constraint). So existing certificates (please see [files/Security](https://github.com/smartdevicelink/sdl_atf_test_scripts/tree/master/files/Security)) need to be updated (or regenerated)
- Since these certificates are obtained using [generate_test_certificates.py](https://github.com/smartdevicelink/sdl_core/blob/master/tools/Utils/generate_test_certificates.py) script it might need to be updated as well

Luxoft team is going to create an appropriate updates once investigation is finished.

dboltovskyi:
@JackLivio Please notice all the issues mentioned in [3582#issuecomment-742630966](https://github.com/smartdevicelink/sdl_core/pull/3582#issuecomment-742630966) are resolved.

Hence this PR (as well as corresponding updates in ATF and Test Scripts) are ready for PMs review.

iCollin (approved):

PR #3654, opened 2021-03-05
Update OpenSSL Context Config

### LINKED ISSUE #3653: OpenSSL 1.1.1 configuration and unit test errors
### Bug Report
Recent updates for Ubuntu 20 to use openssl 1.1.1 have caused some issues for Ubuntu 18 users. By default openssl is adding tls v1.3 ciphers to the defualt list even though SDL does not support this TLS version. This also causes the validation of cipher lists to always return valid, while user specified versions are ignored if invalid. 

Also The updates made on develop to support non-deprecated tls methods cause the protocol handshake to be less restrictive.

ie: Specifying tls v1.1 in the .ini file but SDL Core can still negotiate up to tls v1.3.

Related issue from openssl github: https://github.com/openssl/openssl/issues/7196#issue-359287519

##### Reproduction Steps
1. Install openssl 1.1.1 and build core with unit tests on
2. Run unit tests: failure in security manager tests.

##### Expected Behavior
Test passes due to a failed init from an invalid cipher. 

##### Observed Behavior
Unit test failure for the security manager tests.

* OS/Version: [The OS/Version you were running when the bug occurred] Ubuntu 18
* SDL Core Version: [The version of SDL Core you were using when the bug occurred] Develop
* Testing Against: [What you tested with to observe this behavior. Proxy, App, HMI etc.] OpenSSL 1.1.1

JackLivio:
Fixes #3653 

### Risk

### Testing Plan
- Run unit tests
- Run protocol + security + DTLS atf tests

### Summary
- Only runs a unit test if the used ssl version is 1.1.0 or older. The functionality this test is testing does not apply to open ssl version 1.1.1 on ubuntu 18.
- Add protocol maximum configurations depending on the entered protocol name from the smartDeviceLink.ini

### CLA

### DIFF
--- a/src/components/security_manager/src/crypto_manager_impl.cc
+++ b/src/components/security_manager/src/crypto_manager_impl.cc
@@ -171,6 +171,7 @@ bool CryptoManagerImpl::Init() {
 #else
       SDL_LOG_DEBUG("SSLv3 is used");
       method = is_server ? SSLv3_server_method() : SSLv3_client_method();
+      SSL_CTX_set_max_proto_version(context_, SSL3_VERSION);
       break;
 #endif
     case TLSv1:
@@ -179,6 +180,7 @@ bool CryptoManagerImpl::Init() {
       method = is_server ? TLSv1_server_method() : TLSv1_client_method();
 #else
       method = is_server ? TLS_server_method() : TLS_client_method();
+      SSL_CTX_set_max_proto_version(context_, TLS1_VERSION);
 #endif
       break;
     case TLSv1_1:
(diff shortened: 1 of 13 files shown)


src/components/security_manager/src/crypto_manager_impl.cc
```diff
@@ -203,6 +206,8 @@ bool CryptoManagerImpl::Init() {
       method = is_server ? TLSv1_2_server_method() : TLSv1_2_client_method();
 #else
       method = is_server ? TLS_server_method() : TLS_client_method();
+      SSL_CTX_set_max_proto_version(context_, TLS1_2_VERSION);
+
```
jacobkeeler, line 210:
```suggestion
```

jacobkeeler (approved):

jacobkeeler (approved):

Issue #2314, opened 2018-07-02
Change post-increments to pre-increments


ghost:
### Introduction

Post-increments on iterators may be the reason of performance leaks. Also pre-increments are required by the [Google styleguide](https://google.github.io/styleguide/cppguide.html#Preincrement_and_Predecrement).

### Motivation

When a variable is incremented (++i or i++) or decremented (--i or i--) and the value of the expression is not used, one must decide whether to preincrement (decrement) or postincrement (decrement).
When the return value is ignored, the "pre" form (++i) is never less efficient than the "post" form (i++), and is often more efficient. This is because post-increment (or decrement) requires a copy of i to be made, which is the value of the expression. If i is an iterator or other non-scalar type, copying i could be expensive.

### Proposed solution

The issue is about to change the all post-increment to pre-increment because it will save the memory and won't to copy the object.

iCollin:
Closed via https://github.com/smartdevicelink/sdl_core/pull/3687

PR #3687, opened 2021-04-15
replace applicable post-operators to pre-operators

iCollin:
Fixes https://github.com/smartdevicelink/sdl_core/issues/2314

### Risk

### Testing Plan
build, unit test, atf test

### Summary
Replace all interchangeable post (in/de)crement operators with the less expensive pre version.
Affected files include all `*.cc,*.h` that are not `*_test*,src/3rd*`

### CLA

### DIFF
--- a/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/hmi/on_tts_language_change_notification.cc
+++ b/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/hmi/on_tts_language_change_notification.cc
@@ -88,9 +88,12 @@ void OnTTSLanguageChangeNotification::Run() {
   (*message_)[strings::params][strings::function_id] =
       static_cast<int32_t>(mobile_apis::FunctionID::OnLanguageChangeID);
 
-  const ApplicationSet& applications =
+  const ApplicationSet& accessor =
       application_manager_.applications().GetData();
-  for (auto app : applications) {
+  ApplicationSetIt it = accessor.begin();
+  for (; accessor.end() != it;) {
+    ApplicationSharedPtr app = *it;
+    ++it;
     (*message_)[strings::params][strings::connection_key] = app->app_id();
     SendNotificationToMobile(message_);
 
--- a/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/hmi/on_tts_language_change_notification.cc
+++ b/src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/hmi/on_tts_language_change_notification.cc
@@ -88,11 +88,9 @@ void OnTTSLanguageChangeNotification::Run() {
   (*message_)[strings::params][strings::function_id] =
       static_cast<int32_t>(mobile_apis::FunctionID::OnLanguageChangeID);
 
-  const ApplicationSet& accessor =
+  const ApplicationSet& applications =
       application_manager_.applications().GetData();
-  ApplicationSetIt it = accessor.begin();
-  for (; accessor.end() != it;) {
-    ApplicationSharedPtr app = *it++;
+  for (auto app : applications) {
     (*message_)[strings::params][strings::connection_key] = app->app_id();
     SendNotificationToMobile(message_);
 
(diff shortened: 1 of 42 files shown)


src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/hmi/on_tts_language_change_notification.cc
```diff
@@ -88,11 +88,9 @@ void OnTTSLanguageChangeNotification::Run() {
   (*message_)[strings::params][strings::function_id] =
       static_cast<int32_t>(mobile_apis::FunctionID::OnLanguageChangeID);
 
-  const ApplicationSet& accessor =
+  const ApplicationSet& applications =
       application_manager_.applications().GetData();
-  ApplicationSetIt it = accessor.begin();
-  for (; accessor.end() != it;) {
-    ApplicationSharedPtr app = *it++;
+  for (auto app : applications) {
```
jacobkeeler, line 93: I believe this didn't use a foreach loop for a reason. It looks like this loop can unregister applications, making the iterator invalid.
```suggestion
  ApplicationSetIt it = accessor.begin();
  for (; accessor.end() != it;) {
    ApplicationSharedPtr app = *it;
    ++it;
```

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/hmi/on_ui_language_change_notification.cc
```diff
@@ -80,12 +80,10 @@ void OnUILanguageChangeNotification::Run() {
...
       static_cast<int32_t>(mobile_apis::FunctionID::OnLanguageChangeID);
 
-  const ApplicationSet& accessor =
+  const ApplicationSet& applications =
       application_manager_.applications().GetData();
 
-  ApplicationSetConstIt it = accessor.begin();
-  for (; accessor.end() != it;) {
-    ApplicationSharedPtr app = *it++;
+  for (auto app : applications) {
```
jacobkeeler, line 86: Same here
```suggestion
  ApplicationSetConstIt it = accessor.begin();
  for (; accessor.end() != it;) {
    ApplicationSharedPtr app = *it;
    ++it;
```

src/components/application_manager/src/hmi_language_handler.cc
```diff
@@ -261,12 +261,10 @@ void HMILanguageHandler::VerifyWithPersistedLanguages() {
...
       "mismatch with persisted values.");
 
-  const ApplicationSet& accessor =
+  const ApplicationSet& applications =
       application_manager_.applications().GetData();
-  ApplicationSetIt it = accessor.begin();
-  for (; accessor.end() != it;) {
-    ApplicationConstSharedPtr app = *it++;
 
+  for (auto app : applications) {
```
jacobkeeler, line 267: Same here
```suggestion
  ApplicationSetIt it = accessor.begin();
  for (; accessor.end() != it;) {
    ApplicationConstSharedPtr app = *it;
    ++it;
```

jacobkeeler (changes requested):

src/components/application_manager/rpc_plugins/sdl_rpc_plugin/src/commands/hmi/on_tts_language_change_notification.cc
iCollin: I don't understand the functional difference between

```
for (auto app : applications)
vs
for (auto it = applications.begin(); it != applications.end(); ++it)
```

Am I misunderstanding your suggestion? If the iterator is invalidated I think both statements would be affected the same.

Regardless of the above, I thought this iterator is of the `DataAccessor::GetData`, a copy of the `ApplicationSet`. While `UnregisterApplication` does remove an app from the set `ApplicationManagerImpl::applications_,` it doesn't invalidate the iterator of this copy, and that's why we don't have to hold the lock on `applications_list_lock_ptr_` during the loop also.

jacobkeeler: Yeah, to be honest it might be fine as-is, but considering that this is already handling this is in a safer manner, I would rather not change it. Invalid iterator issues can be quite time-consuming to debug

jacobkeeler (approved):

Issue #2323, opened 2018-07-03
SDL must set media app to FULL and NOT_AUDIBLE in case this app is activated during phone call


HSavynetska:
### Bug Report
SDL must set media app to FULL and NOT_AUDIBLE in case this app is activated during phone call

##### Reproduction Steps
1. SDL and HMI are started.
2. Media app is registered
3. HMI sends `OnEventChanged(PHONE_CALL, isActive=true)`
3. SDL receives `SDL.ActivateApp(<appID_of_media_app>)` request from HMI during phone call

##### Expected Behavior
1. SDL must respond `SDL.ActivateApp (SUCCESS)` to HMI
2. SDL sends `OnHMIStatus(FULL, NOT_AUDIBLE)` to mobile app
3. After the phone call ends and SDL receives `OnEventChanged(PHONE_CALL, isActive=false)` from HMI, SDL must send `OnHMIStatus(FULL, AUDIBLE)` to mobile app

##### Observed Behavior
1. SDL sends `OnHMIStatus(BACKGROUND, NOT_AUDIBLE)` to mobile app during activation
2. SDL does not send `OnHMIStatus(FULL, AUDIBLE)` to mobile app after the phone call ends and SDL receives `OnEventChanged(PHONE_CALL, isActive=false)` from HMI.

##### OS & Version Information
* OS/Version: Ubuntu 16.04
* SDL Core Version: develop branch (ba1da7b)
* Testing Against: ATF: [develop branch](https://github.com/smartdevicelink/sdl_atf/tree/develop)
Script: https://github.com/smartdevicelink/sdl_atf_test_scripts/pull/1956

##### Attach
[TestingReports.zip](https://github.com/smartdevicelink/sdl_core/files/2159501/TestingReports.zip)

HSavynetska:
Environment:
Script: smartdevicelink/sdl_atf_test_scripts#1956
SDL build: develop branch (ba1da7b)
ATF: develop branch

Status: defect is reproduced.

VjKlepikov:
The issue is relevant on the current develop branch

Environment:
Test script: smartdevicelink/sdl_atf_test_scripts#1956
SDL build: https://github.com/smartdevicelink/sdl_core/tree/develop ( 891497b )
ATF: develop branch

dboltovskyi:
Please notice issue is still actual for Core 6.1
[...]

PR #3710, opened 2021-06-01
allow an app to be activated while phone call is in progress

iCollin:
Fixes https://github.com/smartdevicelink/sdl_core/issues/2323

### Risk

### Testing Plan
unit tests, atf, issue steps manually

### Summary
Max HMI level during phone call is FULL, and normal hmi level logic added.

### CLA

### DIFF
--- a/src/components/application_manager/src/hmi_state.cc
+++ b/src/components/application_manager/src/hmi_state.cc
@@ -243,28 +243,28 @@ PhoneCallHmiState::PhoneCallHmiState(std::shared_ptr<Application> app,
 
 mobile_apis::HMILevel::eType PhoneCallHmiState::hmi_level() const {
   using namespace mobile_apis;
-  if (HMILevel::INVALID_ENUM == parent_hmi_level()) {
+  if (HMILevel::INVALID_ENUM == parent_hmi_level() ||
+      HMILevel::HMI_NONE == parent_hmi_level()) {
     return parent_hmi_level();
   }
-  return std::max(parent_hmi_level(), max_hmi_level());
-}
-
-mobile_apis::HMILevel::eType PhoneCallHmiState::max_hmi_level() const {
-  using namespace helpers;
-  using namespace mobile_apis;
 
   if (WindowType::WIDGET == window_type()) {
     return std::max(HMILevel::HMI_FULL, parent_max_hmi_level());
   }
 
-  auto expected = HMILevel::HMI_FULL;
+  auto expected = HMILevel::HMI_BACKGROUND;
   if (is_navi_app() || is_mobile_projection_app()) {
-    expected = HMILevel::HMI_LIMITED;
-  } else if (is_media_app() || is_voice_communication_app()) {
-    expected = HMILevel::HMI_BACKGROUND;
+    if (VideoStreamingState::STREAMABLE == video_streaming_state()) {
+      expected = HMILevel::HMI_LIMITED;
+    }
   }
 
   return std::max(expected, parent_max_hmi_level());
+}
+
+mobile_apis::HMILevel::eType PhoneCallHmiState::max_hmi_level() const {
+  using namespace mobile_apis;
+  return std::max(HMILevel::HMI_FULL, parent_max_hmi_level());
 }
 
 SafetyModeHmiState::SafetyModeHmiState(std::shared_ptr<Application> app,
--- a/src/components/application_manager/src/hmi_state.cc
+++ b/src/components/application_manager/src/hmi_state.cc
@@ -243,8 +243,7 @@ PhoneCallHmiState::PhoneCallHmiState(std::shared_ptr<Application> app,
 
 mobile_apis::HMILevel::eType PhoneCallHmiState::hmi_level() const {
   using namespace mobile_apis;
-  if (HMILevel::INVALID_ENUM == parent_hmi_level() ||
-      HMILevel::HMI_NONE == parent_hmi_level()) {
+  if (HMILevel::INVALID_ENUM == parent_hmi_level()) {
     return parent_hmi_level();
   }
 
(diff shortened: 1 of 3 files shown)


src/components/application_manager/src/hmi_state.cc
```diff
@@ -243,30 +243,30 @@ PhoneCallHmiState::PhoneCallHmiState(std::shared_ptr<Application> app,
...
   }
 
   return std::max(expected, parent_max_hmi_level());
 }
 
+mobile_apis::HMILevel::eType PhoneCallHmiState::max_hmi_level() const {
+  using namespace mobile_apis;
+  return std::max(HMILevel::HMI_FULL, parent_max_hmi_level());
+}
+
```
jacobkeeler, line 269:
```suggestion
```
This can be removed, because this matches the default definition of this function. You will also need to remove the following line from the PhoneCallHmiState definition in `hmi_state.h`:
```
mobile_apis::HMILevel::eType max_hmi_level() const OVERRIDE;
```

src/components/application_manager/src/hmi_state.cc
```diff
@@ -243,30 +243,30 @@ PhoneCallHmiState::PhoneCallHmiState(std::shared_ptr<Application> app,
...
     return std::max(HMILevel::HMI_FULL, parent_max_hmi_level());
   }
 
-  auto expected = HMILevel::HMI_FULL;
+  auto expected = HMILevel::HMI_BACKGROUND;
   if (is_navi_app() || is_mobile_projection_app()) {
-    expected = HMILevel::HMI_LIMITED;
-  } else if (is_media_app() || is_voice_communication_app()) {
-    expected = HMILevel::HMI_BACKGROUND;
+    if (VideoStreamingState::STREAMABLE == video_streaming_state()) {
```
jacobkeeler, line 257: I don't think that the video streaming state should be checked here. I also think that generally this state is derived from the HMI level and app type, not the other way around.

src/components/application_manager/src/hmi_state.cc
```diff
@@ -243,30 +243,30 @@ PhoneCallHmiState::PhoneCallHmiState(std::shared_ptr<Application> app,
...
-mobile_apis::HMILevel::eType PhoneCallHmiState::max_hmi_level() const {
-  using namespace helpers;
-  using namespace mobile_apis;
 
   if (WindowType::WIDGET == window_type()) {
     return std::max(HMILevel::HMI_FULL, parent_max_hmi_level());
   }
 
-  auto expected = HMILevel::HMI_FULL;
+  auto expected = HMILevel::HMI_BACKGROUND;
```
jacobkeeler, line 255: Is there a reason that non-media apps should be deactivated by this event? I think they should still be allowed to be in FULL, as the phone call could be handled in the background. In addition, this was set up to follow previous behavior, so I don't know if it should be changed without a decent reason.

jacobkeeler: That said, I could see an argument to not change the HMI level at all, since the phone call could be started in the background without ever deactivating the app. Perhaps this event should ONLY affect audioStreamingState?

jacobkeeler (changes requested):

src/components/application_manager/src/hmi_state.cc
```diff
@@ -243,30 +243,30 @@ PhoneCallHmiState::PhoneCallHmiState(std::shared_ptr<Application> app,
 
 mobile_apis::HMILevel::eType PhoneCallHmiState::hmi_level() const {
   using namespace mobile_apis;
-  if (HMILevel::INVALID_ENUM == parent_hmi_level()) {
+  if (HMILevel::INVALID_ENUM == parent_hmi_level() ||
+      HMILevel::HMI_NONE == parent_hmi_level()) {
```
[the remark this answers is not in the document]
iCollin: Removed by https://github.com/smartdevicelink/sdl_core/pull/3710/commits/38f664c4bd7691f3208387d718ba78524eccb135

src/components/application_manager/src/hmi_state.cc
```diff
@@ -243,30 +243,30 @@ PhoneCallHmiState::PhoneCallHmiState(std::shared_ptr<Application> app,
...
   if (is_navi_app() || is_mobile_projection_app()) {
-    expected = HMILevel::HMI_LIMITED;
-  } else if (is_media_app() || is_voice_communication_app()) {
-    expected = HMILevel::HMI_BACKGROUND;
+    if (VideoStreamingState::STREAMABLE == video_streaming_state()) {
+      expected = HMILevel::HMI_LIMITED;
+    }
   }
 
   return std::max(expected, parent_max_hmi_level());
```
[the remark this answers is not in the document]
iCollin: Updated in https://github.com/smartdevicelink/sdl_core/pull/3710/commits/38f664c4bd7691f3208387d718ba78524eccb135

src/components/application_manager/src/hmi_state.cc
iCollin: Removed in https://github.com/smartdevicelink/sdl_core/pull/3710/commits/38f664c4bd7691f3208387d718ba78524eccb135

src/components/application_manager/src/hmi_state.cc
iCollin: That makes sense to me, but without this check NAVI apps that are not STREAMABLE will be set to LIMITED HMI level during a phone call, and my understanding was that LIMITED is only valid when an app is at least one of AUDIBLE or STREAMABLE. How can I prevent this without checking the video streaming state?

src/components/application_manager/src/hmi_state.cc
iCollin: Currently all other types of apps are deactivated by this event, and so I feel that non-media apps should be deactivated as well. Do you agree that having all apps deactivated is better than having most deactivated?

I prefer the idea of having Phone Call event only affect audible status from an HMI perspective, I think it makes the event simpler to use, but this would be a larger change. I couldn't think of a great reason to make this change, could you?

jacobkeeler: Turns out this is already part of an accepted proposal: https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0264-Separating-the-change-of-Audible-status-and-the-change-of-HMI-Status.md

iCollin:
Closed as issue will be fixed with the implementation of [proposal 0264](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0264-Separating-the-change-of-Audible-status-and-the-change-of-HMI-Status.md).

Issue #1941, opened 2017-12-06
[SDL 0117] Configurable time before shutdown


theresalech:
Proposal: [Configurable time before shutdown](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0117-configurable-time-before-shutdown.md)

> To prevent missing logs after SDL shutdown, additional ini file options should be added : 
 - Write all logs to file system before shutdown 
 - Maximum time of SDL shutting down

Review: https://github.com/smartdevicelink/sdl_evolution/issues/355

theresalech:
The Steering Committee voted to remove this feature from the Core 8.0 release during the 2021-08-03 meeting.

PR #3740, opened 2021-07-19
Configurable time before shutdown 

VladSemenyuk:
Implements #1941

### Risk

### Testing Plan
ATF scripts

### Summary
Add Logger config settings with params for configurable time before shutdown.
Update SDL config.
Extend LoggerImpl class for shutdown configuration.
Extend MessageLoopThread, MessageQueque and ConditionlVariable classes with new wait mechanism.

### CLA
- [ ] I have signed [the CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform)

### DIFF
--- a/src/appMain/main.cc
+++ b/src/appMain/main.cc
@@ -162,6 +162,7 @@ int32_t main(int32_t argc, char** argv) {
 #endif  // LOG4CXX_LOGGER
 
     logger_impl->Init(std::move(logger));
+    logger_impl->InitLoggerSettings(&profile_instance);
   }
 #endif
 
--- a/src/appMain/smartDeviceLink.ini
+++ b/src/appMain/smartDeviceLink.ini
@@ -411,3 +411,9 @@ RpcPassThroughTimeout = 10000
 [RCModuleConsent]
 ; The period (in days) after which consent for module_id should be removed.
 PeriodForConsentExpiration = 30
+
+[Logger]
+; Write all logs in queue to file system before shutdown 
+FlushLogMessagesBeforeShutdown = false
+; Maximum time to wait for writing all data before exit SDL in seconds
+MaxTimeBeforeShutdown = 30
--- a/src/components/config_profile/include/config_profile/profile.h
+++ b/src/components/config_profile/include/config_profile/profile.h
@@ -46,6 +46,7 @@
 #include "protocol_handler/protocol_handler_settings.h"
 #include "smart_objects/smart_object.h"
 #include "transport_manager/transport_manager_settings.h"
+#include "utils/logger/logger_settings.h"
 #include "utils/macro.h"
 
 namespace profile {
@@ -59,7 +60,8 @@ class Profile : public protocol_handler::ProtocolHandlerSettings,
                 public media_manager::MediaManagerSettings,
                 public policy::PolicySettings,
                 public transport_manager::TransportManagerSettings,
-                public application_manager::ApplicationManagerSettings {
+                public application_manager::ApplicationManagerSettings,
+                public logger::LoggerSettings {
  public:
   // Methods section
 
@@ -589,6 +591,18 @@ class Profile : public protocol_handler::ProtocolHandlerSettings,
 
   const std::vector<std::string>& embedded_services() const OVERRIDE;
   const std::string hmi_origin_id() const OVERRIDE;
+
+  /**
+   * @brief  Returns true if writing all logs to file system before shutdown is
+   * enabled
+   */
+  bool flush_log_messages_before_shutdown() const OVERRIDE;
+  /**
+   * @brief  Returns waximum time to wait for writing all data before exit SDL
+   * in seconds
+   */
+  uint16_t max_time_before_shutdown() const OVERRIDE;
+
   /**
    * @brief Reads a string value from the profile
    *
@@ -1164,6 +1178,8 @@ class Profile : public protocol_handler::ProtocolHandlerSettings,
   int ignition_off_signal_offset_;
   uint32_t rpc_pass_through_timeout_;
   uint16_t period_for_consent_expiration_;
+  bool flush_log_messages_before_shutdown_;
+  uint16_t max_time_before_shutdown_;
 
   std::vector<std::string> embedded_services_;
 
--- a/src/components/utils/include/utils/logger/logger_impl.h
+++ b/src/components/utils/include/utils/logger/logger_impl.h
@@ -74,10 +74,14 @@ class LoggerImpl : public Logger, public LoggerInitializer {
   bool IsEnabledFor(const std::string& component,
                     LogLevel log_level) const override;
   void PushLog(const LogMessage& log_message) override;
+  void InitLoggerSettings(LoggerSettings* settings) override;
+  void InitFlushLogsTimePoint(const TimePoint& time_point) override;
 
   std::unique_ptr<ThirdPartyLoggerInterface> impl_;
+  LoggerSettings* settings_;
   LoopThreadPtr<LogMessageLoopThread> loop_thread_;
   bool use_message_loop_thread_;
+  TimePoint flush_logs_time_point_;
 };
 
 }  // namespace logger
--- a/src/components/utils/include/utils/logger/logger_settings.h
+++ b/src/components/utils/include/utils/logger/logger_settings.h
@@ -1,0 +1,15 @@
+#ifndef SRC_COMPONENTS_UTILS_INCLUDE_UTILS_LOGGER_SETTINGS_H_
+#define SRC_COMPONENTS_UTILS_INCLUDE_UTILS_LOGGER_SETTINGS_H_
+
+#include <stdint.h>
+
+namespace logger {
+class LoggerSettings {
+ public:
+  virtual bool flush_log_messages_before_shutdown() const = 0;
+  virtual uint16_t max_time_before_shutdown() const = 0;
+};
+
+}  // namespace logger
+
+#endif  // SRC_COMPONENTS_UTILS_INCLUDE_UTILS_LOGGER_SETTINGS_H_
--- a/src/components/utils/src/logger/logger_impl.cc
+++ b/src/components/utils/src/logger/logger_impl.cc
@@ -37,7 +37,9 @@
 namespace logger {
 
 LoggerImpl::LoggerImpl(bool use_message_loop_thread)
-    : impl_(nullptr), use_message_loop_thread_(use_message_loop_thread) {}
+    : impl_(nullptr)
+    , settings_(nullptr)
+    , use_message_loop_thread_(use_message_loop_thread) {}
 
 void LoggerImpl::Init(std::unique_ptr<ThirdPartyLoggerInterface>&& impl) {
   assert(impl_ == nullptr);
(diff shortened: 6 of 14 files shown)


src/components/utils/src/logger/logger_impl.cc
```diff
@@ -71,9 +73,20 @@ void LoggerImpl::DeInit() {
 }
 
 void LoggerImpl::Flush() {
-  if (use_message_loop_thread_) {
-    if (loop_thread_) {
+  if (use_message_loop_thread_ && loop_thread_) {
+    assert(settings_);
+
+    if (0 == flush_logs_time_point_.time_since_epoch().count()) {
```
yang1070, line 79: why do we have such a condition? it seems impossible to me that count() is zero, thus impossible to execute next line

src/components/utils/include/utils/logger/logger_impl.h
```diff
@@ -74,10 +74,14 @@ class LoggerImpl : public Logger, public LoggerInitializer {
...
                     LogLevel log_level) const override;
   void PushLog(const LogMessage& log_message) override;
+  void InitLoggerSettings(LoggerSettings* settings) override;
+  void InitFlushLogsTimePoint(const TimePoint& time_point) override;
 
   std::unique_ptr<ThirdPartyLoggerInterface> impl_;
+  LoggerSettings* settings_;
   LoopThreadPtr<LogMessageLoopThread> loop_thread_;
   bool use_message_loop_thread_;
+  TimePoint flush_logs_time_point_;
```
yang1070, line 84: what's the default value of flush_logs_time_point_ if it is not initialized?

src/appMain/main.cc
```diff
@@ -162,6 +162,7 @@ int32_t main(int32_t argc, char** argv) {
 #endif  // LOG4CXX_LOGGER
```
yang1070, line 162: it is marked "not ready for review" in the PR summary , is that correct?

src/appMain/smartDeviceLink.ini
```diff
@@ -411,3 +411,9 @@ RpcPassThroughTimeout = 10000
 [RCModuleConsent]
 ; The period (in days) after which consent for module_id should be removed.
 PeriodForConsentExpiration = 30
+
+[Logger]
+; Write all logs in queue to file system before shutdown
```
yang1070, line 416: the comment "write all logs in queue" may not be correct, and it is misleading

src/components/utils/include/utils/logger/logger_impl.h
VladSemenyuk: Default constructor, creates a time_point representing the Clock's epoch (i.e., time_since_epoch() is zero).

src/components/utils/src/logger/logger_impl.cc
VladSemenyuk: Default constructor, creates a time_point representing the Clock's epoch (i.e., time_since_epoch() is zero).

yang1070: @VladSemenyuk in file on_exit_all_applications_notification.cc and function void OnExitAllApplicationsNotification::Run()
it calls SDL_INIT_FLUSH_LOGS_TIME_POINT(std::chrono::system_clock::now()); to set the value of flush_logs_time_point_ to non-zero, what's the condition that next line get executed? during the SDL running without receiving ExitAllApplicationsNotification?

VladSemenyuk: Yes, ExitAllApplicationsNotification is not the only exit point for SDL.

src/appMain/smartDeviceLink.ini
VladSemenyuk: Fixed

src/components/utils/src/logger/logger_impl.cc
yang1070: @VladSemenyuk could you like what are other conditions or exit point to call logger Flush()?

src/appMain/main.cc
KhrystynaDubovyk: This means the PR is not ready for the Project Maintainer review.

src/components/utils/src/logger/logger_impl.cc
VladSemenyuk: @yang1070 
When the following signals are thrown:
SIGINT
SIGTERM
SIGSEGV
or
"ASSERT" situation.

yang1070: @VladSemenyuk 
from the signals list and some internet search
The SIGINT (“program interrupt”) signal is sent when the user types the INTR character (normally C-c ).
the SIGSEGV is caused by segmentation fault
The SIGTERM signal is a generic signal used to cause program termination. Unlike SIGKILL , this signal can be blocked, handled, and ignored. It is the normal way to politely ask a program to terminate. The shell command kill generates SIGTERM by default.
ASSERT 

@VladSemenyuk it seems all other exit points (except IGNITION_OFF signal ) are not NORMAL exit points, and in those cases flush_logs_time_point_ will be zero block,  is calling loop_thread_->WaitDumpQueue() and wait all logs are written to disk a correct choice?
probably not. for example after segment fault, or after get a ctrl-c

src/appMain/main.cc
yang1070: @KhrystynaDubovyk got it, than you

src/components/utils/src/logger/logger_impl.cc
yang1070: @VladSemenyuk  if without this change, loop_thread_->WaitDumpQueue() is called when condition (use_message_loop_thread_ && loop_thread_) is true, 
will  the signal  IGNITION_OFF cause SDL to call LoggerImpl::Flush(), thus get WaitDumpQueue get called?

if it is yes, then it is contrary to the following statement in the proposal 

In the current implementation, after receiving IGNITION_OFF signal, SDL drops all logs that have not yet been written. Such behavior sometimes prevents analyzing of SDL issues that are found by the ATF script, and requires adding extra timeouts in test scripts before SDL shutdown.

if it is no, in this change where is the logic or code that make sure after receiving IGNITION_OFF signal, LoggerImpl::Flush() is called?

src/components/utils/include/utils/logger/logger_impl.h
yang1070: @VladSemenyuk Got it. Thank you

src/components/config_profile/include/config_profile/profile.h
```diff
@@ -589,6 +591,18 @@ class Profile : public protocol_handler::ProtocolHandlerSettings,
...
   const std::vector<std::string>& embedded_services() const OVERRIDE;
   const std::string hmi_origin_id() const OVERRIDE;
+
+  /**
+   * @brief  Returns true if writing all logs to file system before shutdown is
+   * enabled
+   */
+  bool flush_log_messages_before_shutdown() const OVERRIDE;
+  /**
+   * @brief  Returns waximum time to wait for writing all data before exit SDL
```
yang1070, line 601: typo waximum

src/components/config_profile/include/config_profile/profile.h
```diff
@@ -589,6 +591,18 @@ class Profile : public protocol_handler::ProtocolHandlerSettings,
 
   const std::vector<std::string>& embedded_services() const OVERRIDE;
   const std::string hmi_origin_id() const OVERRIDE;
+
+  /**
+   * @brief  Returns true if writing all logs to file system before shutdown is
```
yang1070, line 596: please remove "all" in comment

src/components/utils/src/logger/logger_impl.cc
VladSemenyuk: @yang1070 The proposal is old and at the time it was written the logger worked differently, and now it does`nt match the proposal. So before this implementation SDL flushed all logs always, and now configurable timeout shutdown, for IGNITION_OFF situation is added.

yang1070: @VladSemenyuk in that case I'm not sure the proposal is still valid and whether or not we need this change.

the proposal solves the problem that SDL drops off log and give options to continue write the log with maximum allowed time configured by the SDL settings in ini file.

Now (before this PR), SDL get some update, it always flush the log, SDL no long drop off the log, the old problem does not exist any more.  The problem may change to SDL takes too much time for flushing the logs? Why we still need to implement the proposal?

src/components/config_profile/include/config_profile/profile.h
VladSemenyuk: @yang1070 Done

src/components/config_profile/include/config_profile/profile.h
VladSemenyuk: @yang1070 Done

src/components/utils/src/logger/logger_impl.cc
VladSemenyuk: > The problem may change to SDL takes too much time for flushing the logs? Why we still need to implement the proposal?

Yes. Problem is it takes  too much time for flushing the logs in some cases on weak environment.

yang1070: @VladSemenyuk in that case, I would like to see an update of SDL-117 and get the update approved first

yang1070: since the default SDL behavior is to flush log,
as in the code change, the setting `FlushLogMessagesBeforeShutdown` will only impact the normal IGNITION_OFF situation.  the name shall be change to something else that is more proper and can descript the situation

ShobhitAd (changes requested):

src/components/config_profile/include/config_profile/profile.h
```diff
@@ -589,6 +591,18 @@ class Profile : public protocol_handler::ProtocolHandlerSettings,
...
   const std::vector<std::string>& embedded_services() const OVERRIDE;
   const std::string hmi_origin_id() const OVERRIDE;
+
+  /**
+   * @brief  Returns true if writing logs to file system before shutdown is
+   * enabled
+   */
+  bool flush_log_messages_before_shutdown() const OVERRIDE;
+  /**
+   * @brief  Returns maximum time to wait for writing all data before exit SDL
```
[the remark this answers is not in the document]
VladSemenyuk: Fixed

src/components/config_profile/include/config_profile/profile.h
```diff
@@ -589,6 +591,18 @@ class Profile : public protocol_handler::ProtocolHandlerSettings,
 
   const std::vector<std::string>& embedded_services() const OVERRIDE;
   const std::string hmi_origin_id() const OVERRIDE;
+
+  /**
+   * @brief  Returns true if writing logs to file system before shutdown is
```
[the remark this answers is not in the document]
VladSemenyuk: Fixed

src/components/utils/include/utils/logger/logger_settings.h
```diff
@@ -0,0 +1,15 @@
+#ifndef SRC_COMPONENTS_UTILS_INCLUDE_UTILS_LOGGER_SETTINGS_H_
+#define SRC_COMPONENTS_UTILS_INCLUDE_UTILS_LOGGER_SETTINGS_H_
+
+#include <stdint.h>
```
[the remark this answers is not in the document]
VladSemenyuk: Fixed

PR #3773, opened 2021-09-10
Add missing internal error codes to security manager

jacobkeeler:
Adds missing security query specification details from https://github.com/smartdevicelink/protocol_spec/issues/40

### Risk

### Testing Plan
HANDSHAKE_FAILED - can be triggered by setting `VerifyPeer=true` using the Java Security Lib [Example Certificates](https://github.com/smartdevicelink/sdl_security_java_suite/tree/master/resources/certs)
INVALID_CERT - can be triggered by sending an app certificate with an app ID which is different from the one that the app registered with
EXPIRED_CERT - can be triggered by sending an expired app certificate (or by modifying the GetSystemTime response from the HMI)

### Summary
Adds logic for missing error codes in SendInternalError query (HANDSHAKE_FAILED, ERROR_INVALID_CERT, and ERROR_EXPIRED_CERT)

##### Bug Fixes
* Adds logic for missing error codes in SendInternalError query

### CLA

### DIFF
--- a/src/components/security_manager/src/security_manager_impl.cc
+++ b/src/components/security_manager/src/security_manager_impl.cc
@@ -557,6 +557,7 @@ bool SecurityManagerImpl::ProcessHandshakeData(
     // no handshake data to send
     return false;
   }
+
   if (sslContext->IsInitCompleted()) {
     // On handshake success
     SDL_LOG_DEBUG("SSL initialization finished success.");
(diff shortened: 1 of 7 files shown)


src/components/security_manager/src/security_manager_impl.cc
```diff
@@ -565,6 +566,29 @@ bool SecurityManagerImpl::ProcessHandshakeData(
...
     SDL_LOG_WARN("SSL initialization finished with fail.");
+    int32_t error_code = ERROR_HANDSHAKE_FAILED;
+    std::string error_text = "Handshake failed";
+    switch (handshake_result) {
+      case SSLContext::Handshake_Result_CertExpired:
+        error_code = ERROR_EXPIRED_CERT;
+        error_text = "Certificate is expired";
+        break;
+      case SSLContext::Handshake_Result_NotYetValid:
+        error_code = ERROR_EXPIRED_CERT;
```
JackLivio, line 577: I think a different error code should be returned for this one. Maybe just the default or ERROR_INVALID_CERT since the cert will eventually be valid. It is not expired so this could cause some confusion for people debugging in this case.
```suggestion
        error_code = ERROR_INVALID_CERT;
```

src/components/security_manager/test/security_manager_test.cc
```diff
@@ -759,7 +760,15 @@ TEST_F(SecurityManagerTest, ProcessHandshakeData_Answer) {
                   RawMessageEqSize(raw_message_size), false, kIsFinal))
       .Times(handshake_emulates)
       .WillRepeatedly(NotifyTestAsyncWaiter(waiter));
-  times += handshake_emulates;
```
JackLivio, line 762: Please add unit test cases for other possible error codes added in this pr

JackLivio (approved):

PR #3916, opened 2022-05-18
Fix/Reset VRHelp array to default values instead of null

### LINKED ISSUE #3909: Generated vrHelp is not properly resumed after ResetGlobalProperties
### Bug Report
Autogenerated vrHelp is not resumed of ResetGlobalProperties was sent previously

##### Reproduction Steps
1. Register app
2. Send AddCommand
3. Send SetGlobalProperties with vrHelp & vrHelpTitle
4. Send ResetGlobalProperties with VRHELPTITLE and VRHELPITEMS (HMI receives autogenerated vrHelp and vrHelpTitle)
5. Disconnect and reconnect app with correct resumption hash ID

##### Expected Behavior
HMI receives autogenerated vrHelp and vrHelpTitle based on commands

##### Observed Behavior
Core resumes vrHelpTitle, but does not provide generated vrHelp

* OS/Version: Ubuntu 20.04
* SDL Core Version: master (8.1.0)
* Testing Against: RPC Builder App JS and Generic HMI (master)

ShobhitAd:
Fixes #3909 

### Risk

### Testing Plan
- Manual testing with RPC Builder with the provided reproduction steps

### Summary
At the moment, when core receives a `ResetGlobalProperties(VRHELPITEMS)` the `ApplicationManagerImpl::ResetVrHelpTitleItems` function resets the apps `vrHelp` value to null but sends different default values in the `UI.SetGlobalProperties` sent to the HMI. 

This PR adds changes to reset `vrHelp` to default values rather than null

### CLA

### DIFF
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -3991,8 +3991,9 @@ bool ApplicationManagerImpl::ResetVrHelpTitleItems(
   smart_objects::SmartObjectSPtr so_vr_help =
       MessageHelper::CreateAppVrHelp(app);
   if (!so_vr_help->keyExists(strings::vr_help)) {
-    SDL_LOG_WARN("Failed to create vr_help items");
-    return false;
+    SDL_LOG_WARN("Failed to create vr_help items. Resetting to empty array");
+    (*so_vr_help)[strings::vr_help] =
+        smart_objects::SmartObject(smart_objects::SmartType_Array);
   }
   app->set_vr_help((*so_vr_help)[strings::vr_help]);
 
--- a/src/components/application_manager/src/application_manager_impl.cc
+++ b/src/components/application_manager/src/application_manager_impl.cc
@@ -3986,8 +3986,15 @@ bool ApplicationManagerImpl::ResetVrHelpTitleItems(
   smart_objects::SmartObject so_vr_help_title(vr_help_title);
 
   app->reset_vr_help_title();
-  app->reset_vr_help();
   app->set_vr_help_title(so_vr_help_title);
+
+  app->reset_vr_help();
+  smart_objects::SmartObjectSPtr vr_help = MessageHelper::CreateAppVrHelp(app);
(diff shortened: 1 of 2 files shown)


jacobkeeler (changes requested):
Seems that this logic existed in `MessageHelper::CreateUIResetGlobalPropertiesRequest`, but for some reason the generated values were only sent in the UI message, rather than being set in the application object. As part of this PR I would change:

```
  if (reset_result.vr_help_title_items) {
    smart_objects::SmartObjectSPtr vr_help = CreateAppVrHelp(application);
    if (!vr_help.get()) {
      SDL_LOG_WARN("Failed to create vr_help");
      return smart_objects::SmartObjectSPtr();
    } else {
      ui_reset_global_prop_request = vr_help;
    }
  }
```

```
  if (reset_result.vr_help_title_items) {
    (*ui_reset_global_prop_request)[strings::vr_help_title] =
        *(application->vr_help_title());
    (*ui_reset_global_prop_request)[strings::vr_help] =
        *(application->vr_help());
  }

```

This would make it so this is only generated in one place

src/components/application_manager/src/application_manager_impl.cc
```diff
@@ -3984,11 +3984,18 @@ bool ApplicationManagerImpl::ResetVrHelpTitleItems(
...
   app->reset_vr_help_title();
-  app->reset_vr_help();
   app->set_vr_help_title(so_vr_help_title);
 
+  app->reset_vr_help();
+  smart_objects::SmartObjectSPtr so_vr_help =
+      MessageHelper::CreateAppVrHelp(app);
+  if (!so_vr_help->keyExists(strings::vr_help)) {
+    SDL_LOG_WARN("Failed to create vr_help items");
+    return false;
```
jacobkeeler, line 3995: Eh, I don't know whether we should return false here, since we did at least clear out `vr_help`. I will note that this case doesn't seem possible at the moment, so I suppose it doesn't matter too much

jacobkeeler (approved):

PR #3920, opened 2022-06-02
Make shared pointer atomic

### LINKED ISSUE #3871: Core crash during App unregistered with PROTOCOL_VIOLATION
### Bug Report
Core crash during App unregistered with PROTOCOL_VIOLATION

##### Reproduction Steps
1. Set StopStreamingTimeout = 3000 in SDL .INI file
2. Start SDL, HMI, connect Mobile device
3. Register PROJECTION (App_1) and NAVIGATION (App_2) applications
4. Activate App_1 and start Video streaming
5. Deactivate App_1 (streaming still continue since app has STREAMABLE state)
6. Activate App_2

##### Expected Behavior
SDL does:
send OnHMIStatus(FULL, STREAMABLE) notification to App_2
  send OnHMIStatus(LIMITED, NOT_STREAMABLE) notification to App_1
  send Navi.OnVideoDataStreaming(false) notification to HMI
  start 1st 'StopStreamingTimeout' timeout
  
 Once timeout is expired SDL does:
     - send Navi.StopStream(App_1) request to HMI
     - send EndService(VIDEO) request to App_1
     - start 2nd 'StopStreamingTimeout' timeout

##### Observed Behavior
Core crash

OS/Version: Ubuntu 18.04
SDL Core Version: 8.0 + develop (8.1 release candidate)

./test_scripts/Defects/6_1/3139_03_Stop_streaming_forcibly_app_continue_streaming.lua

##### Logs:
[2022-02-28_05-18-50.845_failed.tar.gz](https://github.com/smartdevicelink/sdl_core/files/8164132/2022-02-28_05-18-50.845_failed.tar.gz)

VladSemenyuk:
Fixes #3871

### Risk

### Testing Plan
Atf scripts

### Summary
This is actually a template specialization proposed here:
https://en.cppreference.com/w/cpp/memory/shared_ptr/atomic2

If multiple threads of execution access the same std::shared_ptr object without synchronization and any of those accesses uses a non-const member function of shared_ptr then a data race will occur unless all such access is performed through an instance of std::atomic<<std::shared_ptr>>

To avoid these data races, shared pointers itself were made atomic

### CLA


VladSemenyuk:
@theresalech this PR is ready for Livio review

src/components/utils/test/atomic_object_test.cc
```diff
@@ -0,0 +1,56 @@
...
+
+  var = 8;
+  EXPECT_EQ(8, static_cast<int>(var));
+
+  sync_primitives::atomic_bool flag = true;
+
+  EXPECT_EQ(true, static_cast<bool>(flag));
+
+  flag = false;
+  EXPECT_EQ(false, static_cast<bool>(flag));
```
ShobhitAd, line 52: Might be worth adding another case with a pointer or non primitive type (like `sync_primitives::Atomic<std::string>`)

src/components/utils/test/atomic_object_test.cc
```diff
@@ -0,0 +1,56 @@
...
+}
+
```
ShobhitAd, line 54: Might need to expand the unit tests to test the implementation of the Atomic class. Perhaps a case with multiple threads with one setting the value and another trying to access it

ShobhitAd (changes requested):

src/components/utils/test/atomic_object_test.cc
VladSemenyuk: Done

src/components/utils/test/atomic_object_test.cc
VladSemenyuk: I don't see a good solution for simultaneous access to the atomic object from different threads in unit test.

ShobhitAd (approved):

Issue #3857, opened 2022-02-14
SDL core dump on Ignition OFF during registering Apps


VjKlepikov:
## Bug Report
SDL core dump on Ignition OFF during registering Apps

##### Occurrences:
~10%

##### Reproduction Steps
1. SDL and HMI are started
2. Connect mobile device via USB transport.
3. Perform ignition Off during registering Apps (25 Apps)
On "HMI" go to → Click the "Exit Application" button → Select “IGNITION_OFF” in “Exit Application reason” → Click the `Send state` button

##### Expected Behavior
SDL starts the ignition off sequence
1. SDL sends to HMI OnSDLClose and stops working
SDL->HMI: BasicCommunication.OnSDLClose:{}

##### Observed Behavior
SDL starts the ignition off sequence
1. SDL sends to HMI OnSDLClose and stops working
SDL->HMI: BasicCommunication.OnSDLClose:{}
Core dump is observed
SIGSEGV caught
DEBUG [139969695552064][SDLMain] /home/jenkins/workspace/Develop_SDL_NoUT_P/src/appMain/life_cycle_impl.cc:241 {anonymous}::sig_handler: SIGSEGV signal has been caught

##### OS & Version Information
* OS/Version: Ubuntu 18
* SDL Core Version:  https://github.com/smartdevicelink/sdl_core/tree/develop (7a4a4e0)

##### Logs:
[Logs.zip](https://github.com/smartdevicelink/sdl_core/files/8061975/Logs.zip)

VjKlepikov:
Please notice this is a non-regression issue. It's also reproduced on 8.0

VjKlepikov:
@theresalech Please notice Luxoft team is going to prepare a fix

ShobhitAd:
@VjKlepikov I was able to reproduce this issue on my end manually but might need an ATF tests PR to more easily test the fix PR.

VjKlepikov:
@ShobhitAd Please find PR with  the script in https://github.com/smartdevicelink/sdl_atf_test_scripts/pull/2694

ShobhitAd:
Closed via https://github.com/smartdevicelink/sdl_core/pull/3939

PR #3928, opened 2022-06-16
Core dump on ignition off during registering 20 apps

RomanReznichenkoLuxoft:
Fixes #[3857](https://github.com/smartdevicelink/sdl_core/issues/3857)

### Risk

### Testing Plan
Manual test using steps in issue description

### Summary
Problems I fixed.

The first problem. Аn object of this class AppExtension was not given externally without protection. Therefore, I will change the return type from Extensions () to DataAccessor <std :: list <AppExtensionPtr>>. Where the desired mutex is captured.

Second problem. ApplicationManagerImpl had unprotected deletion of objects used in other threads. Also, I found a duplicate of the is_stopping method. I deleted this one and left the IsStopping method.

Third problem. ResumeCtrlImpl captured the mutex queue_lock_, which belongs to it, and the mutex from ApplicationManagerImpl, when calling the application(..) method. This created a deadlock, and not a secure use of the application object. So I added getting all the apps and working with this list.

Fourth problem. It consists of two different crash of the program. The first crash when remove mobile_request_list_. The second is when deleting mobile requests from waiting_for_response_. This was due to the use of smart pointers in different threads. They are copied into threads, so we have to delete them in the thread in which we finish working with it. Since these requests are processed in a separate thread, I delete them there if the core stops working during the processing of these requests. Also added a check to see if the core stopped working while adding such requests to the list.

The fifth problem. This is a drop while processing data in one thread and deleting it in another thread. I created a separate mutex to protect this data. Because using the existing one, we do not invest in the time allotted to the timer to perform certain work.

### CLA

### DIFF
--- a/src/components/application_manager/test/request_controller/request_controller_test.cc
+++ b/src/components/application_manager/test/request_controller/request_controller_test.cc
@@ -45,7 +45,6 @@
 #include "application_manager/event_engine/event_dispatcher.h"
 #include "application_manager/mock_application_manager.h"
 
-#include "application_manager/mock_application_manager.h"
 #include "application_manager/mock_request_controller_settings.h"
 #include "application_manager/mock_request_timeout_handler.h"
 #include "application_manager/policies/policy_handler.h"
@@ -62,6 +61,7 @@ namespace request_controller_test {
 using ::application_manager::request_controller::RequestController;
 using ::application_manager::request_controller::RequestControllerImpl;
 using ::application_manager::request_controller::RequestInfo;
+using ::test::components::application_manager_test::MockApplicationManager;
 using test::components::application_manager_test::MockRequestTimeoutHandler;
 
 using ::test::components::event_engine_test::MockEventDispatcher;
@@ -168,6 +168,7 @@ class RequestControllerTestClass : public ::testing::Test {
   RequestControllerSPtr request_ctrl_;
   RequestPtr empty_mock_request_;
   const TestSettings default_settings_;
+  NiceMock<MockApplicationManager> app_mngr_;
 };
 
 TEST_F(RequestControllerTestClass,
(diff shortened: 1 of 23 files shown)


RomanReznichenkoLuxoft:
@theresalech This PR is ready for Livio review.

src/components/application_manager/test/request_controller/request_controller_test.cc
```diff
@@ -258,11 +241,23 @@ TEST_F(RequestControllerTestClass, IsLowVoltage_SetOnWakeUp_FALSE) {
 }
 
 TEST_F(RequestControllerTestClass, AddMobileRequest_SetValidData_SUCCESS) {
```
ShobhitAd, line 243: This unit test seems to fail sometimes
```
      Start 17: request_controller_test
17/36 Test #17: request_controller_test .............Child aborted***Exception:   1.16 sec
97% tests passed, 1 tests failed out of 36

Total Test time (real) =  39.75 sec

The following tests FAILED:
	 17 - request_controller_test (Child aborted)
```

Seeing this error message in the Testing/Temporary/LastTest.log

```
[ RUN      ] RequestControllerTestClass.AddMobileRequest_SetValidData_SUCCESS
/home/shobhit/sdl/CORE/test/sdl_core/src/components/application_manager/test/request_controller/request_controller_test.cc:254: Failure
Actual function call count doesn't match EXPECT_CALL(app_mngr_, IsStopping())...
         Expected: to be called once
           Actual: never called - unsatisfied and active
pure virtual method called
terminate called without an active exception
<end of output>
Test time =   1.16 sec
----------------------------------------------------------
Test Failed.
```

ShobhitAd (changes requested):

src/components/application_manager/test/request_controller/request_controller_test.cc
ShobhitAd: Hey @RomanReznichenkoLuxoft, the test seems to fail every 1/50 runs. No special conditions, I saw the test fail the first time I ran `make test` and just re-ran `make test` repeatedly until I encountered the issue again.

RomanReznichenkoLuxoft: Hi @ShobhitAd. Thanks, I also detected locally. Now I analyze the tests and life of objects. I will let you know when there is a fix.

PR #3937, opened 2022-07-18
Fix/Reset sending type on ResetGlobalProperties

### LINKED ISSUE #3931: Core stops generating vrHelp and helpPrompt after ResetGlobalProperties
### Bug Report
If ResetGlobalProperties is sent for VRHELP, VRHELPTITLE, and HELPPROMPT, an auto-generated value for each of these fields is sent initially, but these auto-generated values no longer update after this. 

##### Reproduction Steps
1. Register App
2. Send AddCommand, verify UI.SetGlobalProperties + TTS.SetGlobalProperties are sent with info for this command
3. Send SetGlobalProperties with vrHelp + vrHelpTitle + helpPrompt
4. Send ResetGlobalProperties with VR_HELP, VR_HELP_TITLE, verify UI.SetGlobalProperties + TTS.SetGlobalProperties matches step 1
5. Send another AddCommand

##### Expected Behavior
UI.SetGlobalProperties  + TTS.SetGlobalProperties are sent with data for both commands

##### Observed Behavior
UI.SetGlobalProperties  + TTS.SetGlobalProperties are not sent

* OS/Version: Ubuntu 20.04
* SDL Core Version: master (Release 8.1)
* Testing Against: RPC Builder JS

ShobhitAd:
Fixes #3931

### Risk

### Testing Plan
Submitted ATF test PR for defect

### Summary
Implements a `OnResetGlobalPropertiesReceived` method in the help_prompt_manager. When a ResetGlobalProperties request is received from an application the sending_type_ will be reset based on the global property ids in the request

### CLA

### DIFF
--- a/src/components/application_manager/include/application_manager/help_prompt_manager.h
+++ b/src/components/application_manager/include/application_manager/help_prompt_manager.h
@@ -80,6 +80,14 @@ class HelpPromptManager {
       const smart_objects::SmartObject& msg, const bool is_response) = 0;
 
   /**
+   * @brief Triggered when ResetGlobalProperties request is received from an
+   * application. Reset sending_type_ based on which global properties are reset
+   * @param msg containing GlobalProperties
+   */
+  virtual void OnResetGlobalPropertiesReceived(
+      const smart_objects::SmartObject& msg) = 0;
+
+  /**
    * @brief Requests sending type behavior
    */
   enum class SendingType { kNoneSend, kSendHelpPrompt, kSendVRHelp, kSendBoth };
--- a/src/components/application_manager/include/application_manager/help_prompt_manager.h
+++ b/src/components/application_manager/include/application_manager/help_prompt_manager.h
@@ -97,6 +97,16 @@ class HelpPromptManager {
    * @return current sending type
    */
   virtual SendingType GetSendingType() const = 0;
+
+  /**
+   * @brief Construct the helpPrompt parameter
+   */
+  virtual void CreatePromptMsg(smart_objects::SmartObject& out_msg_params) = 0;
+
+  /**
+   * @brief Construct the vrHelp parameter
+   */
+  virtual void CreateVRMsg(smart_objects::SmartObject& out_msg_params) = 0;
 };
 
 }  //  namespace application_manager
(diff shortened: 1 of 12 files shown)


jacobkeeler (changes requested):
Something I think we missed with https://github.com/smartdevicelink/sdl_core/pull/3916/files

I think that the values set after the reset should match up with the ones generated by `help_prompt_manager.cc` if possible, currently they differ slightly because it uses the MessageHelper function rather than the help prompt manager.

ShobhitAd:
> Something I think we missed with https://github.com/smartdevicelink/sdl_core/pull/3916/files
> 
> I think that the values set after the reset should match up with the ones generated by `help_prompt_manager.cc` if possible, currently they differ slightly because it uses the MessageHelper function rather than the help prompt manager.

I tried that on my end but I'm not sure if the items returned for ResetGlobalProperties and AddCommand should match. 

The `MessageHelper::CreateAppVrHelp` function returns vrHelpItems [appName + VR synonyms + VR Commands] while `HelpPromptManagerImpl::GenerateVrItems` only returns [VR Commands]. The MessageHelper function seems more appropriate for ResetGlobalProperties but if it were combined with the HelpPromptManager function it would return [appName + VR synonyms + VR Commands] even after an AddCommand. 

I guess we combine it in such a way that the function would return [appName + VR synonyms] if no VR Commands or maybe just rename the MessageHelper function to `CreateDefaultAppVrHelp` since that function is currently only used for ResetGlobalProperties

ShobhitAd:
Found another issue while creating ATF tests to test the PR. If an app sends ResetGlobalProperties with **only** `"HELPPROMPT"` core sends a response with `{ success = false, resultCode = "SUCCESS"}`. This seems to be caused due to how the responses are combined in `RequestFromMobileImpl::PrepareResultForMobileResponse` (https://github.com/smartdevicelink/sdl_core/blob/develop/src/components/application_manager/src/commands/request_from_mobile_impl.cc#L664)

Fixed in https://github.com/smartdevicelink/sdl_core/pull/3937/commits/b9750e0d91249e318563723d8f3053bf80781ce0

jacobkeeler:
Given the original intent of [SDL-0122](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0122-New_rules_for_providing_VRHelpItems_VRHelpTitle.md) (providing standard generated help information for an app when it does not provide its own), I think we should use the same format for all cases where the app does not have these fields explicitly set (which includes when they are reset).

Personally I think that we should just make some of the `HelpPromptManager` methods (such as `GenerateVrItems`) public and just use those instead of the existing MessageHelper functions, they seem to be outdated with the implementation of SDL-0122 and should probably just be removed.

jacobkeeler (approved):

